PR #5, opened 2015-08-02
Update autogen.sh

rcseacord:
removed config.h.in~ from the libpng_autotools_files list because it was causing a false positive for missing files.


glennrp:
Revised patch to put config.h.in~ in the "libpng_autools_extra" list.  Thanks!

PR #8, opened 2015-08-07
pngtran.c: remove `break;` from default branch.

vszakats:
```
../../../pngrtran.c:979:10: warning: 'break' will never be executed [-Wunreachable-code-break]
         break;
         ^~~~~
```

### DIFF
--- a/pngrtran.c
+++ b/pngrtran.c
@@ -976,7 +976,6 @@ png_set_rgb_to_gray_fixed(png_structrp png_ptr, int error_action,
 
       default:
          png_error(png_ptr, "invalid error action to rgb_to_gray");
-         break;
    }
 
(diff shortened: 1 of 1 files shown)


glennrp:
I combined the three pull requests regarding removing "break" statements into one and pushed to glennrp/libpng "libpng16" branch.  I did not remove the two "unreachable" png_error calls but instead surrounded them with #ifdef PNG_DEBUG/#endif directives.  Thanks.

vszakats:
Thank you! (I didn't remove them, just moved them around - but it's better this way)

vszakats:
Thanks again @glennrp. I'm closing this for now. Will recheck and report in case any warning remains.
(New results will eventually appear on https://travis-ci.org/vszakats/harbour-core)

PR #10, opened 2015-08-24
rewrote png_save_uint_32 and png_save_int_32 for portability and to eliminate undefined behaviors

rcseacord:
rewrote png_save_uint_32 function to be portable across difference sizes of int without undefined behavior

rewrote png_save_int_32 to use png_save_uint32

tested both functions for all integer values.


glennrp:
See libpng-1.6.22beta03

PR #13, opened 2015-09-15
[libpng16] tests backported from libpng 1.7

jbowler:
This updates libpng16 with all the test changes from libpng17,
including changes to pngvalid.c to ensure that the original,
distributed, version of contrib/visupng/cexcept.h can be used.

pngvalid contains the correction to the use of SAVE/STORE_
UNKNOWN_CHUNKS; a bug revealed by changes in libpng 1.7.  More
tests contain the --strict option to detect warnings and the
pngvalid-standard test has been corrected so that it does not
turn on progresive-read (there is a separate test which does
that.)

Some signed/unsigned fixes have been made.


glennrp:
Merged.

PR #15, opened 2015-09-17
[libpng17] correct low-depth-gray code and enable tests

jbowler:
The low-bit-depth gray tests were disabled in prior versions of libpng because
of problems which should have been fixed by the recent changes to libpng17,
enabling the tests reveals bugs in those changes which are fixed by this commit.


jbowler:
Merge verified; everything is at commit d280c3a now and there were no
substantial change after I back merged to my libpng16-1.  My overnight
tests passed completely on x86_64 (including armv6j, armv7a and aarch64
cross-builds).

I'll look at the Coverity issues next; the ones I've checked so far are
pretty straightforward.

Quite a few of the UNTESTED cases still present are unreachable code;
memcpy skipped when source == dest, which always happens with some
transforms because they only run after something else has made source and
destination the same.
​

PR #16, opened 2015-09-18
Coverity related changes

jbowler:
These should fix most of the reported Coverity issues.  The remaining issues
should be the back_b etc assignments, which look like a Coverity bug, and
passing a pointer to a byte to a function that expects a pointer to one or more
bytes, which should (I believe) be fixed in one case and not the other
(next_filter) case; the latter case will probably go away as I am going to
rewrite that piece of code to avoid a spurious buffer allocation.


glennrp:
Merged.

PR #19, opened 2015-09-18
Libpng15

jbowler:
Two commits, one simply copies the libpng17 pngvalid.c to libpng15, however this requires a fix to png_set_filler which byte-swaps the filler in the 16-bit case (i.e. regardless of whether the 16-bit byte swapping is turned on the filler is always wrong!)

We may have discussed this before, the problem is that this change would be bug-incompatible with the current libpng15 and any app that has worked around the bug without reporting it (!) would break in the minor release.  If you want I can simply make pngvalid.c disable png_set_filler tests in libpng versions less than 1.6

### DIFF
--- a/contrib/libtests/pngvalid.c
+++ b/contrib/libtests/pngvalid.c
@@ -1,8 +1,8 @@
 
 /* pngvalid.c - validate libpng by constructing then reading png files.
  *
- * Last changed in libpng 1.5.18 [February 6, 2014]
- * Copyright (c) 2014 Glenn Randers-Pehrson
+ * Last changed in libpng 1.6.18 [(PENDING RELEASE)]
+ * Copyright (c) 2014-2015 Glenn Randers-Pehrson
  * Written by John Cunningham Bowler
  *
  * This code is released under the libpng license.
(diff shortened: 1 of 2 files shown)


glennrp:
We fixed the filler byte order problem in libpng-1.6.17, so it's bug-incompatible with 1.6.16.  I don't remember promising bug-compatibility, so we might as well fix libpng-1.5.x as well.

jbowler:
Worth a warning though.  If someone encountered it and simply worked round
it without filing a bug I think the result is their own problem.

John

On Fri, Sep 18, 2015 at 7:12 PM, Glenn Randers-Pehrson <

> We fixed the filler byte order problem in libpng-1.6.17, so it's
> bug-incompatible with 1.6.16. I don't remember promising bug-compatibility,
> so we might as well fix libpng-1.5.x as well.
>
> —
> <https://github.com/glennrp/libpng/pull/19#issuecomment-141609463>.
>



-- 
+1 (541) 450-9885
PO BOX 3151
KERBY OR 97531-3151
USA

jbowler:
Verified

glennrp:
On Fri, Sep 18, 2015 at 10:14 PM, John Bowler <notifications@github.com>
wrote:

> Worth a warning though. If someone encountered it and simply worked round
> it without filing a bug I think the result is their own problem.
>

Where should the warning go (besides already in CHANGES and ANNOUNCE?)

G

jbowler:
I just meant in CHANGES/ANNOUNCE; something to the effect that if the
argument to png_set_filler has 16 bits they will now be interpreted
correctly rather than being byte-swapped.

John
​

glennrp:
On Fri, Sep 18, 2015 at 11:22 PM, John Bowler <notifications@github.com>
wrote:

> I just meant in CHANGES/ANNOUNCE; something to the effect that if the
> argument to png_set_filler has 16 bits they will now be interpreted
> correctly rather than being byte-swapped.


It's an old bug; libpng10, 12, and 14 still do it wrong.

Fixed in libpng-1.5.24, 1.6.17, and 1.7.0beta46

Glenn

PR #22, opened 2015-09-21
Libpng17

jbowler:
Updated licensing in the contrib/testpngs test files, more test files (covering cases of non-palette files with tRNS chunks) and the previous simplified API sRGB gamma improvement for palette cases.


glennrp:
Squash-merged 8c025bc through bf51c64

PR #29, opened 2015-09-26
Libpng17 s bit tests

jbowler:
Some fixes and additional sBIT tests.


jbowler:
e76a320 is necessary when READ_GAMMA is on but the simplified API is disabled

jbowler:
I think includes the dropped tests/* renames/fixes from the previous pull request, at least I think that is why 8d6c21c is in there.

PR #39, opened 2015-11-19
pngread: avoid clang warning for unreachable code

vszakats:
```
pngread.c:2841:10: warning: 'break' will never be executed [-Wunreachable-code-break]
         break;
         ^~~~~
```

### DIFF
--- a/pngread.c
+++ b/pngread.c
@@ -2838,7 +2838,6 @@ png_image_read_colormap(png_voidp argument)
       default:
          png_error(png_ptr, "invalid PNG color type");
          /*NOT REACHED*/
-         break;
    }
 
    /* Now deal with the output processing */


glennrp:
I've pulled this patch; however, note that my C reference manual (Harbison and Steele) says "While the last break statement is logically unnecessary, it is a good thing to put in as a matter of style."

vszakats:
Thank you! Thanks for the note as well &mdash; it be can debated for sure, `clang` opted to include this in the `-Weverything` warning level, which is also up to debate, but probable some projects will have this enabled even for `libpng`, so I welcome your accepting this PR.

glennrp:
If you have any influence with the clang developers, suggest that they
recognize /* NOT REACHED */ just as they recognize /* FALL THROUGH */
already.

Glenn

vszakats:
Not much, I'm afraid, but I'll keep that in mind, in case I ever will be. Didn't know `/* FALL THROUGH */` is recognised, though, thanks for the tip, and I agree `NOT REACHED` would be very nice in some cases.

glennrp:
I'm not actually sure that clang recognizes /* Fall through */ but I do
know that it suppresses warnings from Coverity.

vszakats:
Coverity: That might be it, I did have added various comments to suppress false alarms with it. With clang, this is the syntax to control warnings: https://stackoverflow.com/questions/14444203/dynamic-forwarding-suppress-incomplete-implementation-warning/14444603#14444603

vszakats:
Just got the suggestion below, after upgrading to clang 3.7, so comment suggestions do exist:
```
harboury.c:6939:21: note: silence by adding parentheses to mark code as
      explicitly dead
  if (/*CONSTCOND*/ 0)
                    ^
                    /* DISABLES CODE */ ( )
```

vszakats:
(last comment deleted - I misinterpreted the output)

PR #43, opened 2015-11-23
Libpng17 buffer enhancements

jbowler:
Fixes for the code where libpng interlace handling is disabled (completely; both read and write).

### DIFF
--- a/contrib/libtests/pngimage.c
+++ b/contrib/libtests/pngimage.c
@@ -36,7 +36,8 @@
 #  include <setjmp.h> /* because png.h did *not* include this */
 #endif
 
-#if defined(PNG_INFO_IMAGE_SUPPORTED) && defined(PNG_SEQUENTIAL_READ_SUPPORTED)
+#if defined(PNG_INFO_IMAGE_SUPPORTED) && defined(PNG_SEQUENTIAL_READ_SUPPORTED)\
+    && defined(PNG_READ_PNG_SUPPORTED)
 /* If a transform is valid on both read and write this implies that if the
  * transform is applied to read it must also be applied on write to produce
  * meaningful data.  This is because these transforms when performed on read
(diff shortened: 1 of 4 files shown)


glennrp:
After pulling #43, I get

FAIL: tests/pngvalid-interlace-transform
during make dist-check

But then ./configure; make pngvalid, and running the test it passes.

gcc (Ubuntu 4.8.4-2ubuntu1~14.04) 4.8.4

Glenn

jbowler:
It might be using some uninitialized data somehow; of all my builds (1.7)
the only issue I have is some kind of weird pixel arithmetic problem in
interlace-transform when PNG_READ_INTERLACING_SUPPORTED is disabled.  I.e.
something is wrong with either the pngvalid handling of interlace when
libpng doesn't do it or the libpng code in that case (the latter seems
unlikely).

That issue has probably always been there.

John

On Mon, Nov 23, 2015 at 7:07 AM, Glenn Randers-Pehrson <

> After pulling #43, I get
>
> FAIL: tests/pngvalid-interlace-transform
> during make dist-check
>
> But then ./configure; make pngvalid, and running the test it passes.
>
> gcc (Ubuntu 4.8.4-2ubuntu1~14.04) 4.8.4
>
> Glenn
>
> —
> <https://github.com/glennrp/libpng/pull/43#issuecomment-158962597>.
>



-- 
+1 (541) 450-9885
PO BOX 3151
KERBY OR 97531-3151
USA

glennrp:
The permission on the test file was 644 instead of 755, when I updated via
43.patch instead of via a "pull".

It's OK now.

G

jbowler:
There's a chance it will make the homebrew bug disappear; quite a large
chance because it shifts the error limits.  In the case where I hit it my
(broken) attempts to take into account the rounding in the gamma correction
(I used 11 bit precision when I should have been using 16) shifted the high
limit below the exact value, so the test was against the exact value.

Normally I would expect two floating point systems with the same precision
to get the same answer, but quite possibly the CPU they have has a
different precision from the three I have to test (Intel x86_64, armv6j,
armv7a).  Also the 'pow' function could easily be wrong.

John

PR #45, opened 2015-11-24
Libpng17 20151124

jbowler:
This is the pngvalid fix.   (Note that there really is only one pull request in this, but it is relative to the previous pull request).

pngvalid had one error in rgb_to_gray and an error-collection in the handling of PNG_MAX_GAMMA_8.  Correcting all the uses of the latter to match the actual behavior just eliminates one problem I was seeing in pngvalid --transform when interlace handling was disabled (the latter is irrelevant; it just causes the test cases to change.)

Since the error in PNG_MAX_GAMMA_8 caused the tests to be far more conservative wrt errors it may explain what is going on elsewhere.

### DIFF
--- a/contrib/libtests/pngimage.c
+++ b/contrib/libtests/pngimage.c
@@ -37,7 +37,7 @@
 #endif
 
 #if defined(PNG_INFO_IMAGE_SUPPORTED) && defined(PNG_SEQUENTIAL_READ_SUPPORTED)\
-    && defined(PNG_READ_PNG_SUPPORTED)
+    && (defined(PNG_READ_PNG_SUPPORTED) || PNG_LIBPNG_VER < 10700)
 /* If a transform is valid on both read and write this implies that if the
  * transform is applied to read it must also be applied on write to produce
  * meaningful data.  This is because these transforms when performed on read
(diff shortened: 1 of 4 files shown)


glennrp:
I applied 44 patch then attempted to apply 45 patch, but got complaint about "-R" patching.
I was however able to "pull" both merges into my repo.  Maybe everything's OK but I'm not sure.

glennrp:
After applying 44 and 45 I still see "FAIL: tests/pngvalid-interlace-transform" during "make distcheck".

jbowler:
Do a git diff against my libpng-1/libpng17-20151124 branch.  I suspect when
you applied the #45 you hadn't pushed #44 back to github and it simply
undid it.

John

On Mon, Nov 23, 2015 at 7:30 PM, Glenn Randers-Pehrson <

> After applying 44 and 45 I still see "FAIL:
> tests/pngvalid-interlace-transform" during "make distcheck".
>
> —
> <https://github.com/glennrp/libpng/pull/45#issuecomment-159140716>.
>



-- 
+1 (541) 450-9885
PO BOX 3151
KERBY OR 97531-3151
USA

jbowler:
On Mon, Nov 23, 2015 at 7:30 PM, Glenn Randers-Pehrson <

> After applying 44 and 45 I still see "FAIL:
> tests/pngvalid-interlace-transform" during "make distcheck"
>
Ok, I checked your github libpng17 and it is identical.  Can you send me
both the tarball and the tests/pngvalid-interlace-transform.log from the
failed build (the log is, IRC, in the _build directory, it shouldn't have
been deleted, the tarball is, IRC, in _check, there might be two, one at
the top level.)

John

glennrp:
The problem was that I had used the "patch -i ..." rather than the "git pull" to update my test copy of the library, and the pngvalid-interlace-transform script got permission 644 instead of 755.

PR #52, opened 2015-12-01
Libpng17 20151130

jbowler:
This separates the png_struct and png_info palette/trans_alpha so they are no longer shared.  Tested on all builds.


glennrp:
After this change, is it now possible to call png_set_tRNS() and/or png_set_hIST() before calling png_set_PLTE()?

jbowler:
I checked 1.6.16:

png_set_PLTE: can be called any time, can be called multiple times.
png_set_tRNS: can be called any time, can be called multiple times.
png_set_hIST: only works after a call to png_set_PLTE, uses num_palette
from the last call, can be called multiple times.

I think the last one is clearly wrong, but it seems to have been that way
for a while.  Changing png_set_tRNS, however, seems like to break existing
apps.  Anyway, with this merge stuff is back to what the 1.6 HEAD does (but
not v1.6.16):

png_set_PLTE: now faults palettes longer than the bit depth implies
png_set_tRNS: still accepts any length of tRNS chunk, but png_write_tRNS
will drop an overlength chunk
png_set_hIST: unchanged, still checks against info_struct::num_palette

Checks against IHDR values are ok in png_set.  The check against
num_palette in png_set_hIST is, however, wrong.  It's an inter-chunk test
so it needs to be in png_write_hIST.

There's actually a similar problem in pngrutil.c which I've been aware of
for some time; out of order chunks cause the error checking to be very
difficult to verify.  Really the parser needs to work out which chunks are
valid in the stream (detect/reject out-of-order) then validate the result
when the first IDAT is encountered.  That's a major change but it's on my
list for 1.7.  In principle it would allow the read and write checking to
be shared.

John

On Tue, Dec 1, 2015 at 9:36 AM, Glenn Randers-Pehrson <

> Merged #52 <https://github.com/glennrp/libpng/pull/52>.
>
> —
> <https://github.com/glennrp/libpng/pull/52#event-479229912>.
>



-- 
+1 (541) 450-9885
PO BOX 3151
KERBY OR 97531-3151
USA

PR #58, opened 2015-12-15
[libpng15] update pngvalid from libpng 1.7 with fixes

jbowler:
Tested on 1.5, 1.6 and 1.7; I will submit separate merge requests but the file should be identical in all three cases.  This does the following:

1) --strict (and -w) now work correctly; they do nothing because the default is to treat png_warning as an error, I added --nostrict to turn this off.
2) The internal pngvalid check on the error limits generated by pngvalid in the rgb_to_transform case has been relaxed for 1.5 and 1.6 (1.7, where the pngvalid error range is considerably less, is unchanged.)
3) In all versions the internal limit check is disabled in RC and release builds.
4) The previous attempt to fix the limit check has been undone.


glennrp:
I'll pull this after the current release cycle is finished on December 17th.

jbowler:
Ok, github has auto-incorporated my last push to my libpng15 branch into this merge request.  I'm still running tests here, but nothing has failed so far and I've got past all the GCC tests into the clangers.

PR #87, opened 2016-02-16
Fix win errors:

NixonK:
These are the errors that I came across and my fixes (using cmake on Windows):
- fatal error C1083: Cannot open source file: ' /DWIN32 /D_WINDOWS /W3': No such file or directory [~\genfiles.vcxproj]
  - removed quotes from line 17; set as a list to strip leading spaces
- fatal error C1083: Cannot open include file: 'zlib.h': No such file or directory [~\genfiles.vcxproj]
  - added line 51; INCLUDES was set but not used in execute_process

### DIFF
--- a/scripts/genout.cmake.in
+++ b/scripts/genout.cmake.in
@@ -14,7 +14,7 @@ set(BINDIR "@CMAKE_CURRENT_BINARY_DIR@")
 
 set(AWK "@AWK@")
 set(CMAKE_C_COMPILER "@CMAKE_C_COMPILER@")
-set(CMAKE_C_FLAGS "@CMAKE_C_FLAGS@")
+set(CMAKE_C_FLAGS @CMAKE_C_FLAGS@)
 set(INCDIR "@CMAKE_CURRENT_BINARY_DIR@")
 set(PNG_PREFIX "@PNG_PREFIX@")
 set(PNGLIB_MAJOR "@PNGLIB_MAJOR@")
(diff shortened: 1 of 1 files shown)


glennrp:
We agree with adding ${INCLUDES} and have done that in the libpng16 branch, but the problem with quotation marks may be with the way you were calling cmake.  What happens if you remove the single-quotes around the flags in your build script?

NixonK:
I haven't set the flags myself but have been using Visual Studio 11 2012 Win64 as the generator with this configure command:
cmake -G " Visual Studio 11 2012 Win64" -DCMAKE_PREFIX_PATH=<-> -DCMAKE_INSTALL_PREFIX:PATH=<->

NixonK:
This instead is only one change but I'm thinking the fix is unique to my side:
```
@@ -15,6 +15,7 @@ set(BINDIR "@CMAKE_CURRENT_BINARY_DIR@")
 set(AWK "@AWK@")
 set(CMAKE_C_COMPILER "@CMAKE_C_COMPILER@")
 set(CMAKE_C_FLAGS "@CMAKE_C_FLAGS@")
+string(STRIP "${CMAKE_C_FLAGS}" CMAKE_C_FLAGS)
```

jbowler:
The cmake gurus still need to comment; it's a matter of definition/documentation in the build environment whether quotes are stripped or not and it has to be defined because a cmake user has to know to get the quotes right in some cases!

However, as a neophyte, it seems correct to me that the _user _ CMAKE_C_FLAGS value should never have  quotes round the whole value.  So CMAKE_C_FLAGS should just be a sequence of tokens which themselves might contain matched quotes (e.g. if there are spaces in one of the tokens).

In practice support for embedded quotes doesn't matter for the vast majority of users in 1.6 because the tokens in CMAKE_C_FLAGS don't themselves contain spaces or special characters.  An exception might arise because of the SYMBOL_PREFIX handling, but that does seem to be happening independently of CMAKE_C_FLAGS.  In 1.7 it's a little different because support for PNG_ZLIB_HEADER depends, absolutely, on the correct quote stripping; the value will typically be something of the form "file-name" (where the value really does have the double quotes).

Nevertheless it seems clear from the above that the rule for processing dfn/dfa files (which uses @AWK@) is stripping one less level of quotes than the rule for processing .c files (i.e. the normal compile) and I can't work out why that is; guru required!

John Bowler

glennrp:
This was fixed by landing dacap's pull request.

PR #88, opened 2016-02-16
Intel optimizations for png filter functions

mattsarett:
Please let me know your thoughts, and if there is anything I can do to help this integrate smoothly.  This shouldn't cause any decoding differences.

I haven't modified the libpng build system (since we have our own build files), but I'm happy to take this on if it would be helpful.

Thanks!
Matt


glennrp:
I attempted "./autogen.sh; ./configure; make" on Ubuntu 14:04, x86_64
and ended up with

    libtool: link: gcc -g -O2 -o .libs/pngfix contrib/tools/pngfix.o  ./.libs/libpng16.so -lz -lm
    ./.libs/libpng16.so: undefined reference to `png_init_filter_functions_sse2'
    collect2: error: ld returned 1 exit status

Evidently I'm missing something.

mattsarett:
That's because I hesitated to modify the libpng build system (we use a different build system) - the new files probably need to be added to the Makefile.  I'll fix this and post an update.

glennrp:
Right.  Look for arm_init.c in Makefile.am and put similar code for sse there, and also add sse stuff near the end of configure.ac in the "HOST SPECIFIC OPTIONS" section.

mattsarett:
The latest two commit should allow libpng to build successfully.

I should mention that the configure.ac file is somewhat foreign to me.  I've mostly just followed the pattern of the Arm code.

Let me know if this works for you!

glennrp:
./autogen.sh; ./configure; make; make test
works for me now on
    Ubuntu, gcc-4.8.4-2ubuntu1~14.04.1
    and on SunOS 5.10, gcc-4.7.1

contrib/intel/filter_sse2_intrinsics.c
```diff
@@ -0,0 +1,285 @@
...
+#else
+   // Read this all as, return x<0 ? -x : x.
+   // To negate two's complement, you flip all the bits then add 1.
+   __m128i is_negative = _mm_cmplt_epi16(x, _mm_setzero_si128());
+
+   // Flip negative lanes.
+   x = _mm_xor_si128(x, is_negative);
+
+   // +1 to negative lanes, else +0.
+   x = _mm_add_epi16(x, _mm_srli_epi16(is_negative, 15));
```
mayeut: `x = _mm_sub_epi16(x, is_negative);` would remove the need for a shift

mattsarett: Yes I think so!  I would be in favor to upstreaming this to libpng.  I'd encourage you to propose the change to glennrp.  Were you able to measure a performance improvement?

mayeut: Not tested yet. I was just reviewing the code.

glennrp: OK no problem.  Just let me know when you have a patch (or a pull request)
ready.

Glenn

glennrp:
Still awaiting a patch or pull request from mayeut. I can check the performance improvement, if any.

felixbuenemann:
@mattsarett Do you have any benchmarks for this patch?

I'm not really seeing any difference when comparing libpng master with and without this patch, while using the CloudFlare zlib instead of the vanilla one gives a huge speedup, but I could jsut be testing the wrong things.

mattsarett:
Can you share a png?  This patch will only affect pngs that use the paeth, avg, or sub filters and are RGBA or RGB.

FWIW, png decodes typically spend more time in zlib than on the filtering step, so improvements to zlib would likely lead to more noticeable performance improvements.

glennrp:
An instrumented pngcrush-1.8.7 says that the INTEL optimization cuts the time about in half to defilter a PAETH-filtered line of RGB pixels.  But the total decoding time is dominated by other things.

felixbuenemann:
I tried with two source images:

* [Earth10k.jpg](https://dl.dropboxusercontent.com/u/7919548/vips/Earth10k.jpg)
* [Earth10k.png](https://dl.dropboxusercontent.com/u/7919548/vips/Earth10k.png)

and encoded to PNG without resizing at compression level 3 with sub, up, paeth or all filters using [vips](https://github.com/jcupitt/libvips):

```
vips copy Earth10.png '10k.png[compression=3,filter=paeth]
```

I am seeing a measurable difference from switching from filter=all to for example up, which is fastest and best on this image. But I don't see any real difference between libpng with or without the sse patch.

I'm using libpng master compiled with Apple's latest clang 7.3.0 (clang-703.0.31) on MacOS X 10.11.6 on a Core i7-5557U 3.1GHz.

mattsarett:
Are you measuring encode performance or decode performance?  The filter optimizations only affect decodes.  Also, the filter optimizations do not affect the UP filter.  Your compiler will autovectorize that anyway.

felixbuenemann:
Ah OK, I was trying to speed up encoding. If it's decode only that explains why I don't see any difference :-)

PR #94, opened 2016-03-31
Fix CMAKE_C_FLAGS with multiple values when genout is compiled on MSVC

dacap:
Without this patch, CMAKE_C_FLAGS will be processed as a string/filename on MSVC (e.g. when `CMAKE_C_FLAGS` is equal to `" /D_WIN32 /D_WINDOWS /W3"`) and the whole compilation fails.

(Same issue as in #87)


glennrp:
Duplicate of #87 Fix win errors

PR #96, opened 2016-03-31
Fix pnglibconf.c compilation on OS X including the sysroot path

dacap:
Without these flags pnglibconf.c compilation fails because it includes zlib.h -> zconf.h -> sys/types.h which is not found until we give a correct SDK path with -isysroot flag.


glennrp:
See issue #100 -- this patch seems to have broken clang builds.

dacap:
Hi @glennrp, there is a patch for this here: https://github.com/aseprite/libpng/commit/74a4abe8d5b73fe10767121e7360aeebba80b9ed

Do you want a PR or can you just cherry-pick the commit? (the commit is in `aseprite` branch in https://github.com/aseprite/libpng)

glennrp:
PR would be helpful

PR #103, opened 2016-05-06
Minor write bug-fixes, remove unimplemented code

jbowler:
A debug() assert fired if windowBits was set to 8 for the Huffman only and
no-compression cases.  This commit changes it to do some extra checking.  Remove
unreachable code in pz_default_settings, eliminate a spurious warning in pngcp
for small files.


jbowler:
This should fix the Coverity dead code issue.

PR #105, opened 2016-05-29
pngvalid.c: validate weird IDAT lengths

jbowler:
Also pngpread.c, fix a warning that is detected as a result.

(NOTE: requires a patch to pngpread.c for the progressive tests to pass.)


glennrp:
I applied this patch manually to libpng-1.6.23beta01.

PR #109, opened 2016-05-30
pngvalid.c: don't use size_t count arguments

jbowler:
Coverity rejects code where an array element count has type size_t, this
elminates the code in question from contrib/libtests/pngvalid.c


glennrp:
I'll apply this patch to libpng15, libpng16, and libpng17.

PR #113, opened 2016-06-11
Change image size checks

jbowler:
This covers the case where PNG_IMAGE_BUFFER_SIZE can overflow in the application
as a result of the application using an increased 'row_stride'; previously
png_image_finish_read only checked for overflow on the base calculation of
components.  (I.e. it checked for overflow of a 32-bit number on the total
number of pixel components in the output format, not the possibly padded row
length and not the number of bytes, which for linear formats is twice the number
of components.)


glennrp:
Patch applied manually

PR #116, opened 2016-06-16
Fix png encoding perf regression in libpng 1.6

mattsarett:
Applications that care about encoding performance may choose to
select a fast encoding filter.  Ex:
png_set_filter(png, PNG_FILTER_TYPE_BASE, PNG_FILTER_SUB);
or
png_set_filter(png, PNG_FILTER_TYPE_BASE, PNG_FILTER_UP);

This improves encode performance (1) because applying the filter
is fast and (2) because we have selected a single filter, there
is no need to calculate and track filter choosing heuristics.

libpng 1.2 takes advantage of both (1) and (2), but I'm seeing a
significant performance regression in libpng 1.6 (1.75x slower on
a particular benchmark) because it does not take advantage of (2).
png_setup_sub_row() is inlined, so the compiler should be able to
detect that "sum" is unused and decide not to calculate it.
However, since we check the "sum" on every loop iteration, the sum
is being calculated regardless of whether we need it or not.

I'm proposing that we don't bother to check (sum > lmin) on the
"fast filters" since they are fast anyway (so it's not too bad to
complete the entire row, even if it turns out to be unnecessary).
This will make them much faster when they are the only filter
selected.

Another potential fix would be to actually inline the
png_setup_sub_row() code (like libpng 1.2 does) rather than
sharing the code in a function.

Let me know what you think!

### DIFF
--- a/pngwutil.c
+++ b/pngwutil.c
@@ -2270,9 +2270,6 @@ png_setup_sub_row(png_structrp png_ptr, const png_uint_32 bpp,
    {
       v = *dp = (png_byte)(((int)*rp - (int)*lp) & 0xff);
       sum += (v < 128) ? v : 256 - v;
-
-      if (sum > lmins)  /* We are already worse, don't continue. */
-        break;
    }
 
    return (sum);
@@ -2295,9 +2292,6 @@ png_setup_up_row(png_structrp png_ptr, const png_size_t row_bytes,
    {
       v = *dp = (png_byte)(((int)*rp - (int)*pp) & 0xff);
       sum += (v < 128) ? v : 256 - v;
-
-      if (sum > lmins)  /* We are already worse, don't continue. */
-        break;
    }
 
    return (sum);


jbowler:
You might like to take a look at the arguments to 'png_set_compression' in libpng 1.7; around line 1696 in png.h and implemented primarily in pngwutil.c .  I got those settings using about 147,000 test files and (in fact) primarily the 1.6 encoder; they're derived primarily from consideration of the file *size* then working out what switching off the options that slow down the encoder and decode speed does to the file size.  There's a set of figures around line 1353 of pngwutil.c in the implementation of pz_default_settings.

Anyway, 1.7 doesn't suffer from the issue you identified in 1.6; the single-filter route bypasses the entire heuristic calculation, however libpng 1.7 just does all the filters at once so it's probably slower as a result.  We've never optimized the write code!

As for your fix it doesn't work because in the 'else' branch where mins is used it can overflow.  An example of the failure that results is with the vertical-stripe PNG, where 'up' is optimal but 'sub' can overflow (particularly on 16-bit systems).

Anyway, isn't there a bug in there?   Like, a fairly serious one?  Doesn't it bail anyway, even if sum isn't used (as a result of the (void) of the call result?)  I think that probably leaves a row terminated by garbage, but maybe I'm seeing bugs where they don't exist.  If I'm right the (correct) fix to the overflow bug would probably fix your issue without potentially selecting the wrong filter, though it would probably slow down the rest of the code.

One simple fix is to delete the overflow test everywhere except the basic test at line 2454; move that to the head of the routine and if it fails execute 'filter_to_do &= -filter_to_do' to leave just one (the lowest numbered) filter.

jbowler:
libpng 1.0 selects the 'NONE' filter if all the filters in the list overflow *and* 'NONE' is *not* in the list.  If 'NONE' is in the list the sum is calculated for it but overflow is ignored.  If overflow doesn't happen this might cause overflow in the subsequent calculations to be ignored, if the 'NONE' sum is >= 0xfffffff8.

Given this behavior selecting the first filter in the list if overflow is possible seems reasonable.  Passing an 'mins' of PNG_SIZE_MAX to the png_setup routines in the case where there is only one filter will prevent the 'break' and turn the optimization back on (because the fixed value for lmins will cause the test and break to be eliminated in the in-line version.)

jbowler:
I tested:

https://github.com/jbowler/libpng/tree/20160618-filter-select

I think it's correct, it passes all my tests (on x86_64), but it's another case of code review required; I find it very difficult to even see the original bug and once I've been able to see it it's difficult to keep in the crosshairs.  I think what I wrote is more clear, but then I wrote it...

glennrp:
See commit 29135161d74e5f6f640ec1fc2ab3ceebc34e339b

mattsarett:
Thanks a lot for this fix!  I've verified locally that this fixes the performance regressions.

mattsarett:
So it turns out that this fix works great on Mac/Linux clang, but has no effect on some GCC and Windows builds.

I know it's a pain to support various versions of dumb compilers, but can we reopen this and reconsider landing https://github.com/glennrp/libpng/commit/29135161d74e5f6f640ec1fc2ab3ceebc34e339b ?

This seems to fix the problem everywhere.

jbowler:
I don't think that's appropriate.  Different compilers produce different
optimizations; some are better than others.  Some architectures and some
compilers favor code size over CPU time, for very good reasons; loading the
code can take as long as running it.

Manual code duplication cannot be easily undone by the compiler.  It needs
to be done very very carefully and certainly not because a given compiler
chooses to *not* do it.  Normally the compiler writer knows best.

The amount of inlining done is very architecture dependent, although with
GCC there is explicit control, so what happens with those options?  Anyway;
what are the compilers (i.e. the arch-os-compiler plus optimization
options?)

This is all particularly true in this case; we are talking about two or
three extra instructions per byte yet there are hundreds of *other*
instructions per byte.  Even though the case is concocted (turning off
everything except 'none' and 'up') the whole of the deflate code is still
there and it does a string search for each byte.

So, second question, what are the overall timing figures?
​

mattsarett:
```
I don't think that's appropriate.  Different compilers produce different
optimizations; some are better than others.  Some architectures and some
compilers favor code size over CPU time, for very good reasons; loading the
code can take as long as running it.

Manual code duplication cannot be easily undone by the compiler.  It needs
to be done very very carefully and certainly not because a given compiler
chooses to *not* do it.  Normally the compiler writer knows best.
```

I can't speak for other clients, but I suspect that we care more about performance than the trivial (for a large application) increase in code size that this duplication would cause.  Though I'm aware that not all clients are so large...

> The amount of inlining done is very architecture dependent, although with
> GCC there is explicit control, so what happens with those options?  Anyway;
> what are the compilers (i.e. the arch-os-compiler plus optimization
> options?)

Agreed that all of this stuff is very architecture/compiler dependent.  That's why, IMHO, fixing a performance regression with a compiler dependent optimization might not make sense.

Here's a few specifics on the impact.  The benchmark test I'm running is dominated by PNG encode time.

Android Nexus 5X (GCC 4.6)
Benchmark Time libpng1.2   871ms
Benchmark Time libpng1.6  1114ms

Windows 7 (MSVS)
Benchmark Time libpng1.2  189ms
Benchmark Time libpng1.6  246ms

I can't provide specifics on the compiler options, but FWIW, they are set to maximize optimizations.

> This is all particularly true in this case; we are talking about two or
> three extra instructions per byte yet there are hundreds of *other*
> instructions per byte.  Even though the case is concocted (turning off
> everything except 'none' and 'up') the whole of the deflate code is still
> there and it does a string search for each byte.

I tend to disagree that this case is "concocted".  Setting the encode options to a specific, fast filter seems like a reasonable way to client to optimize filter runtime.

It is strange that this set of a few instructions is causing such a large slowdown.  I would guess that having a branch instruction inside of what would be a loop of plain arithmetic prevents auto-vectorization of the code.

FWIW, we also make similar "optimized for runtime" decisions on the zlib settings.

-----------------------------------------------------------------------------------

I hope the above doesn't come across as argumentative - I'm aware that there is a certainly a performance vs. code size/clarity/maintainability trade-off here.  And that's a decision that you (as libpng maintainers) certainly should be the ones to make.

What's troublesome for me is that, in trying to update an application's libpng, I have the (possibly unfair) burden of demonstrating that the new version is "better" than the old across a wide variety of tests.  And I don't want to let that block what I believe is a positive change.

jbowler:
Those aren't the figures I was asking about; I was interested in the
difference in the total time between the two versions of pngwutil.c
currently in the 1.6 tree.  Specifically between Glenn's backed out change
from commit 291xxx and the tip revision, commit ed5xxx.  It is easy just to
git-checkout each of the pngwutil.c versions and test and, indeed, that is
what I am doing.

It's pretty near impossible for me to hypothesize where a 30% difference in
time comes from when two such different pieces of code are compared.  When
I compare 291 with ed5 using tests that are substantially 'write', take
about 160s to run and can be analyzed statistically (I run them a number of
times) So far I haven't got a significant difference in time, but the 291
executable is certainly bigger.  I.e. the original bug report is currently
not-repro; if I can repro it I can run perf and find out what's going on.
​
>What's troublesome for me is that, in trying to update an application's
libpng,
>I have the (possibly unfair) burden of demonstrating that the new version
is "better"
>than the old across a wide variety of tests.

No, you can just patch libpng any way you like; Glenn has already provided
a different version!  The issue is that changes to the Official source code
have to be for the benefit of everyone (on aggregate), not just one very
specific commercial case.  In the past (when I was actually paid to do
this) I maintained patched versions of libpng because the changes I made
were for a single, proprietary, application and the management (well, more
senior developers) were on my back about code bloat.

John Bowler

mattsarett:
Sorry for being unclear.  The benchmark times I provided for libpng 1.2 are virtually identical to libpng 1.6 with Glenn's patch.  Please interpret those times as "libpng 1.6 plus Glenn's patch".  The 30% difference is entirely due to the sum calculation and branch instruction in the middle of the png_setup_sub_row() loop.

I'm disappointed you can't reproduce, though I'd guess that's because this seems to be a compiler/platform dependent case.  Is it possible for you to try Windows MSVC or an Android device under the following settings?
png_set_compression_level(png, 3);
png_set_compression_strategy(png, Z_FILTERED);
png_set_filter(png, PNG_FILTER_TYPE_BASE, PNG_FILTER_SUB);

> No, you can just patch libpng any way you like.

Yes that is certainly a potential solution.  Obviously, I'd prefer to have a libpng that is maintained by the libpng experts (you and Glenn), rather than taking than taking on that responsibility.  And once we start adding patches to libpng, it makes it much more difficult to update to future versions.

jbowler:
On Mon, Jun 27, 2016 at 12:53 PM, mattsarett <notifications@github.com>
wrote:

> Sorry for being unclear. The benchmark times I provided for libpng 1.2 are
> virtually identical to libpng 1.6 with Glenn's patch. Please interpret
> those times as "libpng 1.6 plus Glenn's patch". The 30% difference is
> entirely due to the sum calculation and branch instruction in the middle of
> the png_setup_sub_row() loop.
>
So I can't repro *that*; with gcc-5.4.0 -O2 -finline-functions and using
gdb when I put a breakpoint on png_setup_sub_row it does, as expected, put
a break in two locations.  When I run "pngcp --filter=sub" I do hit the
PNG_SIZE_MAX case and there is no png_setup_sub_row function.  When I step
through this there is no 'sum' calculation.

With Glenn's code the same thing happens regardless of -finline-functions
because there is only one call to each of the static functions so, at -O2,
GCC inlines them because '-finline-functions-called-once' is turned on at
-O2.

So I think you can see what I suspect...

I don't have 4.6; it's pretty old.  I checked the manual for 4.6.4 and the
description of -finline-functions and -finline-functions-called-once is the
same, I guess maybe it could have a bug, but I think you should double
check this statement:

>I can't provide specifics on the compiler options, but FWIW,
>they are set to maximize optimizations.

I can't see why those instructions exist in your code if you really have
enabled -O3 (i.e. optimize for speed even at the cost of significant space
increase).

I'll test with those zlib options; it's going to be interesting if the
inline vs non-inline case does produce +30%...

mattsarett:
Thanks for looking into this!  I'll do my best to verify the compiler flags.  Agreed that GCC 4.6 is old.

jbowler:
I used GCC 4.8.5 and a minimally configure libpng, producing (at -O2) a statically linked pngcp of text size 58kbytes on X86-64.  I used Glenn's hand duplicated 291xxx revision and the 1.6 ed5xxx tip revision.

Using a randomly selected test set of 59002 files from my bigger test set I ran pngcp with the above zlib options and filter=sub on all the files at once.  This took 80s with 291.  Swapping to the non-inlined version in ed5 added 4s to this time.  Using other tests such as forcing 291 through the other (still inlined) branch with filter=sub,none and testing ed5 with -finline-functions showed that it isn't the inlining that helps; it is the code duplication.

So I looked at what perf reports for the two versions; perf diff 291 ed5.  ed5  was about 5% slower than 291 and perf record did record 4.8% more samples, perf diff --compute=ratio showed that most of the time when inside deflate; the top three functions were all part of deflate and took over 56% of the total time in 291.  inflate and the libpng Paeth filter accounted for more than 20% of the remaining time.

So far as png_write_find_filter is concerned, however, it took 1.79% of the time in 291 and the non-inlined function png_setup_sub_row took the equivalent of 6.56% of the samples in ed5 (png_write_find_filter did not appear in ed5).  The difference, 4.79%, accounts for the increased samples in ed5 even though other sample counts moved around quite a lot in other functions; I assume that this is simply a result of the change in png_write_find_filter changing where cache waits happen.  The time increase is 6.56/1.79 = 3.66.

So... examining the code under gdb, comparing the two inlined/hand crafted versions of png_setup_sub_row, seems to show that the problem is the compiler/architecture inability to deal with the expression:

      sum += (v < 128) ? v : 256 - v;

Just eliminating the 'if (sum > lmins)' test doesn't make much difference; less than 0.5s.  (Notice that the test is unnecessary; overflow is now handled so it is only worth doing as an optimzation!)  The problem is that the code without the 'sum' calculation introduces many instructions into the loop.  Without it the inlined version is a minimal byte subtract/copy; there are 6 instructions in the whole loop, including 3 for the loop counter/test.  With the sum calculation there are 16 instructions, with the same three instruction overhead, and, curiously, the copy part is compiled with 4 instructions not 3.  The sum calculation takes 7 instructions and the test 2.
j
Ok, so the time increase is 3.66 and the instruction count increase is 16/6; 2.66.  This is less even though not all 16 instructions are executed each time round.  I believe the problem is either that the CPU has to wait for the ALU pipeline to do the check or that the CPU branch prediction fails causing a wait.

It is relative easy to rewrite the expression to avoid tests, either by using signed arithmetic and abs() or unsigned arithmetic.  The figures this gives, in seconds, are:

291: 79.8s +/- 0.1s
ed5: 84.1s +/- 0.5s  [+4.3s +/-0.5s]
ed5-inline: 79.7s +/- 0.1s [no change]
     This is just ed5 compiled -O2 -finline-functions.  The text size of pngcp increases to 69kbytes (+11kbytes/+18%).
ed5-noif: 84.0s +/- 0.2s [+4.2s +/- 0.3s]
     In this and the following figures I removed the test on lmins and the break (leaving lmins unused).  Notice that this is the original suggestion; so far as I can see it doesn't may much change!
ed5-abs: 82.2s +/- 0.1s [+2.4s +/-0.1s]
     I changed the sum += calculation to use "128 - abs(v-128)"; note that v is signed.
ed5-bitx: 81.6s +/- 0.1s [+1.8s +/- 0.1s]
     I changed 'v' to be (unsigned int) and used the following bit manipulation:
              unsigned int q = 0U - (v >> 7);   /* v < 128 ? 0 : -1 */
              sum += (0x7F & (v ^ q)) - q;

The 'bitx' approach is really just one of the standard 'abs' implementations with optimization of the operand.  It's a pity GCC fails to detect this optimization.

Based on this and the fact that 'abs' is a function that is normally inlined, and normally inlined well (the GCC version isn't too bad) I think it's reasonable to use this approach - better than the hard coded v < 128 implementation currently used.  I experimented with adding the test on lmins back in to both the bitx and abs variants.  This showed the following results:

ed5-abs-if: 82.4s +/- 0.1s [extra 0.2s over no if]
ed5-bitx-if: 82.3s +/- 0.2s [extra 0.7s over no if]

So the branch cost is the difference between the GCC 'abs' and 'bitx' compilations and putting in the extra branch makes no difference.  All the same, using abs increases the performance of the normal code (where the sum is required).

As a result I propose using 'abs' but including the early-out.  As discussed any disadvantages are only likely to apply to the single-filter case and this is rare.  I'll prepare and test a patch with the same changed applied to all the 'setup' functions; I'll be able to test what, if anything, this does to the normal libpng case where png_set_filter is not called and all the filters are checked.

I can abstract the sum heuristic as a macro which can be redefined if required.

jbowler:
These changes are here:

https://github.com/jbowler/libpng/tree/20160627-pngcp

16f25a is the base change above with 'sum' calculation moved to a macro,
there's a macro in there to disable the early-out, but as suggested
previously it isn't present in;

6a3797 is based on the libpng 1.7 approach and calculates the sum
separately, then just runs the selected filter.  It still writes a separate
(third) 'filtered' row but I think that is unnecessary; either row_buf or
prev_row can be overwritten, avoiding the need to allocate a third row
buffer.

I'm testing these now, but this will take some time; tests with the
heuristic take about 200s to run each and I want to run 10 of each
possibility to check for statistical significance in any results.
​
John Bowler

glennrp:
Note that the abs() calls were replaced with ternary statements in 1998 (see CHANGES)

Version 1.0.1a [April 21, 1998]
  Optimized Paeth calculations by replacing abs() function calls with intrinsics
  plus other loop optimizations. Improves avg decoding speed by about 20%.

Evidently things have changed in the compiler world since then.  The PNG_USE_ABS
macro has always been available to switch back to using abs() but it's undefined by default.

jbowler:
Yep.  Optimizing for GCC has always been a mistake.

jbowler:
65f4392 removes the need for the third row buffer (so only row_buf and prev_row are ever needed) and includes some optimizations to deal with the consequences.  The code size with O2 is, curiously, exactly the same as the current HEAD of the official libpng16.

I'm having difficulty getting statistically significant results; even though I use /usr/bin/time with %U (for just the user time of the process) the results seem highly dependent on, I think, file access timing.

All the same, 65f allocates half the number of row buffers as the current HEAD and gets times that are certainly about the same.

jbowler:
I think I need to check on ARM v8 (well, RPi3) too.  Even in 1998
(actually, from the dawn of ARM, 1986) abs() of an integer expression just
computed was one instruction, when correctly compiled; RSBMI result, arg,
#0; if (flags say minus) result := 0 - arg.

Since the heuristic depends so heavily on an operation conditional on
'minus' there really shouldn't be much change on ARM with a good compiler.

Gresham wrote a law for compilers.
​

glennrp:
Comparing pngcrush built with libpng-1.6.23 with and without PNG_USE_ABS produces imperceptible difference in the time to decode an all-PAETH image.  gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4

jbowler:
Yes; same here on the Paeth encode side (i.e. I can't get statistically
significant results testing filter=paeth; I'm really only getting something
that is probably significant with filter=sub!)

I think the only reason this shows up is because the png_setup_sub_row loop
comes down to just this (leaving out the casts to avoid undefined behavior
from clearing a sign bit):

for (i=bpp; i < row_bytes; ++i) {
   png_byte filtered = 0xFFU & (rp[i] - rp[i-bpp]));
   dp[i] = filtered;
   PNG_FILTER_HEURISTIC_SUM(sum, filtered);
}

Then the 'abs' implementation of the heuristic is twice as fast, on an
x86_64, as the (filtered < 128 ...) version with GCC 4.8.5.

Of course that is half of a very small number; 6 x86_64 instructions as
opposed to 13 (or there-abouts), so it disappears in the noise unless the
test case is set up to be really, really special.

The real argument for using abs() is that it gives the compiler a piece of
code that, with any luck, it is good at compiling; it may not always make a
difference but, absent pre-ANSI function level C compilers, it should not
make things worse.

I'm still trying to get statistically significant results.  I have an ARM
build now and it's running on the same basic CPU as the Nexus machine, but
it runs in 32-bit mode.  I don't know if the Nexus was operating in 32-bit
or 64-bit mode and it is said that the AArch64 64-bit instruction set
removes the conditional instructions (or most of them), so it would be
interesting to know if the Nexus is 64 or 32-bit.

I'm left wondering whether the whole thing comes down to bogus compilation
of that particular loop with GCC 4.6 and 4.8.5, but until I can reliably
get statistically significant results (I think I need to use the
clock_gettime approach from timepng) I'm unwilling to run hundreds of tests
on different compiler versions.

John Bowler

mattsarett:
Sorry for the delayed response!

> I used GCC 4.8.5 and a minimally configure libpng, producing (at -O2) a statically linked pngcp of text size 58kbytes on X86-64. I used Glenn's hand duplicated 291xxx revision and the 1.6 ed5xxx tip revision.
>
> ...
> 
> So far as png_write_find_filter is concerned, however, it took 1.79% of the time in 291 and the non-inlined function png_setup_sub_row took the equivalent of 6.56% of the samples in ed5 (png_write_find_filter did not appear in ed5). The difference, 4.79%, accounts for the increased samples in ed5 even though other sample counts moved around quite a lot in other functions; I assume that this is simply a result of the change in png_write_find_filter changing where cache waits happen. The time increase is 6.56/1.79 = 3.66.

Thanks for your time and effort in reproducing this.  I'm a little confused about why the paeth filter shows up on the profile when you are specifying filter=sub, but I understand the general results.

Just a guess as to why your results seem to be more zlib dominated than ours: we are compressing web graphics which should have more identical pixels and be easier to handle than a photograph (for example).

> Just eliminating the 'if (sum > lmins)' test doesn't make much difference; less than 0.5s. (Notice that the test is unnecessary; overflow is now handled so it is only worth doing as an optimzation!) The problem is that the code without the 'sum' calculation introduces many instructions into the loop. Without it the inlined version is a minimal byte subtract/copy; there are 6 instructions in the whole loop, including 3 for the loop counter/test. With the sum calculation there are 16 instructions, with the same three instruction overhead, and, curiously, the copy part is compiled with 4 instructions not 3. The sum calculation takes 7 instructions and the test 2.

Yes absolutely!  I think it'd be great to get those extra instructions out of the loop.

----------------------------------------------------------------------------------------

I was actually able to eliminate the regressions using -O2 and -finline-functions (function gets inlined, sum calculation get optimized out), but it turns out that isn't an acceptable fix for our platform.  Sorry for having misspoken earlier.

I'm currently awaiting performance results of using PNG_USE_ABS.  I'm hopeful it will help, but not so confident that it will completely solve the problem.

John, I've also been reading through your recent commits to jbowler/libpng.  Looks like https://github.com/jbowler/libpng/commit/6a3797ae7dd79419b56e18e2494fe204dcfd5f76 would solve the regression?  Is this going to be merged to libpng16?

jbowler:
Paeth shows up in all the tests on the *read* side because the input files
(which are web graphics, sampled from Googles image search a while back)
use Paeth.

My testing covers 'all' filters, the 'fast' subset (none,sub,up) and my
'superfast' subset (none,sub) together with each filter individually.  I
don't believe other combos are worth testing in 1.6 because, in practice,
no one will select them and the results are showing that obvious
optimizations can have bad effects on the 'all' case; that case is the
primary one to optimize because it is the default!

My tests take some time because I'm running them with the default zlib
settings, with strategy/filtered, level/3 (which is a recommended high
speed setting) and with strategy/RLE level/1 (which is the fastest and
actually quite good for read speed.)  I'm also running on a high speed
x86_64 box and on an ARMv8 (RPi3) running in ARMv7a emulation mode; I don't
know whether the Nexus platform is AArch 64-bit or ARM 32-bit; is it
possible to reveal that?  I've got no way of doing 64-bit testing unless
the RPi foundation release a 64-bit kernel/user-space and I suspect they
won't do that because, quite frankly, running 64-bit, even on a Nexus,
seems very likely to be a mistake.

The executive summary is that I can't repro anything like a 30% change
unless I compress really really big (ridiculously big) images; 4000+ pixels
per row.  On those images the timing figures flip from a significant
inter-dependence between the filter code and the deflate code to one where
the deflate time and the row-filter time seem independent.  Reducing the
time in deflate (via the RLE/1 settings) then starts to show timing figures
in the write side (specifically just measuring the time for png_write_png
to complete) that show up to about 15% differences.

The positive effect of abs() is restricted to the normal sized images, and
is not consistent.  With the large images abs results in a slow-down with
either GCC 4.8.5 or 5.4.0 on both platforms.  I don't understand why this
happens on ARM, maybe the v7a core no longer supports RSBMI, or maybe GCC
still generates bad code.  After all the x86_64 code looks pretty weird too.

The one good thing is that I can halve the dynamic memory requirement of a
'streamed' write where the row is sufficiently large without much affecting
the time.  Worst case is 10%, and I think that is because my new code is
using abs().  (I'm going to change it to handle PNG_USE_ABS).

As with any optimization doing it the first time requires a lot of
measurements.

John Bowler

>

--

mattsarett:
> Paeth shows up in all the tests on the *read* side because the input files
> (which are web graphics, sampled from Googles image search a while back)
> use Paeth.

Ahh thanks!

> My tests take some time because I'm running them with the default zlib
> settings, with strategy/filtered, level/3 (which is a recommended high
> speed setting) and with strategy/RLE level/1 (which is the fastest and
> actually quite good for read speed.)  I'm also running on a high speed
> x86_64 box and on an ARMv8 (RPi3) running in ARMv7a emulation mode; I don't
> know whether the Nexus platform is AArch 64-bit or ARM 32-bit; is it
> possible to reveal that?  I've got no way of doing 64-bit testing unless
> the RPi foundation release a 64-bit kernel/user-space and I suspect they
> won't do that because, quite frankly, running 64-bit, even on a Nexus,
> seems very likely to be a mistake.

FWIW, the Nexus 5X and 6P are AArch64.

> The executive summary is that I can't repro anything like a 30% change
> unless I compress really really big (ridiculously big) images; 4000+ pixels
> per row.  On those images the timing figures flip from a significant
> inter-dependence between the filter code and the deflate code to one where
> the deflate time and the row-filter time seem independent.  Reducing the
> time in deflate (via the RLE/1 settings) then starts to show timing figures
> in the write side (specifically just measuring the time for png_write_png
> to complete) that show up to about 15% differences.

Yes, I'd agree that the 30% change is an anomaly related to a specific test case.  And the performance is highly variable across image size/content/filter settings/zlib setting etc.  But I think we have identified to cause of the change.

> As with any optimization doing it the first time requires a lot of
> measurements.

Yes thanks for your efforts on this.  To follow up on my above comment, unfortunately the PNG_USE_ABS change hasn't impacted any of my tests.  Any news on integrating jbowler@6a3797a?

glennrp:
See commit f7d5419816d85d0b996c403e788bb707b552524f

mattsarett:
Woohoo, thanks Glenn!

PR #117, opened 2016-06-20
More efficient absolute value on SSE2

mattsarett:


mattsarett:
On behalf of mayeut from:
https://github.com/glennrp/libpng/pull/88#issuecomment-227030633

PR #131, opened 2016-08-29
Added build support for MIPS MSA

mandar-sahastrabuddhe:
Added build support for MIPS MSA with one msa optimized function: png_read_filter_row_up_msa


glennrp:
This is included in libpng-1.6.25rc03.  I don't have access to a MIPS platform to test it though.

PR #140, opened 2016-10-01
Unsigned overflow

jbowler:
Remove all currently detected cases of unsigned overflow.  Detection is
runtime, so test case dependent.  The changes to pngvalid.c eliminate
spurious and probably invalid tests with one while loop exception.

Apart from that and the change to the dependence on the intended
unsigned overflow in pngtrans.c the changes are limited to altering the
meme for an unsigned 'x' from:

   while (x-- > 0)

to

   for (; x > 0; --x)

This works because, in all cases, the control variable is not used in
the loop.  The 'while' meme was, at one time, warn'ed by GCC so it is
probably a good change, for some weird religious value of good.

### DIFF
--- a/contrib/libtests/pngunknown.c
+++ b/contrib/libtests/pngunknown.c
@@ -478,7 +478,7 @@ get_valid(display *d, png_infop info_ptr)
       png_textp text;
       png_uint_32 ntext = png_get_text(d->png_ptr, info_ptr, &text, NULL);
 
-      while (ntext-- > 0) switch (text[ntext].compression)
+      while (ntext > 0) switch (text[--ntext].compression)
       {
          case -1:
             flags |= PNG_INFO_tEXt;
--- a/contrib/libtests/pngvalid.c
+++ b/contrib/libtests/pngvalid.c
@@ -1242,7 +1242,7 @@ store_image_check(const png_store* ps, png_const_structp pp, int iImage)
 
       image += 2; /* skip image first row markers */
 
-      while (rows-- > 0)
+      for (; rows > 0; --rows)
       {
          if (image[-2] != 190 || image[-1] != 239)
             png_error(pp, "row start overwritten");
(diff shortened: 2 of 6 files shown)


glennrp:
Your pull request came in after I pushed a fix.  I had kept the "while (y-- > 0"  statements but added 1 to
the initialization of the various "y" and changed the statments to "while (y-- > 1)".  One of the loop variable (ntext) was being used.

jbowler:
Um, where did those come from?  They weren't there when I submitted the pull request.  (This from gentoo; Qupzilla, which works quite well.)

glennrp:
While (ntext-- > 0) is in pngunknown.c

jbowler:
I don't think you can guarantee ab-initio that 'y' is not 0xFFFFFFFF.

I did think of changing it to just put the --y after the {, but once I had looked at the loop body it was obvious (aka /y/) that the control variable wasn't used so the for loop expansion seemed more correct; the code does not use the value of 'y-1' when 'y' is 0, and must not.

It's clearer in pngvalid.c 2425 where, irc, the variable is 'rows'.  Really 'y' should be 'h' in all the simplified API cases; obfuscated code on my part.

glennrp:
OK.  I don't mind switching from while() to for() anyhow.  Just take care with ntext.

glennrp:
Height is limited to 7FFFFFF so y+1 will be OK.

jbowler:
I commented the ntext case.

I did review all of the changes to double check that the control variable was not used after the loop.

jbowler:
The thing about 'x+1' is that it is not self-evident that it won't overflow, so using the meme to prevent an overflow in another place seems wrong.   The while loops (both the ntext cases and the y cases) were being lazy; relying on the post-decrement value not being used, so regardless of my objection to the original report not being, itself, ISO-C, it seems reasonable to adopt a meme where the overflow cannot happen.

The for loop is self evidently correct even if the control variable is used in the loop.  The only change is to the use after the loop and, yes, 'y' cannot be 0 (I'm not so sure about ntext), but if y, or ntext, is 0, surely the code after the loop must be bogus?   (Not wrong, bogus; it's subtly different.)

Oh, and I hate it when people fail to put the dot over an i, unless they are Turkish of course.

PR #141, opened 2016-10-28
added PNG_ARM_NEON switch and enabled building with NEON support with CMake

andy-held:
CMAKE_SYSTEM_PROCESSOR is checked to discern whether to build for arm or not, as it is the canonical variable for this.
Behaviour is copied from autoconf; by default PNG_ARM_NEON is set to 'check' and NEON support is checked dynamically at runtime. Other possible values are 'on' and 'off'. The deprecated option 'api' is not used here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,30 @@ option(PNGARG        "Disable ANSI-C prototypes" OFF)
 set(PNG_PREFIX "" CACHE STRING "Prefix to add to the API function names")
 set(DFA_XTRA "" CACHE FILEPATH "File containing extra configuration settings")
 
+# set definitions and sources for arm
+if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "^arm")
+  set(PNG_ARM_NEON_POSSIBLE_VALUES check on off)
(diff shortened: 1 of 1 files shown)


jbowler:
Regardless of the cmake stuff there seems to be some misunderstanding here, from the comment (the change matches this):

"Behaviour is copied from autoconf; by default PNG_ARM_NEON is set to 'check' and NEON support is checked dynamically at runtime. Other possible values are 'on' and 'off'. The deprecated option 'api' is not used here."

The _default_ is, and must be, to not set PNG_ARM_NEON.

It is the 'check' option which is deprecated!  It only works on a subset of GNU/Linux operating systems, so enabling it by default will break libpng on pretty much everything except the slew of IoT devices
that are not secure.

In any case the only reason --enable-arm-neon still exists in 1.5 and 1.6 (it isn't in 1.7) is that it was a part of the original, very broken, implementation.  All it does is set certain build configuration CFLAGS  however, as was pointed out soon after people started moving from 1.4, this can't be done at configuration time; it has to be done automatically at compile time.  This is why the default must be 'unset' (see the test in pngpriv.h).

Libpng 1.6 and, I think 1.5, automatically handle the ARM NEON optimizations.  I'm not aware of any situation where this doesn't work though as with any configurable item of libpng people might want to
change things; this is done by setting CFLAGS when pnglibconf.h is built.

andy-held:
Oh, I see.
Evidently I did not understand this correctly.

PR #143, opened 2016-11-11
added sources under 'arm/' to build when building for arm with CMake

andy-held:
CMAKE_SYSTEM_PROCESSOR is checked to discern whether to build for arm or not, as it is the canonical variable for this.
Behaviour is copied from autoconf; by default PNG_ARM_NEON is not set. If it is manually set to 'api', 'on' or 'off' the same defines as in using autotools are defined. The deprecated 'check' option is not supported here.


glennrp:
I think this has been completed in the libpng16 and libpng17 branches.

PR #147, opened 2016-12-27
Remove use of 'defined' within a macro expansion

jbowler:
This contains the libpng15 change to pngvalid.c plus removal of an additional pair of 'defined within macro' cases that only exists in 1.7

pngvalid.c is a copy of 1.5; this contains the copyright change you (so far as I can see) made a while back in 1.5 and 1.6.

At this point clang unsigned overflow sanity checking will not pass on 1.7; clang seems to object to any expression which converts a negative integer to unsigned.  Adding 'U' to some constants will probably fix it but I'm not going to even try until I've merged my development branch back in.


glennrp:
Not sure why this is still open.  See https://github.com/glennrp/libpng/commit/8d4110bd618eb03a252b18456c546cef0a8475bb#diff-0e5c6c7610a0e71539bb73bab1bdefc0

PR #151, opened 2017-01-29
PowerPC VSX optimized filter functions

### LINKED ISSUE #136: PowerPC Power8 VSX SIMD optimized filter functions
Implement SIMD vector support for Power8 VSX SIMD equivalent to libpng support for SIMD optimizations for Intel SSE, ARM Neon, and MIPS MSA.  Achieve speedup appropriate for PowerPC VSX vector width and processor pipeline.

barkovv:
@glennrp
Resolves #136 
It can be compiled with gcc or clang. Optimization speedup is measured by gprof with size of row 1024 kB.
#### Status of work (both for bpp=4 and bpp=3)
- [x] filter_up (spedup x5.6)
- [x] filter_sub (speedup x1.5)
- [x] filter_avg (speedup x2.8)
- [x] filter_paeth (speedup x1.08)

Now code is for POWER7 or later due to vec_vsx_ld function. I can try to find analog for earlier CPU's if needed (but I think that optimize code for Power G5 from 2004 is a worthless idea in 2017)


barkovv:
@glennrp
Code is ready. Please review and merge it. It is for POWER8 or later due to vec_vsx_ld function. I can try to find analog for earlier CPU's if needed (but I think that optimize code for Power G5 from 2004 is a worthless idea in 2017)

jbowler:
I should add that my position is that using configure to enable or disable hardware optimizations is fundamentally broken; it breaks multilib builds because those have to decide at compile time what to build.  The ARM optimizations only have the configure option to support older build scripts that use it; deleting it requires a major version change.

barkovv:
@jbowler thanks for you remark about adding "deprecated" code. I will remove this phrase from comment but I can't imagine more reliable way to detect architecture than reading /proc/cpuinfo.

barkovv:
@jbowler I've change default runtime CPU checking method to auxv (it requers >=2.6 kernel, is it okay?)

barkovv:
@jbowler jbowler It's ready to review (again).

barkovv:
@jbowler could you review it ?

jbowler:
You changed it to use auxv, which is even less portable than /proc/cpuinfo
and is only nominally supported by Android; if you try to build code using
auxv (at least in the ARM variant) it doesn't even link.

You also deleted the explanation of the option, which is, I guess, a good
way of stopping people using it; I was pointing out that it wasn't
deprecated because you were supporting it!

My position was and still is that doing this in configure (or, for that
matter, cmake, or the makefiles) is flawed because it cannot possibly work
in a multilib build and anyway the approach isn't portable (show me a
Windows (!Ubuntu) version, show me a BSD version), and it creates a never
ending stream of bugs which no one understands (see the history of the ARM
attempts).

That doesn't mean you can't do it, but you have to work out a response for
when the bugs come from people building on Windows or BSD or, indeed,
Solaris; a system which supports PowerPC (it was designed for it) but which
doesn't have /proc/cpuinfo.

John Bowler
​

barkovv:
@jbowler What should I do? Shall I add comment about advicing not using this internal check code? Or add support for *BSD, OSX detection ?

barkovv:
@jbowler I added warning about list of runtime detection supported OSes. Issue starter said he wants only linux support (which is implemented). When compiling, developer will receive warning and check that list in README. If he will need this feature on other OS he should implement it himself. Is it acceptable solution?

jbowler:
That seems reasonable.  So long as the default is to use the compiler flags
and there is some warning there it should raise minimal problems; the ARM
stuff had particular problems because the default didn't work on most
systems and there was no documentation.

John Bowler
​

barkovv:
@jbowler glennrp 
 So if this it ready to merge or I should do anything else?

jbowler:
I think Glenn is waiting for testing from someone who actually has a
PowerPC build, I'm not aware of anything else blocking it, but I don't know
how Glenn wants to merge it or if he has issues.​

John Bowler

glennrp:
I requested, months ago, testing of the MIPS-MSA code that was donated.
I have not heard a peep from anyone, and am considering removing it.

Glenn

barkovv:
@glennrp What kind of testing is needed? Am I capable to do that or it must be done by someone else?

jbowler:
The MIPS code requires GCC 7.0, while this is the "current" GCC Gentoo doesn't even have it yet as an unstable pacakge.   The code won't get independently tested until GCC 7.0 is widely available.

*This* code should  be supported on any recent version of GCC; I checked and the Altivec support is there in 4.8.4 (which is what my Ubuntu uses, my dev system uses 5.4.0 by default; 6.3.0 isn't supported yet on x86_64).

Unfortunately I tried installing powerpc 5.4.0 and gcc crashed while compiling *itself*, so I would probably have to got back to 4.8.5 or 4.9.3 or forward to 6.3.0.  It's difficult with cross compilers working out what works in advance.

barkovv:
@jbowler 
I can provide you access to Power8 little endian virtual machine via SSH (hosted on osuosl.org). Please give me your RSA public key (email me on neverscaired@gmail.com or via Github).

jbowler:
I can run tests on that only if it is UNIX-like and has gcc, bash, awk
(the-one-true-awk is sufficient) and an *installed* zlib (I mean the
headers and a linkable library; it doesn't much matter which version).
Some tests use POSIX specific code but IRC they compile without it, it just
removes the capability to do any timing tests.

If that works, how many cores does it have, and do you mind if I use them
all?  I'll send you the key separately.  I've been unable to build a GCC
cross-compiler regardless of version; the compiler always crashes compiling
itself somewhere in handling the Gentoo linux headers.

John Bowler

barkovv:
@jbowler @glennrp 
Good news. I've fixed some bugs and now all of functions pass all tests on ppc64le. Now I'll will test code on ppc64 big endian and report results.

barkovv:
@jbowler This virtual machine has Debian on board so it has all common Linux tools you probably need to test with. You can provide me your public RSA key for ssh login if you want to test this code (neverscaired@gmail.com).

jbowler:
I sent you the key yesterday; it's in the attachment (you responded to the
email, albeit a response intended for Glenn).

John Bowler
​

barkovv:
@jbowler @glennrp ppc64 big endian tests pass too. Now it is ready to merge completely. Waiting for your testing.

jbowler:
1) The LE build does not compile when I test it; I think somewhere something is using a non-C90 structure; I think it's a literal array {x,y,z}, which is not something I've seen before, but it is in a macro expansion so something else may be happening.  The build log is here:

debian@dlib-debian-le-1:~/jcb/src/libpng/barkovv/build/powerpc64le-linux-gcc-6.3.0+vsx=on/make.out

2) I only tested gcc 6.3.0; the gcc 4.9.2 installed on that system (and the g++ 4.9.2) are, I assume, not the right compiler and, anyway, a basic build of an unmodified libpng 4.9.2 fails because of strict-overflow problems in that version of GCC (connected to warning messages being produced as a result of O2 optimizations).

3) I didn't test with the "correct" compiler flags and the default builds (which all pass) of the modified libpng16 all pass because the VSX code does not get built.  I need to research the compiler flags again, but I assume that if I get the flags right all the builds will fail because of the non-ANSI code.

jbowler:
There is a test script that anyone can use on the test system.  I made hacks to support powerpc and the weird way the compilers are installed ("gcc-6", which is fine, but there is no "gcc-6.3.0").  The script is:

~debian/jcb/bin/regen

to use it:

PATH=~debian/jcb/bin:$PATH export PATH

then clone the github repo in a directory called "code", mkdir a sister directory called "build" and;

cd code
autoreconf -f -i
cd ../build
regen -f powerpc64le-linux-gcc-6.3.0+\*

and that will build each of about 38 build configurations 2 at a time (the machine only has on CPU).  The arguments to regen are a list of the builds to build, it creates identically named directories.  The primary log file is make.out, everything else is normal build stuff (config.out etc).

barkovv:
@jbowler Did you provied -mvsx -maltivec flags during testing? Building on my bigendian machine failes without it, so I guess you get this missing too.

barkovv:
@jbowler Problem is fixed, please continue testing if it is needed.

jbowler:
I didn't provide those flags; that is my next text, to ensure that the
correct code gets compiled with the VSX option turned on at the compiler
level.  (This is the requirement for a successful multilib build).

John Bow;ler
​

significant-bit:
I'd be glad to test this on Power7 big endian.

Just git clone, configure, and make?

What is a good way to see the performance difference?

barkovv:
@significant-bit
```
git clone ...
./autogen.sh
./configure --enable-powerpc-vsx=yes
make
#if build failes
make CFLAGS="-mvsx -maltivec"
```

Code compiles successful on Power8le Debian machine without theese flags. But when I compile code on my qemu emulator (Debian Power8 PPCbe), it fails to compile and require to provide "-maltivec -mvsx".

jbowler:
On Sat, Feb 11, 2017 at 4:26 PM, Mike Erwin <notifications@github.com>
wrote:

> I'd be glad to test this on Power7 big endian.
>
> Just git clone, configure, and make?
>
This *SHOULD* work but doesn't, it looks like the autodetection of VSX is
not functioning:

git clone --branch=libpng16 --depth=1 https://github.com/barkovv/libpng.git
barkovv
cd barkovv
autoreconf -f -i
CC="gcc -mvsx -maltivec" ./configure
make check

Choose the CC definition so you get the compiler you want running with the
machine options you need.

This should pass (though I only just started testing the LE variant).  If
it does it will generate a program timepng in the build directory.  You
then run something like:

timepng --assemble bigfile png1.png png2.png png3.png ...
timepng --dissemble bigfile <count>

Where <count> is the number of files you put into "bigfile" with the first
command.  It times the decode.  You can put every PNG in your internet
browser cache in to get a random sample of typical PNG files on the
internet (well, typical of the www pages you visit.)  timepng outputs a
help message.

jbowler:
I'm at the version you pulled into jcb/src/libpng/barkovv/code, i.e. this revision:

a343882c74ecde75b119072cae766ace88c6beac

And with vsx=on I am now getting 0 sized objects from "size powerpc/*.o".  l also couldn't see any VSX related options in pnglibconf.h (the ARM_NEON ones are there, but nothing for VSX).  There's nothing for MIPS either.  The common bug there is that the required changes are missing from scripts/pnglibconf.dfa; search for "NEON".

That's not a complete explanation though because I expect to see PNG_POWERPC_VSX_OPT in a gcc command line somewhere in 'make.out' and I don't see it.  The checks in pngpriv.h that set PNG_POWERPC_VSX_OPT are wrong; I hacked PNG_POWERPC_VSX_OPT into pnglibconf.h (with value 2) and the powerpc object modules started to have code in them (and the tests passed).

I then tried this:

gcc-6.3.0 -mvsx -maltivec -E - <<EOF
__ppc64__
__ALTIVEC__
__VSX__
EOF

and I got this:

# 1 "<stdin>"
# 1 "<built-in>"
# 1 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 1 "<command-line>" 2
# 1 "<stdin>"
__ppc64__
1
1

I think that speaks for itself.

jbowler:
The bug is it should be uppercase, i.e:

`__PPC64__`

barkovv:
@jbowler Great thanks for finding that. Fixed.

barkovv:
@jbowler I've added some similar to ARM code to scrpits/ files. Now pnglibconfig.h is correct.

jbowler:
My previous comment about the "check" option was incorrect; I was getting
mixed up with the regular (no-option) option; I suggest just reverting to
the ARM/MIPS comment.

jbowler:
I completed some timepng based speed tests last night.  Analysis of the results suggests a 1/74% speed up from using the Altivec code; this is suspiciously large so other tests would be valuable.  The test methodology is as I described before.  I used contrib/testpngs/*.png for the input, producing an assembly with 104 files and a bigger one by adding the same 104 files 100 times:

```
debian@dlib-debian-le-1:~/jcb/src/libpng$ ls
barkovv  big.10400.ass  libpng16  test.104.ass  timepng.out
```

The file "timepng.out" is the result of running two "release" builds of the libpng16 and barkovv git heads 100 times each (I used a faked release build to eliminate non-release tests and warnings).  The result (in seconds), expressed as average +/- standard deviation, is:

```
jbowler@PETUNIA:~$ ./timepng.awk timepng.out
barkovv 2.36439 +/- 0.00553637
libpng16 3.17475 +/- 0.00700532
barkovv/libpng16:  74.475%
```

Note that these figures are the real process time used to decode those 104 PNG files and the PNG files represent all of the PNG formats; so the selection is not representative of frequency of occurrence of formats.  The files that hit the optimizations are also quite large and very regular; 256x256 pixels in the RGB cases and this will give a false over-estimate of the speed up.  Nevertheless the test is a real-world test, so apart from my doubts about the quality of test data the speed up seems significant.

As always I qualify these comments with the observation; in software engineering you can prove *anything* by using the right data.

barkovv:
@jbowler it seems that 3644db2 comment doesn't make any sense. I can revert it ofcourse (if it is not needed). Are some aditional tests going to be done?

jbowler:
On Sun, Feb 12, 2017 at 3:23 PM, Vadim Barkov <notifications@github.com>
wrote:

> @jbowler <https://github.com/jbowler> it seems that 3644db2
> <https://github.com/glennrp/libpng/commit/3644db298b3eb02cf23c6b6578fd2b38dd2871c3>
> comment doesn't make any sense.
>

Well, I think I wrote the original (ARM) version and it makes sense to me
but it may benefit from some editing to make it clear; it is documenting
[PNG_]ARM_NEON_OPT.  *Normally* this is not set to anything and it gets
defaulted in pngpriv.h.

Setting it to 0 via -D, or via configure (though I don't like the use of
configure) is useful to disable the machine code.

Setting it to 1 should not be done because it happens automatically if
either ARM_NEON_API or ARM_NEON_CHECK (well,
PNG_ARM_NEON_API/CHECK_SUPPORTED) is set (defined).  This happens when
pnglibconf.h is built.

Setting it to 2 shouldn't be done because pngpriv.h does this automatically.

In the original implementations the people providing the code required the
person who built libpng to get everything right; setting the required
#defines *and* getting the compiler flags right.  This was complex and
resulted in bug reports because, as you seem to have found, sometimes it is
possible to get code like this to compile without adding special compiler
flags (PowerPC64LE) and sometimes it isn't (ARM works in a cross-build
without -mfpu=neon but does not in a host build, PowerPC64BE apparently
doesn't work.)

This is why I added the "inadvisable" comment about setting "2".

-- 
+1 (541) 450-9885
PO BOX 3151
KERBY OR 97531-3151
USA

barkovv:
@jbowler  okay, I get it. would you lile to continue testing?

jbowler:
Well, I will continue testing.  Whether I like it, I don't know.​

jbowler:
Final results: about 8% improvement in (just) decode time of typical PNGs; remember this is just the decode time, it doesn't include any modifications or use of the resultant bitmap data.

I rebuilt everything and retested with a better sample set; 1000 randomly sampled files.  I've described this set before on the png-mng-implement mailing list.  It reflects a static snapshot of web pages spidered by Google; I've previously published the issues with this approach to sampling.

I observe that the provided test machine has compilers (both of them) which have -maltivec -mvsx built in as the target defaults, so my tests reveal nothing about the behavior with respect to the same code compiled *without* VSX; it may be faster or slower, I don't know.

So, on a system which is built *with* the VSX hardware switched on and with GCC 6.3.0 the time to decode my 1000 files is 0.480 seconds.  When the machine code is used, however, the time is reduced to 0.443 seconds; an 8% reduction (speed up).

I also tested a complete set of my build configurations and all passed.

barkovv:
@glennrp
Can this PR be merged now?

glennrp:
Merged into libpng16, thanks.

glennrp:
Here are my results for Earth10k.png with an instrumeted pngcrush.
Note that there is some speedup, but a much larger payoff is obtained
by running "pngcrush -speed" on the file to avoid using the AVG and PAETH
filters.

Glenn

Timing tests for firefox linux earth10k (not optimized)
Linux dlib-debian-le-1 4.9.0-1-powerpc64le #1
  SMP Debian 4.9.6-3 (2017-01-28) ppc64le GNU/Linux
gcc (Debian 4.9.2-10) 4.9.2
Earth10k.png      CPU time decode 2.633772, total 2.801144 sec
Earth10k_fast.png CPU time decode 1.898273, total 2.006459 sec
Earth10k_slow.png CPU time decode 2.601996, total 2.712496 sec

Timing tests for firefox linux earth10k (powerpc-vsx optimized)
Linux dlib-debian-le-1 4.9.0-1-powerpc64le #1
   SMP Debian 4.9.6-3 (2017-01-28) ppc64le GNU/Linux
gcc (Debian 4.9.2-10) 4.9.2
Earth10k.png      CPU time decode 2.458045, total 2.623693 sec
Earth10k_fast.png CPU time decode 1.841627, total 1.947361 sec
Earth10k_slow.png CPU time decode 2.434269, total 2.544447 sec

-rw-r--r-- 1 debian debian 72360235 Feb 22 16:53 Earth10k.png
-rw-r--r-- 1 debian debian 63285602 Feb 22 18:03 Earth10k_fast.png
-rw-r--r-- 1 debian debian 66843433 Feb 22 18:13 Earth10k_slow.png

Earth10k.png was provided by debian
Earth10k_fast is from "pngcrush -brute -speed -force Earth10k.png"
Earth10k_slow is from "pngcrush -force Earth10k.png"

barkovv:
@glennrp Could you test Intel in similar way please?

jbowler:
On Wed, Feb 22, 2017 at 10:41 AM, Glenn Randers-Pehrson <

> Here are my results for Earth10k.png with an instrumeted pngcrush.
> Note that there is some speedup, but a much larger payoff is obtained
> by running "pngcrush -speed" on the file to avoid using the AVG and PAETH
> filters.
>
Well, indeed.  From your numbers:

Fast decode: +6% file size, +24% speed improvement

The "right" answer depends on the usage scenario.  Someone might care about
that 6%, and not care about the 24% cost; only the user can know.

I guess one useful test is to compare the cost of Paeth vs Sub; i.e. encode
the same image with just Sub or just Paeth and see whether a performance
improvement substantially reduces the 24% cost.  All the same, even if a
hardware improvement does make Paeth less of a cost on decode, so what?  In
most applications when a PNG is authored the author does not know it will
be decoded on specific hardware.

-- 
+1 (541) 450-9885
PO BOX 3151
KERBY OR 97531-3151
USA

glennrp:
Timing tests for firefox linux earth10k on intel:

Linux studio 4.4.0-62-generic #83-Ubuntu SMP Wed Jan 18 14:10:15 UTC 2017
x86_64 x86_64 x86_64 GNU/Linux
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

pngcrush-1.8.12:
Earth10k.png      CPU time decode 2.880897, total 3.077706 sec
Earth10k_fast.png CPU time decode 2.317034, total 2.457742 sec
Earth10k_slow.png CPU time decode 2.994704, total 3.149903 sec

pngcrush_1.8.12_no_opt:
Earth10k.png      CPU time decode 3.266601, total 3.469199 sec
Earth10k_fast.png CPU time decode 2.359044, total 2.504768 sec
Earth10k_slow.png CPU time decode 3.355562, total 3.509613 sec

The speedup is slightly better but still nothing to write home about.

Glenn

barkovv:
@glennrp can you say that PowerPC and Intel optimizations are about equal by speed? I cannot understand results and how I should interpret it.

glennrp:
On Wed, Feb 22, 2017 at 2:29 PM, John Bowler <notifications@github.com>
wrote:

> On Wed, Feb 22, 2017 at 10:41 AM, Glenn Randers-Pehrson <
> notifications@github.com> wrote:
>
> > Here are my results for Earth10k.png with an instrumeted pngcrush.
> > Note that there is some speedup, but a much larger payoff is obtained
> > by running "pngcrush -speed" on the file to avoid using the AVG and PAETH
> > filters.
> >
> Well, indeed. From your numbers:
>
> Fast decode: +6% file size, +24% speed improvement
>
> The "right" answer depends on the usage scenario. Someone might care about
> that 6%, and not care about the 24% cost; only the user can know.
>

Well, it's +6% *improvement* on filesize, so it doesn't matter if they care.
I don't know why the "-speed" option resulted in a smaller Earth10k.png
file, but
there it is.

Glenn

glennrp:
Just based on Earth10k.png, the Intel optimization produced a slightly greater speedup
than the Powerpc optimization.  But neither improves the decoding speed enough
for anyone to notice it without instrumenting their code.

jbowler:
It isn't meaningful to compare hardware optimizations like this across different architectures, indeed, it isn't meaningful to compare hardware optimizations like this on the *same* CPU between *different* memory memory architectures or (even worse) different OS versions.

jbowler:
Glenn; I tested the assembly I had used before on the PowerPC system but using the Intel code and I couldn't get results that were statistically significant a the 90% level when I ran 100 test pairs, I had to run 1000 test pairs to get something that might be.  The results are greatly affected by machine load (even though timepng only measures process CPU time).  Are your results (on Intel) statistically significant?

pngcrush-1.8.12: Earth10k_slow.png CPU time decode 2.994704, total 3.149903 sec
pngcrush_1.8.12_no_opt: Earth10k_slow.png CPU time decode 3.355562, total 3.509613 sec

That's around a 12% speed up, but I saw the same and *THAT* was not statistically significant.

The PowerPC figures were significant; see my results above (the first test with 104 PNGs in the assembly), but that's a machine with, apparently, little load (it helped that I was running the tests at 3AM in the morning and I'm in the same state ;-)

John

glennrp:
My tests were on an otherwise idle workstation.  Each line actually
represented the smallest result for 5 runs (using pngcrush -bench 5).
I've just repeated the test six more times and the results look rock
solid:+

Timing tests for firefox linux earth10k

Linux studio 4.4.0-62-generic #83-Ubuntu SMP Wed Jan 18 14:10:15 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

pngcrush:
Earth10k.png      CPU time decode 2.882007,total 3.088203 sec
Earth10k.png      CPU time decode 2.881014,total 3.079939 sec
Earth10k.png      CPU time decode 2.880405,total 3.077050 sec
Earth10k.png      CPU time decode 2.881332,total 3.077632 sec
Earth10k.png      CPU time decode 2.880103,total 3.087936 sec

Earth10k_fast.png CPU time decode 2.316118,total 2.463541 sec
Earth10k_fast.png CPU time decode 2.316148,total 2.456554 sec
Earth10k_fast.png CPU time decode 2.315933,total 2.456326 sec
Earth10k_fast.png CPU time decode 2.315886,total 2.454998 sec
Earth10k_fast.png CPU time decode 2.316710,total 2.456888 sec

Earth10k_slow.png CPU time decode 2.994725,total 3.142210 sec
Earth10k_slow.png CPU time decode 3.001122,total 3.153591 sec
Earth10k_slow.png CPU time decode 2.994478,total 3.140198 sec
Earth10k_slow.png CPU time decode 2.995042,total 3.150825 sec
Earth10k_slow.png CPU time decode 2.994208,total 3.141076 sec

pngcrush_1812_no_opt:

Earth10k.png      CPU time decode 3.266963,total 3.473339 sec
Earth10k.png      CPU time decode 3.266716,total 3.474361 sec
Earth10k.png      CPU time decode 3.267167,total 3.473555 sec
Earth10k.png      CPU time decode 3.266481,total 3.473016 sec
Earth10k.png      CPU time decode 3.266488,total 3.481209 sec

Earth10k_fast.png CPU time decode 2.357406,total 2.512220 sec
Earth10k_fast.png CPU time decode 2.357070,total 2.512753 sec
Earth10k_fast.png CPU time decode 2.359321,total 2.506265 sec
Earth10k_fast.png CPU time decode 2.357918,total 2.512319 sec
Earth10k_fast.png CPU time decode 2.359186,total 2.505328 sec

Earth10k_slow.png CPU time decode 3.355757,total 3.511169 sec
Earth10k_slow.png CPU time decode 3.355778,total 3.510183 sec
Earth10k_slow.png CPU time decode 3.354734,total 3.507664 sec
Earth10k_slow.png CPU time decode 3.355413,total 3.508895 sec
Earth10k_slow.png CPU time decode 3.356333,total 3.509265 sec

Glenn

glennrp:
Also about my tests,

 | Executable name is pngcrush
 | It was built with   bundled libpng-1.6.29beta02
 | and is running with bundled libpng-1.6.29beta02
 |    Copyright (C) 1998-2004, 2006-2016 Glenn Randers-Pehrson,
 |    Copyright (C) 1996, 1997 Andreas Dilger,
 |    Copyright (C) 1995, Guy Eric Schalnat, Group 42 Inc.,
 | and bundled zlib-1.2.11, Copyright (C) 1995-2017,
 |    Jean-loup Gailly and Mark Adler,
 | and using "clock_gettime(CLOCK_MONOTONIC_RAW,&t)".

Glenn

jbowler:
What I observed is that if I look at the standard deviation of the timing
figures from a single test it is large, however I interleaved the
optimized/not tests and if I divide times from successive pairs and
consider the standard deviation of the resultant ratio it is small.  This
is on a loaded machine.

Anyway, but doing that rearrangement of the numbers I get results which are
about -14% time improvement with the SSE optimizations and my test
assembly.  Consistent with Glenn's results and (now) reproducible.

John Bowler
​

PR #154, opened 2017-03-21
Support NUGET

mancoast:


glennrp:
Who assigns the version number (7799, 7800, ...)?  What happens when libpng version number is bumped to (e.g.) 1.6.30beta01 ?

mancoast:
Greetings Glen,

The nuspec files contains the version number.
In general, I generate Nuget version from a simple suffix to the version number of base project.
Nugets enable easy continuous integration.

glennrp:
Yes, I saw that the version number is in nuspec.  My question is how do I maintain the "nuspec" file when I update libpng to a new version?

mancoast:
The nuget require the cmake build.  Given the new version doesn't break the cmake build.  simply edit the nuspec to your desire.  Then re-execute the build scripts.  The resulting nuget is what you upload to nuget servers for use with appveyor, etc...

glennrp:
I still need to know where the 7799/7800 version number comes from.  What is the version number for libpng-1.6.30beta01?  Do I just assign it myself, e.g., to 7801?

mancoast:
These numbers (7799/7800) are arbitrary values.  Nuget requires each package has a unique version. For libpng-1.6.30beta01; 1.6.30.01.7801

glennrp:
OK, thanks.  And for libpng-1.6.30beta02, libpng-1.7.0beta87, libpng-1.6.30rc01, libpng-1.6.30?  May I use the same number (7801) for all of the 1.6.30* and a different number (what?) for all of 1.7.0* ?

mancoast:
Essentially, you can use whatever number you want to use as long as it keeps increasing with each nuget package release/upload.  I try to keep nuget versions as an extension of the actual library version number.  Add an arbitrary nuget version number to the end of the png library version.  You could use the same number (7801), but that will confuse the end users when they refer back to this GIT repo to checkout the exact commit used to build the nuget.  Increment the nuspec every time you re-run the build batch.

mancoast:
Also, I would avoid release beta/RC nuget packages.  It could accidentally break someone else's build on continuous integration.  Nugets cannot be easily deleted, and end users such as myself, tend to update nugets frequently in MSBuild Solutions using Nuget Packet Manager.

glennrp:
> avoid release beta/RC nuget packages

That's not consistent with libpng's philosophy of testing everything thoroughly in beta/rc before public release.  How do you propose that we validate the nuget stuff, then?

mancoast:
Imagine Nuget as the windows version of 'YUM' or 'APT'.
The goal is to avoid 'release' of packages that have not already been tested.

The current Nuget includes stable libpng code base.
Most of these Nuget downloads are for (autonomous) continuous integration.
https://www.nuget.org/packages/libpng-msvc14-x64/

glennrp:
Understood.  But we don't distribute .deb and similar packages; we leave those to downstream distributors to develop.  It seems that the Nuget packaging stuff belongs in Nuget (as it is now, apparently), not in the libpng distribution.

mancoast:
Honestly, I made the Pull Request to offer new features to libpng.  I try to publish the Nuget build scripts upstream for commonly used packages as Nuget doesn't really have a distribution.  It is up to you and PNG Development Group if you want to support Windows Continuous Integration.  That is why the files are located in a subfolder of the contrib folder.   This is just one of MANY libraries that I use for continuous integration of Mandelbulber.  Please feel free to close this PR without merging.  However, I would prefer that the build scripts for libpng Nuget get merged.
REF: https://github.com/buddhi1980/mandelbulber2/commit/24b6d7976fd1dc6b78be220cef500689f7ce1f30

ctruta:
I agree with Glenn: I appreciate the usefulness of this contribution, but we have traditionally left downstream packaging (scripts, artifacts, etc.), as well as the associated maintenance efforts, for the downstream package maintainers.

PR #160, opened 2017-03-31
URL updates

vszakats:
It'd be nice to enable HTTPS on the SourceForge website of `libpng`, as described here:
https://sourceforge.net/blog/introducing-https-for-project-websites/

### DIFF
--- a/README
+++ b/README
@@ -90,7 +90,7 @@ You may also want a copy of the PNG specification.  It is available
 as an RFC, a W3C Recommendation, and an ISO/IEC Standard.  You can find
 these at http://www.libpng.org/pub/png/documents/
 
-This code is currently being archived at libpng.sf.net in the
+This code is currently being archived at libpng.sourceforge.io in the
 [DOWNLOAD] area, and at ftp://ftp.simplesystems.org.  If you can't find it
 in any of those places, e-mail me, and I'll help you find it.
 
--- a/contrib/arm-neon/android-ndk.c
+++ b/contrib/arm-neon/android-ndk.c
@@ -17,7 +17,7 @@
  *
  * Documentation:
  *    http://www.kandroid.org/ndk/docs/CPU-ARM-NEON.html
- *    http://code.google.com/p/android/issues/detail?id=49065
+ *    https://code.google.com/p/android/issues/detail?id=49065
  *
  * NOTE: this requires that libpng is built against the Android NDK and linked
  * with an implementation of the Android ARM 'cpu-features' library.  The code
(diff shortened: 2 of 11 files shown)


vszakats:
@glennrp Thank you for the merge. But, I've noticed that in this commit: https://github.com/glennrp/libpng/commit/82add753e511de10f6d3f633ac5903b0765791d6

All of these were reverted. Was this intended? If so, what was wrong with these updates?

vszakats:
By the looks of it, only those files were reverted where the date got bumped. So this may have been the result of a script doing it automatically based on a pre-patch version of these text files.

glennrp:
I think it was just libpng.3 and libpng-manual.txt which are actually built from another file that did not get properly updated.  I've reapplied 160.patch to the man pages.  See commit 64f81108db99383169d058c3dce3bd0d8a46bc29

vszakats:
Thank you @glennrp, it looks alright now!

glennrp:
There are still a few remaining "http:" instances in the man page, though.

vszakats:
I may have missed some. Which one(s) do you have in mind?

glennrp:
www.w3.org responds to both "http:" and "https:".  www.libpng.org and www.poynton.com still only respond to "http:".

vszakats:
Yes, I had to skip those unfortunately. (I use a script to find the ones responding to HTTPS with similar output)

On a more careful check some updates stayed reverted indeed inside `libpng.3`. Here's the patch to correct those:
```diff
diff --git a/libpng.3 b/libpng.3
index ce0159b0a..3079c7dca 100644
--- a/libpng.3
+++ b/libpng.3
@@ -5987,7 +5987,7 @@ release number plus "betaNN" or "rcNN".
 .LP
 .IR libpng :
 .IP
-http://libpng.sourceforge.net (follow the [DOWNLOAD] link)
+https://libpng.sourceforge.io/ (follow the [DOWNLOAD] link)
 http://www.libpng.org/pub/png
 
 .LP
@@ -5997,7 +5997,7 @@ http://www.libpng.org/pub/png
 .I libpng
 or at
 .br
-ftp://ftp.info-zip.org/pub/infozip/zlib
+http://zlib.net/
 
 .LP
 .IR PNG specification: RFC 2083
@@ -6006,11 +6006,11 @@ ftp://ftp.info-zip.org/pub/infozip/zlib
 .I libpng
 or at
 .br
-ftp://ftp.rfc-editor.org:/in-notes/rfc2083.txt
+https://www.ietf.org/rfc/rfc2083.txt
 .br
 or (as a W3C Recommendation) at
 .br
-http://www.w3.org/TR/REC-png.html
+https://www.w3.org/TR/REC-png.html
 
 .LP
 In the case of any inconsistency between the PNG specification
```

vszakats:
Issued a PR with the above: #161

PR #164, opened 2017-05-04
Don't double eval CMAKE_SYSTEM_PROCESSOR

KindDragon:


glennrp:
Patches should be against the libpng16 branch, not the master.  I've applied your change manually to the libpng16 branch.  See  commit 30d1f90

PR #177, opened 2017-09-06
Add support for loading images with associated alpha.

ioquatix:
Let me know what you think.

### DIFF
--- a/png.h
+++ b/png.h
@@ -2819,6 +2819,8 @@ typedef struct
 #  define PNG_FORMAT_FLAG_AFIRST 0x20U /* alpha channel comes first */
 #endif
 
+#define PNG_FORMAT_FLAG_ASSOCIATED_ALPHA 0x40U /* alpha channel is associated */
+
 /* Commonly used formats have predefined macros.
  *
  * First the single byte (sRGB) formats:
--- a/pngread.c
+++ b/pngread.c
@@ -3759,7 +3759,13 @@ png_image_read_direct(png_voidp argument)
          mode = PNG_ALPHA_PNG;
          output_gamma = PNG_DEFAULT_sRGB;
       }
-
+      
+      if ((change & PNG_FORMAT_FLAG_ASSOCIATED_ALPHA) != 0)
+      {
+         mode = PNG_ALPHA_OPTIMIZED;
+         change &= ~PNG_FORMAT_FLAG_ASSOCIATED_ALPHA;
+      }
+      
       /* If 'do_local_background' is set check for the presence of gamma
        * correction; this is part of the work-round for the libpng bug
        * described above.
@@ -3984,6 +3990,10 @@ png_image_read_direct(png_voidp argument)
 
       else if (do_local_compose != 0) /* internal error */
          png_error(png_ptr, "png_image_read: alpha channel lost");
+
+      if ((format & PNG_FORMAT_FLAG_ASSOCIATED_ALPHA) != 0) {
+         info_format |= PNG_FORMAT_FLAG_ASSOCIATED_ALPHA;
+      }
 
       if (info_ptr->bit_depth == 16)
          info_format |= PNG_FORMAT_FLAG_LINEAR;


ioquatix:
I didn't bother writing documentation because I'm not sure what will change... but once we settle on something (or nothing), I can write something.

ioquatix:
One more thing which I think would be useful, would be the ability to save PNG on disk using an associated alpha format. The only real difference is a flag in the binary data stream, and the ability to convert to pre/post multiplied on load as required.

ioquatix:
Oh, one more thing to consider... if the buffer is converted to associated alpha on load, it may also need to be converted back to unassociated on save... unless we directly support that in the file format.

glennrp:
Converting back to non-premultiplied from premultiplied is necessarily lossy.  At the extreme, when alpha is 0 the color samples become 0 as well regardless of their original content.  It's not possible to recover the original underlying color from a {0,0,0,0} pixel.

ioquatix:
@glennrp Absolutely, there is a fundamental difference between associated and un-associated alpha, and the ability to convert between them in either direction is a lossy process.

Both formats require different blending functions, but with modern compositors preferring associated alpha. (iOS CoreImage converts PNGs to associated alpha format, OpenGL, DirectX, Vulkan all only support associated-alpha frame buffers AFAIK).

I think what would make sense is for PNG to add support for associated alpha in the file format, and to allow the user to specify what they want in the pixel buffer. PNG should convert as required.

When saving a file, it should be saved in it's native format, unless the user explicitly requests that it should be converted.

ioquatix:
Taking into consideration the blending function for associated alpha pixels, one can see that when `alpha=0` the blend function becomes additive, which means the `rgb` components are added to the destination. It's a useful trait of the data representation. Just FYI.

glennrp:
There's no real good way to signal that a PNG file contains associated alpha.  It would require either a new chunk (like CgBI but not placed ahead of IHDR) or a revised existing chunk (such as a new field in IHDR).  One possibility would be some additional colortypes, e.g., with bit 4 representing associated alpha (color_type 12: black-white-premultiplied_alpha, 14: color,premultiplied_alpha). Either of these would be incompatible with existing decoders.  Alternatively add a new ancillary chunk to signal that the alpha is associated.  Probably the latter is the only way that is really feasible.

ioquatix:
All good thoughts.

As it seems you know, Apple is already doing something with CoreImage and making pre-multiplied images for iOS. This happens when you compile code with Xcode for iOS deployment. CoreImage somehow detects these mutant PNGs and loads them appropriately. http://iphonedevwiki.net/index.php/CgBI_file_format

I know that at the very least there would be some incompatibility, even if you just load the pixel data, it won't be displayed correctly, but this can be fixed going forward.

In addition, even if applications just get raw pixel data, at least it can be displayed, albeit incorrectly.

Given what we have to deal with, I'd probably vote for a new chunk, but my intuition is not very good since I don't know this project inside out. If you think it should come before the IHDR, but right now Apple doesn't? We can support both but issue a warning for Apple's format?

glennrp:
> If you think it should come before the IHDR, but right now Apple doesn't?
The other way around.  No chunks should come before IHDR.
Apple's CgBI format does other weird things like omitting CRC's and rearranging sample order.

ioquatix:
I see.

I think from a pragmatic point of view.

Let's try to offer Apple a standard way forward that doesn't alienate all the existing users of that tool.

We may disagree with how they did it. We can ignore the file format, but it simply alienates users.

My suggestion is to:

1/ Implement it how we think is best.

2/ At some level, support Apple's implementation where possible. Perhaps optionally emit a warning if the structure doesn't conform to 1 above.

ioquatix:
I just wanted to add, that Apple is probably doing the arrangement of samples in order to maximise throughput when loading from disk. The less work they have to do to get the image into the GPU memory, the better. I think you'll find there are a lot of use cases where this would be of benefit.

ioquatix:
Did you test this code before merging? I think we should consider documentation and tests too..

glennrp:
No, I just put it out there for people to try.  Perhaps it would have been more sensible to put it in libpng17.

ioquatix:
I see. I would like to think that we can bring a more comprehensive solution, but this is a good start.

In my case, I needed to support pre-multiplied alpha in both WebP and PNG, so I finally implemented the conversion a layer up: https://github.com/kurocha/images/blob/master/source/Images/Conversions.cpp - a bit of a hack to get things working and probably worth revisiting at some point. We have to convert 8-bit sRGB from unassociated to associated alpha.

However, I do think it would be a great addition to support it in the format itself.

PR #184, opened 2017-11-03
Misc. typos

luzpaz:
Some are user facing. Some are in actual code. Most are in source comments. Also, please double check the changes in contrib/tools/pngfix.c


luzpaz:
I would have submitted PR to `master` but it was behind both libpng17 and libpng16
Is there a preference as to which branch I should submit to ?

glennrp:
libpng16 first, then a separate patch for libpng17.  "master" will be taken care of automatically from libpng16.

luzpaz:
cool 👍

PR #197, opened 2017-12-20
avoid buffer overrun

mpw96:
Hi,
buffer has size `length`. If buffer does not contain a 0 byte, we can leave the for-loop a few lines above with `keyword_length==length-1`.

The check for the compression type below will then try to access `buffer[length]`, which is illegal.

Best regards
Marc-Philip


ctruta:
The issue that you're mentioning doesn't happen. The previous check already covers this case, because `(keyword_length + 3 > length)` is less strict than `(keyword_length + 1 >= length)`. You'll only get to see (correctly) the "truncated" error.

mpw96:
Yes, you are right. Thank you for checking!

PR #203, opened 2018-02-01
Optimize png_do_expand_palette for ARM

richard-townsend-arm:
ARM-specific optimization processes 8 or 4 pixels at once. Improves performance by around 10-22% on a recent ARM Chromebook.

### DIFF
--- a/arm/palette_neon_intrinsics.c
+++ b/arm/palette_neon_intrinsics.c
@@ -1,0 +1,137 @@
+/* palette_neon_intrinsics.c - NEON optimised palette expansion functions
+ *
+ * Copyright (c) 2017-2018 Arm Holdings. All rights reserved.
+ * Written by Richard Townsend <Richard.Townsend@arm.com>, February 2017.
+ *
+ * This code is released under the libpng license.
+ * For conditions of distribution and use, see the disclaimer
+ * and license in png.h
+ */
+
+#include "../pngpriv.h"
+
+#if PNG_ARM_NEON_IMPLEMENTATION == 1
+
+#include <arm_neon.h>
+
+/* Build an RGBA palette from the RGB and separate alpha palettes. */
+void
+png_riffle_palette_rgba(png_structrp png_ptr, png_row_infop row_info)
+{
+   png_const_colorp palette = png_ptr->palette;
+   png_bytep riffled_palette = png_ptr->riffled_palette;
+   png_const_bytep trans_alpha = png_ptr->trans_alpha;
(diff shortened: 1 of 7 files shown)


richard-townsend-arm:
I recently tested this in Chromium and [found that it it improves performance considerably](https://chromeperf.appspot.com/report?sid=9fbc808ed73ce39fd5be1cf7d24ab177c94e3e2970d20b27a02c147a2f37e1ba) for palette-based PNGs with transparency (24% on a Cortex A53 little core, and 10% on an Cortex A72 core) and without (22% on an A53 core). With that complete, I hope we can improve libpng for everybody, so please let me know what you think.

glennrp:
Thanks.

mtklein:
Hey there!  I helped review[1] this to get it into Chromium, and you may remember the work Matt Sarett and I did a while back on the optimized filter functions.

The particular approach this patch takes is something that's really only directly possible with NEON's deinterlacing load and reinterlacing store instructions.  It probably would be possible to accelerate these functions using SSSE3 and above (taking advantage of pshufb) but it wouldn't really be a "port"... I'd expect the x86 approach would want some new thought and not be a 1:1 translation.

All that's to say, while I think x86 versions of these routines are possibly valuable, here on ARM is where platform-specific code will have the greatest impact.  On ARM it's very natural to work with packed 3-byte rgb arrays, but that'll be a bit of an interesting obstacle on x86.

[1] https://chromium-review.googlesource.com/c/chromium/src/+/817116 for most of the history of that review, landed for good in https://chromium-review.googlesource.com/c/chromium/src/+/868396

Adenilson:
We should start shipping this optimization on Chromium M66 (branch on March 1th), it would be awesome if it could be merged upstream in the future.

richard-townsend-arm:
Hi folks, given that this project is still under active development, it'd be great to contribute this optimisation so that Chromium can re-sync with the canonical libpng. Does anybody have some review comments (especially @glennrp, @ctruta)?

arm/palette_neon_intrinsics.c
```diff
@@ -0,0 +1,137 @@
...
+   }
+}
+
+
+/* Expands a palettized row into RGBA. */
+int
+png_do_expand_palette_neon_rgba(png_structrp png_ptr, png_row_infop row_info,
+   png_const_bytep row, const png_bytepp ssp, const png_bytepp ddp)
+{
+
```
ctruta: Remove empty line

PR #208, opened 2018-02-17
Avoid mixing incompatible CRTs when linking to libpng.dll

pps83:
If libpng.dll is built using static linking (without importing C runtimes from whatever version of ms compiler that was used to build it), or when it was build by mingw then such dll won't work properly as stdio fread/fwrite will try to interpret FILE pointers from unrelated CRT and will result in undefined behavior.

### DIFF
--- a/png.c
+++ b/png.c
@@ -699,6 +699,9 @@ png_get_io_ptr(png_const_structrp png_ptr)
  * PNG_NO_STDIO or otherwise disabled PNG_STDIO_SUPPORTED, you must use a
  * function of your own because "FILE *" isn't necessarily available.
  */
+#ifdef png_init_io
+#undef png_init_io
+#endif
 void PNGAPI
 png_init_io(png_structrp png_ptr, png_FILE_p fp)
 {
@@ -710,6 +713,36 @@ png_init_io(png_structrp png_ptr, png_FILE_p fp)
    png_ptr->io_ptr = (png_voidp)fp;
 }
 #  endif
+
+void PNGAPI
+png_init_io2(png_structrp png_ptr, png_FILE_p fp, png_fread_ptr fread_fn,
+   png_fwrite_ptr fwrite_fn, png_fflush_ptr fflush_fn)
+{
+   if (png_ptr->read_data_fn != NULL)
+   {
+      png_init_read_io(png_ptr, fp, fread_fn);
+   }
+   else
+   {
+      png_init_write_io(png_ptr, fp, fwrite_fn, fflush_fn);
+   }
+}
+
+void PNGAPI
+png_init_read_io(png_structrp png_ptr, png_FILE_p fp, png_fread_ptr fread_fn)
+{
+   png_debug(1, "in png_init_read_io");
+   png_set_read_fn2(png_ptr, fp, png_default_read_data, fread_fn);
+}
+
+void PNGAPI
+png_init_write_io(png_structrp png_ptr, png_FILE_p fp, png_fwrite_ptr fwrite_fn,
+   png_fflush_ptr fflush_fn)
+{
+   png_debug(1, "in png_init_write_io");
+   png_set_write_fn2(png_ptr, fp, png_default_write_data, png_default_flush,
+      fwrite_fn, fflush_fn);
+}
 
 #  ifdef PNG_SAVE_INT_32_SUPPORTED
 /* PNG signed integers are saved in 32-bit 2's complement format.  ANSI C-90
--- a/png.h
+++ b/png.h
@@ -811,6 +811,12 @@ typedef PNG_CALLBACK(void, *png_read_status_ptr, (png_structp, png_uint_32,
 typedef PNG_CALLBACK(void, *png_write_status_ptr, (png_structp, png_uint_32,
     int));
 
+typedef PNG_CALLBACK(size_t, *png_fread_ptr, (png_voidp, png_size_t,
+   png_size_t, png_FILE_p));
+typedef PNG_CALLBACK(size_t, *png_fwrite_ptr, (png_const_voidp, png_size_t,
+   png_size_t, png_FILE_p));
+typedef PNG_CALLBACK(int, *png_fflush_ptr, (png_FILE_p));
+
 #ifdef PNG_PROGRESSIVE_READ_SUPPORTED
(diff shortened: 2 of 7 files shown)


pps83:
In my case I built a libpng dll using MS VS2015 compiler and wanted to run tests to validate that everything is ok. To run tests I compiled a separate version of libpng using msys/mingw and simply replaced dll with my own built. Some of tests failed:

image

I tried to debug and it appears that FILE pointers from the side of test executable are passed across DLL boundary. This change introduces three new api functions: `png_init_io2`, `png_init_read_io` and `png_init_write_io` and these differ from `png_init_io` by passing function pointers to fread/fwrite/fflush along with FILE pointer.
Also, png_init_io is redefined as a macro that to forwards FILE pointer to png_init_io2 along with fread/fwrite/fflush. This way simple recompilation of existing code that uses `png_init_io` will make it use `png_init_io2`.

jbowler (commented):
This is an attempt to fix an unfixable problem.  The only fix is, in fact, a one liner; delete png_init_io from the API.

If an app creates a FILE* to use by libpng it must either use the "set...fn" APIs and set the read/write/flush functions (exactly as if it were any other IO device pointer) or it must call png_init_io and rely on libpng to call the stdio fread/fwrite (etc) APIs.

On Microsoft systems a DLL and an application which uses it may link to equivalent but different implementations of the same API, however this means that on those systems data that is manipulated by the API may not safely be passed between the application and DLL.  This is a well known fact.

png_init_io violates this requirement and can't be changed to not violate it.  The alternative png_read/write_fn APIs are ok (since the io_ptr is not interpreted by libpng if the API is called correctly).

So the only fix is to delete png_init_io from the API, however on non-Microsoft systems and even on MS systems with static linking the API is incredibly useful; it makes writing PNG read/write code incredibly easy.

The "simplified" API fixes this but the cost is large; the API does not permit an arbitrary ANSI-C compliant IO pointer to be used.

So this fix is an unbelievably complex way of not fixing anything: the real fix to the reported issue is to delete line 1597 of png.h (PNG_EXPORT(74, void, png_init_io...) and that is a one line fix.

pps83:
> This is an attempt to fix an unfixable problem. The only fix is, in fact, a one liner; delete png_init_io from the API.

Yes, effectively, that's not ok to use FILE pointer where it wasn't created. My PR still keeps png_init_io as exported API for binary compatibility, but makes all new code use `png_init_io2` instead which will make sure that fread/fwrite/fflush are used from the side of the user of libpng.

jbowler:
> Yes, effectively, that's not ok to use FILE pointer where it wasn't created. 

Right, but my suggestion (which doesn't really work for reasons I will explain) was to remove png_init_io with the intention of removing all the stdio.h code.  This takes libpng out of the IO business; all IO happens in the app.  This eliminates a bug-farm.  It's not just the mixed-implementation problem on WindowsNT which, as you point out, happens anywhere (not just Windows) with a static link of a libpng DLL which includes the stdio implementation.  There are other issues such as the fact that libpng does not work with a non-blocking FILE*; it can take an app programmer a long time to find that out!

It's easy to give sample code and it already exists in many places which don't have the #ifdef mess of pngrio.c and pngwio.c; app writers can just copy'n'paste this into their programs.

Indeed libpng has always supported this via `PNG_NO_STDIO` and (now) "`option stdio off`" (I'm quoting from memory; I currently don't have access to my systems.)  App developers can use this to verify forward compatibility.

The gotcha is that removing stdio support from libpng does not remove the need for png_init_io, although it does remove the definition.  This is shown by the simplified API code, which I intended to use just the exported APIs (so it could be extracted to a separate library).  See line 1484; originally the code called png_init_io, but then it didn't build with PNG_NO_STDIO so I had to violate the libpng abstraction by directly assigning to the io_ptr.

You can see that png_init_io is not specific to stdio.  It's also horribly misnamed; it does not do any initialization, it just sets io_ptr.  It can also be called multiple times and called after the read/write functions are set up.

So libpng faces a choice: either remove stdio support completely, which is bound to annoy someone, or isolate it properly which is what your proposal would do.  In fact your proposal could be adopted and immediately deprecated; that would preserve the existing API for one major release then drop it.

The functionality of png_init_io still has to stay but it should renamed as png_set_io_ptr.  The png_init_io_2 name is just plain confusing: it's png_init_stdio and should take a FILE* argument.  Having a macro which just takes a FILE* and passes read/write/flush from the C functions seems fine; this would be a lot easier with C++ :-)

The rewrite of the simplified API is a larger change of course but some of the non-core test programs use the stdio API.  I'm not sure it is tested by the core tests (IRC pngstest uses memory IO).

So ABI compatibility is dropped, which is fine on a major release, but API compatibility remains with some things deprecated.  I think just `#define png_init_io png_set_io_ptr` works because the code that initializes the read/write functions to stdio can just output a png_warning about use of a deprecated API.  I don't know if your change does that already; I'm suggesting a slightly different structure here but it's not that different based on a my cursory examination.

This still isn't a complete review; I'm not discussing specifics of your code here, just exploring general issues.  There are lots of opportunities for code cleanup here and it might be really good to put all the stdio stuff in a new file, '`pngstdio.c`' for example, if @ctruta is amenable to adding source files to the build.

ctruta:
Personally I wouldn't mind adding new source files. There is a problem, however, with projects that use their own custom build systems. Not a big problem, mind you, but a disruptive one, nonetheless.

OTOH libpng-ng is a good opportunity for us to break our current big and giant `png*.c` source files into smaller and more manageable ones.

jbowler:
> OTOH libpng-ng is a good opportunity for us to break our current big and giant `png*.c` source files into smaller and more manageable ones.

Certainly.  10,000 line files are just plain bad and in static builds apps which use libpng end up getting stuff that can never be called.  The whole 'minconfig' stuff (scripts/pnglibconf.dfa etc) is a messy #define approach to fix that.

The directory layout also leaves much to be desired.  It seems that there is a meta-standard these days to put the implementation in src/ and the interface in inc/.  That strikes me as good.

jbowler:
#622 resolves this in all three APIs which use FILE.

jbowler:
#622 is rejected by @ctruta so this solution, which is the one @ctruta favours, needs to be elevated.

PR #215, opened 2018-03-26
[CMake] Corrected typo regarding disabling PNG_INTEL_SSE

VileLasagna:
The ideal solution would be to change PNG_INTEL_SSE into a proper cmake OPTION, but I'm not 100% confident casing on ON/OFF wouldn't be an issue so this is the smaller change. 
Regardless, code as is is broken as it's impossible to disable it should the user want to, for whatever reason


ctruta:
A fix for all architectures was applied in commit ddc842bf0690bb6da6dfdd3cd8a53b4421249db0.

PR #233, opened 2018-06-26
contrib/gregbook: fix syntax errors

jwilk:


tangyaofang:
Obvious grammatical mistakes, thank you for discovering.

jbowler:
The first change is correct but hardly matters; this is documentation, not code.  The second change appears to be wrong and misleading; the point of the code is to protect against overflow, the code correctness crowd can complain about this way of doing it but the original seems correct to me.

PR #253, opened 2018-08-15
Fix error in arm_neon.h on msvc ARM64

zzzzRuby:


zzzzRuby:
OK. I merged the three commits into one and added some comment.
And my real name is Zhijie Liang.
Thanks.

PR #258, opened 2018-11-08
Allow overwriting library debug postfix

eldruin:
This allows overwriting the library debug postfix from the outside by making it a cmake cache variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,7 +222,7 @@ endif(PNG_HARDWARE_OPTIMIZATIONS)
 set(PNG_LIB_NAME png${PNGLIB_MAJOR}${PNGLIB_MINOR})
 
 # to distinguish between debug and release lib
-set(CMAKE_DEBUG_POSTFIX "d")
+set(CMAKE_DEBUG_POSTFIX CACHE "d")
 
 include(CheckCSourceCompiles)
 option(ld-version-script "Enable linker version script" ON)


ctruta:
Out of curiosity: why would people want to override this? I thought the 'd' postfix is more or less standard. Besides, this (I would guess) is only for local workflows, since debug builds don't usually get published.

eldruin:
We use different debug suffixes in our builds: `_d` on Windows and Linux and `_debug` on macOS.

ctruta:
I understand your use case. However, as per my understanding CMake, your proposed solution will affect every single project processed after libpng in an unexpected and undesired way. The `CMAKE_DEBUG_POSTFIX` variable should be local to the libpng source tree, not global, and not visible to other projects.

I am still far from being a CMake expert, though. I will see if adding
```set_target_properties(png PROPERTIES DEBUG_POSTFIX "d")```
will do the trick.

ctruta:
Could you please review commit e519af8b49f52c4ac400f50f23b48ebe36a5f4df and let me know this solution is to your liking?
Feel free to comment, or to suggest an alternative solution in a PR.

eldruin:
That should be fine for us, thank you!

ctruta:
Thank you for your 5-year-long patience :slightly_smiling_face:

eldruin:
No worries, I know how it goes :)

Issue #256, opened 2018-10-08
libpng error: [00][B8][05][A9]: invalid chunk type


kadogo:
Hello,

I use Debian Stretch and I find out a bug when I concat many pictures together.

```
libpng: 1.6.28-1
feh: 2.18-2
```

![tnew](https://user-images.githubusercontent.com/5759396/46617499-2d578280-cb1d-11e8-9ea1-cf6a49828d8e.png)

I have this message if I try to read the picture with feh. Feh don't show anything.

```
$ feh tnew.png 
libpng warning: IDAT: bad parameters to zlib
libpng error: [00][B8][05][A9]: invalid chunk type
libpng warning: IDAT: bad parameters to zlib
libpng error: [00][B8][05][A9]: invalid chunk type
libpng warning: IDAT: bad parameters to zlib
libpng error: [00][B8][05][A9]: invalid chunk type
```

The result of pngcheck are correct

```
$ pngcheck tnew.png 
OK: tnew.png (100x50000, 24-bit RGB, non-interlaced, 98.3%).
```

If I try to read the picture on firefox or on Gimp it work without any issues;

Thanks for your help, don't hésitate to ask more information if needed.

Cheers

Kcchouette:
Reproducible using feh 2.19.3 and libpng 1.6.34

tangyaofang:
No one seems to be paying attention to this problem?
Has this problem been solved in the new version?

kadogo:
No change in Debian so I can always reproduce the bug with my image from the first post.

Kcchouette:
Reproduced using feh 3.1.3 and libpng 1.6.36

kadogo:
I'm still able to reproduce on Debian Buster

```
 *** 1.6.36-6 500
        500 http://ftp.be.debian.org/debian buster/main amd64 Packages
        100 /var/lib/dpkg/status
```

@ctruta Sorry to ping you but it look like the issue is not saw at all, is there maybe some direction we can look about to find a solution to solve this?

Cheers.

Kcchouette:
On manjaro I haven't the same kind of error:

```
$ feh 3e516f48-9d05-4499-bf7a-375b4f45a77e.png
feh WARNING: 3e516f48-9d05-4499-bf7a-375b4f45a77e.png - No Imlib2 loader for that file format
feh: No loadable images specified.
See 'feh --help' or 'man feh' for detailed usage information
```

It's feh 3.3 with libpng 1.6.37

PR #259, opened 2018-11-08
Allow overwriting png library name

eldruin:
This allows overwriting the png library output name from the outside by making it a cmake cache variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,7 +219,7 @@ endif()
 endif(PNG_HARDWARE_OPTIMIZATIONS)
 
 # SET LIBNAME
-set(PNG_LIB_NAME png${PNGLIB_MAJOR}${PNGLIB_MINOR})
+set(PNG_LIB_NAME CACHE png${PNGLIB_MAJOR}${PNGLIB_MINOR})
 
 # to distinguish between debug and release lib
 set(CMAKE_DEBUG_POSTFIX "d")


ctruta:
Thank you, this may be useful, but before that:

I inherited the code base with both PNGLIB_NAME and PNG_LIB_NAME. There should only be one, and I need to resolve that, eventually.

As far as nomenclature is concerned, I see that CMake uses both `_LIB_NAME` and `_LIBRARY_NAME`, and I don't know which name is appropriate for which use case.

(I'm not a CMake expert, BTW.)

eldruin:
I see. I do not know if there is an actual convention about `_LIB_NAME` vs .`_LIBRARY_NAME`. What is somewhat unusual is not having an underscore before `LIB` (i.e. `PNGLIB_NAME`), I would say. Although I see that it matches the other `PNGLIB_*` variable names nicely.
Looking at the CMake modules, here are the occurrences of [`LIB_NAME`](https://gitlab.kitware.com/search?utf8=%E2%9C%93&snippets=&scope=&search=LIB_NAME&project_id=541) and [`LIBRARY_NAME`](https://gitlab.kitware.com/search?utf8=%E2%9C%93&snippets=&scope=&search=LIBRARY_NAME&project_id=541)
I think having just one variable is much more important than whatever is named. I saw that the difference in both variables is that `PNGLIB_NAME` contains the library prefix for all platforms (but is only used for POSIX platform cases) and `PNG_LIB_NAME` does not contain the library prefix and is used to actually give the library the name, which will make CMake add the `lib` prefix only where appropriate.
I will open a separate PR addressing this.

ctruta:
As I mentioned in #256:
* I withdrew completely `PNGLIB_NAME` and I used the hard-coded name "libpng" instead, because that name should never change.
* I replaced `PNG_LIB_NAME` with `PNG_SHARED_OUTPUT_NAME` and `PNG_STATIC_OUTPUT_NAME`. Rather than making them cached, I opted to let the user set prefix and suffix modifiers, such as `CMAKE_SHARED_LIBRARY_PREFIX`, etc., `CMAKE_STATIC_LIBRARY_SUFFIX`, etc. I understood that this is CMake's "proper" way of modifying target file names.

eldruin:
I need to check on it but I think that should work for us and will make a new suggestion if it somehow does not.
Thank you!

PR #260, opened 2018-11-26
Repurpose PNGLIB_NAME to remove internal variable PNG_LIB_NAME

eldruin:
As discussed in #259


eldruin:
I just resolved the conflicts.

ctruta:
While fixing an unrelated defect, I deleted both `PNGLIB_NAME` and `PNG_LIB_NAME`. Most confusingly, `PNGLIB_NAME` and `PNG_LIB_NAME` were used for two different purposes, and keeping those in sync would have been a nightmare. I opted to drop not only the variable name, but also the purpose of `PNGLIB_NAME`, by hard-coding the "libpng" name in the CMakefile. The other variable gets to live under different names.

See commit efc96c9d59a9052c4db9245ca0a204fc267fe396.

I'm closing this PR, and we can continue the conversation in #259.

Issue #187, opened 2017-11-07
pngrutil.c:429:16: 'inflateValidate' is only available on macOS 10.13 or newer


sgowen:
See the png_inflate_claim method inside pngrutil.c

What would be a good work-around for applications targeting pre < macos 10.13  ?

Is it safe for me to just undefine PNG_SET_OPTION_SUPPORTED ?

glennrp:
This should be sufficient

ZLIB_VERNUM >= 0x1290 

The definition is in zlib.,h

sgowen:
Changing zlib.h isn't an option for me, since it is defined in the macos and ios sdks.

So basically, libpng is making the assumption that it can call inflateValidate purely because ZLIB_VERNUM >= 0x1290, but Xcode is telling me that 'inflateValidate' is only available on macOS 10.13 or newer.

Yeah so the real issue is that I am compiling against mac os 10.13, but want to support 10.10 and up. If I change my deployment target to 10.13, the issue goes away.

But it would be nice to have another flag instead of purely relying on ZLIB_VERNUM

sgowen:
I've made the above change in my fork of libpng, which is totally fine.

I just prefer to use the official repo unchanged wherever I can.

sgowen:
Actually, now that I am thinking about it more, a better change (for my purposes) is to just undefine PNG_IGNORE_ADLER32, that way all my platforms are affected equally and not just mac.

But I'm still curious if you've ran into this issue before and what you might've done to resolve it.

sgowen:
@glennrp sorry didn't see your edit, you are exactly right. So, since I am going to continue to use the zlib embedded into mac os and iOS, undefining PNG_IGNORE_ADLER32 seems to be my best solution. I'll close this issue now.

mrserb:
@glennrp Hello, I am faced the same issue while working on the OpenJDK 12 project.
The problem is that there is a call to "inflateValidate" function in pngrutil.c, guarded by a preprocessor check of ZLIB_VERNUM being high enough and by the "PNG_IGNORE_ADLER32". If we compile this call in and link to the newer version of zlib, we will get link errors if the code is executed on an older Mac/Solaris with an older version of zlib. 

I understand that it is possible to fix it by the change of “ZLIB_VERNUM” or “PNG_IGNORE_ADLER32” but both are not the “official” options which usually defined in "pnglibconf.h"

[...]

PR #261, opened 2018-11-27
Support of PNG_ADLER32_SUPPORTED

mrserb:
This is the fix for https://github.com/glennrp/libpng/issues/187

The new "official" options is added to enable/disable ADLER32


tangyaofang:
It is a good solution.

PR #270, opened 2019-01-10
Fixes pertaining background color and grayscale conversions

CyberShadow:
After adding libpng support to my image library, I wrote some test code which generated PNG files with all valid combinations of relevant parameters, and asked libpng to parse each into every compatible format supported by my image library. Some of these combinations produced unexpected results, which seem to be caused by bugs in libpng.

This pull request includes two patches which attempt to rectify these bugs. Please see the individual commits' commit messages for details.

FWIW - the source code for my test suite that identified this bug is here:
https://github.com/CyberShadow/ae/blob/master/utils/graphics/libpng.d

### DIFF
--- a/contrib/libtests/pngadhoc.c
+++ b/contrib/libtests/pngadhoc.c
@@ -1,0 +1,153 @@
+/* pngadhoc.c
+ *
+ * Copyright (c) 2019 Vladimir Panteleev
+ *
+ * This code is released under the libpng license.
+ * For conditions of distribution and use, see the disclaimer
+ * and license in png.h
+ *
+ * Ad-hoc libpng regression tests.
+ */
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+/* Define the following to use this test against your installed libpng, rather
+ * than the one being built here:
+ */
+#ifdef PNG_FREESTANDING_TESTS
+#  include <png.h>
+#else
+#  include "../../png.h"
+#endif
+
+struct buffer
+{
+   png_bytep data;
+   size_t length;
+};
+
+static void
+png_write_callback(png_structp png_ptr, png_bytep data, png_size_t length)
+{
+   struct buffer *buffer = (struct buffer *)png_get_io_ptr(png_ptr);
+   buffer->data = realloc(buffer->data, buffer->length + length);
+   if (buffer->data == NULL)
+      png_error(png_ptr, "OOM while writing PNG");
+   memcpy(buffer->data + buffer->length, data, length);
+   buffer->length += length;
+}
+
+static void
+png_read_callback(png_structp png_ptr, png_bytep data, png_size_t length)
+{
+   struct buffer *buffer = (struct buffer *)png_get_io_ptr(png_ptr);
+   if (buffer->length < length)
+      png_error(png_ptr, "EOF while reading PNG");
+   memcpy(data, buffer->data, length);
+   buffer->data += length;
+   buffer->length -= length;
+}
+
+static int
+png_test_bKGD_rgb_to_gray()
+{
+#if defined(PNG_SEQUENTIAL_READ_SUPPORTED) && \
+   defined(PNG_bKGD_SUPPORTED) && \
+   defined(PNG_READ_BACKGROUND_SUPPORTED) && \
+   defined(PNG_READ_EXPAND_SUPPORTED) && \
+   defined(PNG_READ_STRIP_ALPHA_SUPPORTED) && \
+   defined(PNG_READ_RGB_TO_GRAY_SUPPORTED)
+
+   png_structp png_ptr = png_create_write_struct(PNG_LIBPNG_VER_STRING,
+      0, 0, 0);
+   png_infop info_ptr = NULL;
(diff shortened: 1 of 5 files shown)


contrib/libtests/pngadhoc.c
```diff
@@ -0,0 +1,246 @@
+/* pngadhoc.c
```
CyberShadow: After looking for a while, I couldn't find an existing place in the test suite where such tests would be really appropriate, so I added a new file.

contrib/libtests/pngadhoc.c
```diff
@@ -0,0 +1,246 @@
...
+
+   png_set_IHDR(
+      png_ptr, info_ptr,
+      1, 1,
+      8,
+      PNG_COLOR_TYPE_RGBA,
+      PNG_INTERLACE_NONE,
+      PNG_COMPRESSION_TYPE_DEFAULT,
+      PNG_FILTER_TYPE_DEFAULT);
+   png_set_bKGD(png_ptr, info_ptr, &write_background);
```
CyberShadow: Technically, adding `bKGD` to the test PNG (here and in the second test) is not necessary, but I wanted to illustrate that the bug happens with the `png_set_background` usage as suggested by the manual.

pngrtran.c
```diff
@@ -1202,6 +1202,10 @@ png_init_palette_transformations(png_structrp png_ptr)
 #endif /* READ_EXPAND && READ_BACKGROUND */
 }
 
+#ifdef PNG_READ_RGB_TO_GRAY_SUPPORTED
+static int png_do_rgb_to_gray_1(png_structrp png_ptr, png_color_16p color, png_byte bit_depth);
+#endif
```
CyberShadow: Not sure what the policy is regarding forward declarations versus shuffling function order around.

pngrtran.c
```diff
@@ -1275,13 +1279,42 @@ png_init_rgb_transformations(png_structrp png_ptr)
...
+               png_ptr->rgb_to_gray_status=1;
+               if ((png_ptr->transformations & PNG_RGB_TO_GRAY) ==
+                   PNG_RGB_TO_GRAY_WARN)
+                  png_warning(png_ptr, "nongray background color");
+
+               if ((png_ptr->transformations & PNG_RGB_TO_GRAY) ==
+                   PNG_RGB_TO_GRAY_ERR)
+                  png_error(png_ptr, "nongray background color");
+            }
          }
```
CyberShadow: I didn't add code to do the same conversion of `png_ptr->trans_color`. I couldn't find a situation where doing so would matter, and I don't understand why the grayscale code above is doing it. In any case, the conversion here is lossy, and since its result can't be used to actually perform color-key transparency, it would only have informational value, but even then, there's no way to query it.

pngrtran.c
```diff
@@ -3188,6 +3221,55 @@ png_do_rgb_to_gray(png_structrp png_ptr, png_row_infop row_info, png_bytep row)
...
+
+   png_bytep dp = row;
+   if (png_ptr->bit_depth == 16)
+   {
+      *(dp++) = (png_byte)((color->red >> 8) & 0xff);
+      *(dp++) = (png_byte)(color->red & 0xff);
+      *(dp++) = (png_byte)((color->green >> 8) & 0xff);
+      *(dp++) = (png_byte)(color->green & 0xff);
+      *(dp++) = (png_byte)((color->blue >> 8) & 0xff);
+      *(dp++) = (png_byte)(color->blue & 0xff);
```
CyberShadow: This is ugly (encode / swap bytes just so `png_do_rgb_to_gray` decodes them again), but the alternative is an even uglier refactor.

tangyaofang:
Thank you for your use and contribution to libpng.

jbowler:
Yes; the transforms do not always compose correctly.  I identified a number of cases (I'm not sure if this is one, probably not because memory format conversions are not a high priority).

The problem is that it is whack-a-mole with the 1.6 (and earlier) implementation.  I tried to fix this in 1.7 and believe I got it all right but that's my version of 1.7; this (the transform code) is one place where a complete rewrite is in order (that's how I fixed it.)

As a result it's impossible to review a fix like this as just a one file change (to pngrtran.c).  The first step is to come up with a test case using pngstest - i.e. a set of arguments to pngstest that demonstrate the problem or, possibly, since pngstest has no support at present for the swap transforms, to implement a new test program.

Without a test program accepting such a patch is inadvisable.  Maybe in a major release if someone is committed to testing everything.

A better patch for libpng 1.6 is to detect the incompatible transforms when they are set and simply png_error out; that is way better than producing a completely valid PNG full of garbage.  Clearly since this bug has not been widely reported people who encounter it have already worked round it by doing one or other of the composed transforms themselves.

jbowler:
@CyberShadow do you get a warning from inside png_init_read_transformations, something like:

"libpng does not support gamma+background+rgb_to_gray"

but any warning should be reported.

CyberShadow:
No, I do not remember encountering such a warning.

Worth pointing out that this pull request is over five years old, so my recall might not be perfect.

PR #277, opened 2019-01-30
Adds podspec to repo

### LINKED ISSUE #273: Cocoapods support
Hi,

It looks like a podspec [has been submitted to cocoapods](https://cocoapods.org/pods/libpng) to make it easily integrable in Objective-C / Swift.

Unfortunately it seems the podspec wasn't committed to this repo and there's an [issue when building for arm devices](#219).

I'd like to submit either a new podspec or update the existing podspec to make it work for arm devices and then push it to the trunk.

Would you be ok to accept a PR that adds a single `.podspec` file to this repo?

/cc @Interfere who I believe wrote the original podspec.

cuva:
This repo is the source of https://cocoapods.org/pods/libpng. The current podspec wasn't checked in which makes updating the pod more difficult.

This spec resolves #273


ctruta:
Thank you for your contribution, but I closed #273 as WONTFIX.
In my opinion, the dedicated cocoapods repo is better suited for this kind of work.

Issue #274, opened 2019-01-25
improve libpng coverage on OSS-Fuzz


kcc:
requesting help with gaining more coverage on OSS-Fuzz, details here: https://github.com/google/oss-fuzz/issues/2111

kcc:
My first attempt to improve the coverage: https://github.com/google/oss-fuzz/pull/2114
Please give feedback, and consider taking over this fuzz target.
It very quickly finds a memory leak:

```
==1==ERROR: LeakSanitizer: detected memory leaks
--
  | Direct leak of 256 byte(s) in 1 object(s) allocated from:
  | #0 0x4efdcf in malloc _asan_rtl_
  | #1 0x55c0cd in OSS_FUZZ_png_malloc libpng/pngmem.c:179:10
  | #2 0x599de2 in OSS_FUZZ_png_set_tRNS libpng/pngset.c:1019:35
  | #3 0x58a79c in OSS_FUZZ_png_handle_tRNS libpng/pngrutil.c:1914:4
  | #4 0x55ceac in OSS_FUZZ_png_read_info libpng/pngread.c:245:10
  | #5 0x55fae6 in OSS_FUZZ_png_read_png libpng/pngread.c:1052:4
  | #6 0x538b83 in LLVMFuzzerTestOneInput /src/libpng_transforms_fuzzer.cc:100:5
```

kcc:
The leak above is fixed. Now the blocker for further fuzzing is #275

kcc:
@sboukortt thanks for the leak fix. 
Maybe you can suggest improvements to my experimental PNG fuzz target? 
https://github.com/google/oss-fuzz/blob/master/projects/libpng-proto/libpng_transforms_fuzzer.cc

The goal is to be able to cover more parts of the libpng code. 
The one we currently have in libpng trunk (https://github.com/glennrp/libpng/tree/libpng16/contrib/oss-fuzz) is significantly weaker than the new one, but there is still lots to improve.

LocutusOfBorg:
can you please try again with the patches I mentioned here?
https://github.com/glennrp/libpng/issues/275#issuecomment-481162675

kcc:
Err. What do you mean by "try again"? 
The fuzzing runs automatically and picks up the fresh master every day. 
(I've just realized that asan has been disabled on oss-fuzz due to #275, I've enabled it back.

LocutusOfBorg:
ok, nice! I was not aware of the continuous integration against master... :)

[...]

PR #279, opened 2019-02-14
Fix the memory leak mentioned in #274

sboukortt:
The old structure of the code was along the lines of:

    allocate trans_alpha;
    if (problem) {
      // Jumps away from this function
      png_warning("tRNS chunk has out-of-range samples for bit_depth");
    }
    mark trans_alpha as to-free;

### DIFF
--- a/pngset.c
+++ b/pngset.c
@@ -1019,6 +1019,9 @@ png_set_tRNS(png_structrp png_ptr, png_inforp info_ptr,
           info_ptr->trans_alpha = png_voidcast(png_bytep,
               png_malloc(png_ptr, PNG_MAX_PALETTE_LENGTH));
           memcpy(info_ptr->trans_alpha, trans_alpha, (size_t)num_trans);
+
+          info_ptr->valid |= PNG_INFO_tRNS;
+          info_ptr->free_me |= PNG_FREE_TRNS;
        }
        png_ptr->trans_alpha = info_ptr->trans_alpha;
    }


sboukortt:
This may not be the best or most elegant fix but now that the problem is identified, possible solutions can be discussed.

kcc:
This leak is still not fixed, right?

sboukortt:
Indeed not, as far as I can tell. The file that this pull request changes has not been touched since the request was opened.

tangyaofang:
@kcc :
Do you validate this modification scheme? 
To ensure that the incorporated code really solves the problem.
I don't see sboukortt submitting test cases.

sboukortt:
The test cases are the fuzz targets mentioned in [#274](https://github.com/glennrp/libpng/issues/274#issuecomment-458021367), and I have verified that my modifications do stop the fuzz target from erroring.

kcc:
Yes, I did verify that this change fixes the leak. (@tangyaofang, was that your question?)

tangyaofang:
But why did the author not merge these modifications into the code library.

sboukortt:
Sorry, I am not sure that I understand the question. Do you mean the fuzz targets?

sboukortt:
@ctruta Is there anything I can help with to unblock this?

ctruta (approved):

ctruta:
Integrated, tested, acknowledged. Many thanks!

PR #284, opened 2019-04-12
Fix typo in the new v2 license

ojeda:
The new v2 license has a typo in one of the sentences. I guess it was introduced when the text was manually re-formatted from the all caps style of other licenses which contain a very similar sentence (e.g. BSD).


ctruta:
I must apply your change both to the license file and to png.h. Thank you very much!

PR #285, opened 2019-04-16
Use arm_neon.h when compiling aarch64 intrinsics with clang-cl.

rvandermeulen:
When building for Windows aarch64 with clang-cl, linking fails due to a number of missing ARM intrinsics symbols. This is because arm64_neon.h is an MSVC-specific header and clang-cl doesn't know what to do with it.

The solution is to add a check for clang in the conditional so that clang-cl builds will use arm_neon.h like clang builds on other platforms.

### DIFF
--- a/arm/filter_neon_intrinsics.c
+++ b/arm/filter_neon_intrinsics.c
@@ -18,7 +18,7 @@
 /* This code requires -mfpu=neon on the command line: */
 #if PNG_ARM_NEON_IMPLEMENTATION == 1 /* intrinsics code from pngpriv.h */
 
-#if defined(_MSC_VER) && defined(_M_ARM64)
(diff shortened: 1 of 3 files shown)


ctruta:
Thank you, Ryan, for the fix. I would have normally asked you to restyle your patch a little bit, and place the two compiler macros near one another. But then I noticed that @mtklein did exactly that, independently, in a later pull request (#286), which he has eventually withdrawn.

OTOH I prefer the subject line from this commit (#285), because it's more descriptive while fitting in the 72 character limit, and there is still room left for the `arm:` prefix.

Would it be ok if I credit you both, by the means of "Signed-off-by" lines, and then later on, in the release change log?

mtklein:
Sure, however you like.  Thanks for merging #285.

PR #286, opened 2019-04-17
use arm_neon.h with clang-cl

mtklein:
MSVC's arm64_neon.h header doesn't work with clang-cl;
if you try you'll find yourself missing symbols like
"neon_foo" at link time.  These are cl.exe intrinsics,
but clang-cl doesn't know that.

Instead have clang-cl use the normal arm_neon.h.


mtklein:
Oh, I guess I should have looked at the open pull requests!  Looks like https://github.com/glennrp/libpng/pull/285 beat me to this.

Issue #269, opened 2019-01-05
memory leak in png_create_info_struct


zerokeeper:
Hi,libpng team. there is a memory leak  in the  file png.c:368 of  function png_create_info_struct.
the bug is trigered by `./pngcp  poc  /dev/null`

[libpng_poc.zip](https://github.com/glennrp/libpng/files/2729282/libpng_poc.zip)

the asan debug info is as follows:

=================================================================
==10300==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 360 byte(s) in 1 object(s) allocated from:
    #0 0x7fe088bf9602 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x98602)
    #1 0x422f95 in png_create_info_struct /root/fuzz/libpng-1.6.36/png.c:368

SUMMARY: AddressSanitizer: 360 byte(s) leaked in 1 allocation(s).

https://github.com/glennrp/libpng/blob/eddf9023206dc40974c26f589ee2ad63a4227a1e/png.c#L352-L376

carnil:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-6129

pgajdos:
```
$ pngcp libpng_poc /dev/null
libpng_poc: error(libpng): read: Not a PNG file

=================================================================
==30270==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 360 byte(s) in 1 object(s) allocated from:
    #0 0x7f07e0ecced0 in malloc (/usr/lib64/libasan.so.5+0xebed0)
    #1 0x7f07e1d3c23f in png_malloc_base /usr/src/debug/libpng16-1.6.36-0.x86_64/pngmem.c:95
    #2 0x7f07e1d24d55 in png_create_info_struct /usr/src/debug/libpng16-1.6.36-0.x86_64/png.c:368
    #3 0x55806e30a07b in read_png contrib/tools/pngcp.c:1775
    #4 0x55806e30d0b1 in cp_one_file contrib/tools/pngcp.c:2180
    #5 0x55806e30dba4 in cppng contrib/tools/pngcp.c:2288
    #6 0x55806e30e081 in main contrib/tools/pngcp.c:2351
    #7 0x7f07e0a43fea in __libc_start_main (/lib64/libc.so.6+0x22fea)

SUMMARY: AddressSanitizer: 360 byte(s) leaked in 1 allocation(s).
$
```
[...]

PR #293, opened 2019-06-10
Repair of CVE-2019-6129 #269

tangyaofang:
Repair of CVE-2019-6129 #269

### DIFF
--- a/contrib/tools/pngcp.c
+++ b/contrib/tools/pngcp.c
@@ -506,7 +506,7 @@ static void
 display_clean_read(struct display *dp)
 {
    if (dp->read_pp != NULL)
-      png_destroy_read_struct(&dp->read_pp, NULL, NULL);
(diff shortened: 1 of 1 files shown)


heaven-hq:
@ctruta @ctruta I have also verified this PR. When an abnormal situation is entered, it will enter 
`static void display_clean_read(struct display *dp)`
```
if (dp->read_pp != NULL)
      png_destroy_read_struct(&dp->read_pp, NULL, NULL);
```
branch and cause memory leak.This patch could solve this problem very well. Do you plan to merge this PR?

ctruta:
@jbowler could you please review this fix as well? The commit does seem to solve the problem, but I am not sure it's the cleanest solution, unless you confirm that it is.

jbowler:
NOTE: this is a duplicate of #269, I'll put comments in the original PR; this should be closed.

ctruta:
A complete fix was provided by the original author, and integrated into the master branch.

Issue #255, opened 2018-09-12
pngcp has a segfault


tianxiaogu:
Is it a bug of glibc? Not sure whether it is important. The crash can be reproduced with the HEAD or the one in pngtools released in Ubuntu 18.04.

[test-case.zip](https://github.com/glennrp/libpng/files/2373935/test-case.zip)

```
test-case: error(libpng): read: IEND: out of place
AddressSanitizer:DEADLYSIGNAL
=================================================================
==11158==ERROR: AddressSanitizer: SEGV on unknown address 0x1000982d4c88 (pc 0x7fd677330fac bp 0x7ffcc1727990 sp 0x7ffcc1726258 T0)
==11158==The signal is caused by a WRITE memory access.
    #0 0x7fd677330fab  /build/glibc-OTsEL5/glibc-2.27/string/../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:185
    #1 0x52fb4f in cppng /home/t/Projects/afl/fuzzing-experiments/subjects/libpng/contrib/tools/pngcp.c:2302:1
    #2 0x529195 in main /home/t/Projects/afl/fuzzing-experiments/subjects/libpng/contrib/tools/pngcp.c:2351:16
    #3 0x7fd6771c3b96 in __libc_start_main /build/glibc-OTsEL5/glibc-2.27/csu/../csu/libc-start.c:310
    #4 0x41b789 in _start (/home/t/Projects/afl/fuzzing-experiments/subjects/libpng/.libs/pngcp+0x41b789)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /build/glibc-OTsEL5/glibc-2.27/string/../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:185 
==11158==ABORTING
```

```
>>> apt-cache show pngtools
Package: pngtools
Architecture: amd64
Version: 0.4-1.3
Priority: optional
Section: universe/graphics
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Nelson A. de Oliveira <naoliv@debian.org>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 64
Depends: libc6 (>= 2.11), libpng16-16 (>= 1.6.2-1)
Suggests: optipng | pngcrush
[...]

PR #294, opened 2019-06-10
Repair of CVE-2019-6129  #269 and issue #255

tangyaofang:
Repair of CVE-2019-6129  #269 and issue #255


ctruta (commented):
This code does not actually fix the underlying problem. It merely moves it from one place to another, and it is still dependent on the specific compiler behaviour.

ctruta:
I am closing this pull request, and discussing the proper solution with @jbowler in issue #255.

Issue #288, opened 2019-05-09
Memory Leak in png2pnm


tangyaofang:
issue： https://github.com/glennrp/libpng/issues/239  unsolved.  The same error in the new version：

root@KWE1000859128:/home/test/libpng-1.6.37/contrib/pngminus# ./png2pnm crash_png2pnm_memory_leak 1.pnm
libpng error: Read Error
PNG2PNM
Error:  unsuccessful conversion of PNG-image

=================================================================
==100062==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 160 byte(s) in 1 object(s) allocated from:
    #0 0x4b8da8  (/home/test/libpng-1.6.37/contrib/pngminus/png2pnm+0x4b8da8)
    #1 0x4ebb16  (/home/test/libpng-1.6.37/contrib/pngminus/png2pnm+0x4ebb16)
    #2 0x4eab4d  (/home/test/libpng-1.6.37/contrib/pngminus/png2pnm+0x4eab4d)
    #3 0x7f42e320582f  (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)

Indirect leak of 2400 byte(s) in 1 object(s) allocated from:
    #0 0x4b8da8  (/home/test/libpng-1.6.37/contrib/pngminus/png2pnm+0x4b8da8)
    #1 0x4eba76  (/home/test/libpng-1.6.37/contrib/pngminus/png2pnm+0x4eba76)
    #2 0x4eab4d  (/home/test/libpng-1.6.37/contrib/pngminus/png2pnm+0x4eab4d)
    #3 0x7f42e320582f  (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)

SUMMARY: AddressSanitizer: 2560 byte(s) leaked in 2 allocation(s).

tangyaofang:
@ctruta :

The issue is:
In the contrib\pngminus\png2pnm.c file, when the png_read_image (png_ptr, row_pointers) statement fails, the previously requested png_pixels row_pointers memory is not released.
I think I have modified this problem and released the requested png_pixels row_pointers memory when an exception is caught.
The following is my modification, please confirm that my modification is correct:

if (setjmp (png_jmpbuf(png_ptr)))
  {
    png_destroy_read_struct (&png_ptr, &info_ptr, NULL);
    if (row_pointers != (unsigned char**) NULL)
    {
      free (row_pointers);
      row_pointers = NULL;
    }      
[...]

Issue #238, opened 2018-07-12
SEGV in function png_free_data


fouzhe:
Hi,all!
We find a bug in **libpng 1.6.34** when using [pngwriter](https://github.com/pngwriter/pngwriter),which take libpng as dependence. 
You can get more information by inferring [this](https://github.com/pngwriter/pngwriter/issues/129)!

ax3l:
Concretely speaking, the [recommended error handling](http://libpng.org/pub/png/libpng-manual.txt) for `png_read_image` via
```C
    if (setjmp(png_jmpbuf(png_ptr)))
    {
       png_destroy_read_struct(&png_ptr, &info_ptr,
           &end_info);
       fclose(fp);
       return (ERROR);
    }
```

results in a SEGV during `png_destroy_read_struct`. We traced our specific call stack [in this comment](https://github.com/pngwriter/pngwriter/issues/129#issuecomment-404244078).

The error occurs [in `png_free_data`](https://github.com/glennrp/libpng/blob/libpng16/png.c#L479-L482) during the free of text-related data with specifically crafted files, reading invalid/unowned memory.

You can test it with simple (non-incremental) reads via @fouzhe's fuzzy-file [here](https://github.com/fouzhe/security/blob/master/pngwriter/crash_SEGV_readfromfile).

ax3l:
@fouzhe can you try if you can reproduce that issue also with `png2pnm` in `contrib/pngminus`? I can't get it there as well but see a memory leak.

fouzhe:
Yeah,I will try it!
@ax3l Did you use the file I attached?

ax3l:
> Did you use the file I attached?

yes, via: `./png2pnm crash_SEGV_readfromfile crash.pnm`

here is my diff to get it compiled with clang-6.0:
<details>
```bash
export CFLAGS="-g -fsanitize=address" LDFLAGS="-fsanitize=address" CC=clang-6.0
./autogen.sh
./configure
cd contrib/pngminus/
make -f makefile.std

./png2pnm crash_SEGV_readfromfile crash.pnm
```
```diff
diff --git a/contrib/pngminus/makefile.std b/contrib/pngminus/makefile.std
index 14e25cd64..553bbe354 100644
--- a/contrib/pngminus/makefile.std
+++ b/contrib/pngminus/makefile.std
@@ -2,7 +2,7 @@
 # Linux / Unix
 
 #CC=cc
-CC=gcc
+CC=clang-6.0
[...]

Issue #239, opened 2018-07-13
Memory Leak in png2pnm


fouzhe:
As mention in [this issue](https://github.com/glennrp/libpng/issues/238#issuecomment-404487829).There is an issue in **png2pnm** in **libpng 1.6**  found by [ax3l](https://github.com/glennrp/libpng/issues/238#issuecomment-404482134).
Those command can be used to compile png2pnm with **Clang 6.0**:

```
export CFLAGS="-g -fsanitize=address" LDFLAGS="-fsanitize=address" CC=clang-6.0
./autogen.sh
./configure
make
cd contrib/pngminus/

make -f makefile.std
```

The `makefile.std` in folder `contrib/pngminus` should also be modified before the command `make -f makefile.std` as mention in [this comment](https://github.com/glennrp/libpng/issues/238#issuecomment-404487829):

```diff
index 14e25cd64..553bbe354 100644
--- a/contrib/pngminus/makefile.std
+++ b/contrib/pngminus/makefile.std
@@ -2,7 +2,7 @@
 # Linux / Unix
 
 #CC=cc
-CC=gcc
+CC=clang-6.0
 LD=$(CC)
 
 RM=rm -f
@@ -13,18 +13,19 @@ RM=rm -f
 #PNGLIBS = $(PNGPATH)/lib/libpng16.a
 PNGINC = -I../..
 PNGLIB = -L../.. -lpng
-PNGLIBS = ../../libpng.a
+PNGLIBS = ../../.libs/libpng16.a
 
-#ZPATH = /usr/local
-#ZINC = -I$(ZPATH)/include
-#ZLIB = -L$(ZPATH)/lib -lz
+ZPATH = /usr/local
+ZINC = -I$(ZPATH)/include
+ZLIB = -L$(ZPATH)/lib -lz
 #ZLIBS = $(ZPATH)/lib/libz.a
-ZINC = -I../../../zlib
-ZLIB = -L../../../zlib -lz
-ZLIBS = ../../../zlib/libz.a
+#ZINC = -I../../../zlib
+#ZLIB = -L../../../zlib -lz
+#ZLIBS = ../../../zlib/libz.a
 
 CPPFLAGS=$(PNGINC) $(ZINC)
-CFLAGS=
+CFLAGS=-g -fsanitize=address
+LDFLAGS=-fsanitize=address
 LDLIBS=$(PNGLIB) $(ZLIB)
 LDLIBSS=$(PNGLIBS) $(ZLIBS)
[...]

PR #295, opened 2019-06-10
Repair of CVE-2018-14048 #238

tangyaofang:
Repair of CVE-2018-14048  #238

### DIFF
--- a/contrib/pngminus/png2pnm.c
+++ b/contrib/pngminus/png2pnm.c
@@ -208,6 +208,16 @@ BOOL png2pnm (FILE *png_file, FILE *pnm_file, FILE *alpha_file,
   if (setjmp (png_jmpbuf (png_ptr)))
   {
     png_destroy_read_struct (&png_ptr, &info_ptr, NULL);
+    if(row_pointers != NULL)
+    {
+      free(row_pointers);
+      row_pointers = NULL;
+    }
+    if(png_pixels != NULL)
+    {
+      free(png_pixels);
+      png_pixels = NULL;
+    }
     return FALSE;
   }
 


heaven-hq:
@ctruta I have verified this PR. When an abnormal situation is entered, it will enter 
```
if (setjmp (png_jmpbuf(png_ptr)))
{
    png_destroy_read_struct (&png_ptr, &info_ptr, NULL);  
    return FALSE;
}
```
branch and cause memory leak.This patch could solve this problem very well. Do you plan to merge this PR?

contrib/pngminus/png2pnm.c
```diff
@@ -208,6 +208,16 @@ BOOL png2pnm (FILE *png_file, FILE *pnm_file, FILE *alpha_file,
   if (setjmp (png_jmpbuf (png_ptr)))
   {
     png_destroy_read_struct (&png_ptr, &info_ptr, NULL);
+    if(row_pointers != NULL)
```
DerDakon, line 211: From what I see this can never be true as row_pointers is only allocated in line 316.

contrib/pngminus/png2pnm.c
```diff
@@ -208,6 +208,16 @@ BOOL png2pnm (FILE *png_file, FILE *pnm_file, FILE *alpha_file,
...
+    {
+      free(row_pointers);
+      row_pointers = NULL;
+    }
+    if(png_pixels != NULL)
```
DerDakon, line 216: You need neither the if() not the assignment: ```free(NULL)``` is fine, and it's a local variable that goes out of scope by the return anyway.

vesalvojdani:
Note: This has nothing to do with the SEGV "vulnerability", but it does fix the ASAN reports of a memory leak within `png2pnm`, which people keep reporting. The fix requires just two lines of freeing `row_pointers` and `png_pixels`. Also compare:

https://github.com/glennrp/libpng/blob/f8e5fa92b0e37ab597616f554bee254157998227/contrib/pngminus/pnm2png.c#L463-L468

This is not a serious leak, but many issues are kept open because of this:
* It was a confounding factor in #238
* Separate issue in #239
* Further complaint that it's not fixed #288

contrib/pngminus/png2pnm.c
vesalvojdani: Note the `setjmp`. This is definitely reached with non-null values when subsequent libpng fail and longjmp back here. This can be confirmed by both ASAN and simple `printf` debugging...

jbowler (commented):
The code in the fix is wrong; row_pointers is not volatile so shall not be referenced after the setjmp returns.

contrib/pngminus/png2pnm.c
jbowler: In fact your are incorrect :-)  row_pointers is an auto and is not marked volatile, so on many systems you are correct and it will always be NULL but on many other systems it can be, well, any value.

jbowler:
@wwwillem you may care to comment.

The program immediately exits after the setjmp returns non-zero, but then why call png_destroy_read_struct?

It's valid in ANSI-C to do so because png_ptr is not modified after the initial call to setjmp (the one that returns 0), but it's misleading as documentation because other autos that are modified after the initial call to setjmp are not freed (and inserting calls to free them will result in warnings on some compilers).

The code as written is completely correct; the program exits after the error and that frees all the memory.

This code is just documentation; someone who copies it and doesn't understand all the issues that surround setjmp is worthy of a CVE but that is still nothing to do with libpng.

NOT a libpng issue @ctruta

ctruta:
I confirm that CVE-2018-14048 is invalid, although I can understand how it is far from obvious that libpng is not leaky. After all, the png and info structures are being cleaned up in both png2pnm.c and pnm2png.c at the end of processing; and yet, there's an ASan report -- and I can understand that it looks suspicious.

In any case (FYI @wwwillem) I just pushed a bunch of rather significant changes to the pngminus suite. See the commits c993ae4c67e99dbed65258412794be5e071b8f95, abb8d4a71fb58e9988d38dd9d726c2e947d0893f, bdbbcaa4578f804734ddc6678fb24378cc18ebcf.

In conclusion:

Neither CVE-2018-14048 nor the memory leak reported by ASan should be problematic any further.

Issue #264, opened 2018-12-25
buffer overflow issue 1#


niugx:
there is a buffer overflow issue in pngfix.c

line 3985            else if (suffix != NULL)
line 3986               memcpy(temp_name, *argv, outlen);
line 3987
line 3988            temp_name[outlen] = 0;

[...]

PR #296, opened 2019-06-10
Repair of issue 264

tangyaofang:
Repair of issue #264


ctruta:
@jbowler, for not a very good reason, I postponed the integration of a bunch of fixes implemented by @tangyaofang, and then I forgot about them. (Apologies.)

Could you please review this commit and advise on whether to integrate it? The fix looks good to me, but I would appreciate your opinion also.

jbowler:
This is a duplicate of #294

jbowler:
The fix is not likely to work given the problem.

ctruta:
I see. Admittedly, I haven't tested it. I'm integrating the other solution, proposed by @thealberto in #350.

PR #303, opened 2019-07-04
fix(cmake): use correct ZLIB_* variables

ledocc:
ZLIB_LIBRARIES and ZLIB_INCLUDE_DIRS are official cmake variable name
ZLIB_LIBRARY should be empty


arsnyder16:
@ledocc @ctruta  This also needed to change in `genout.cmake.in`

https://github.com/glennrp/libpng/blob/libpng16/scripts/genout.cmake.in

PR #304, opened 2019-07-08
B windowbits

tangyaofang:
Fix comments.

### DIFF
--- a/contrib/tools/png-fix-itxt.c
+++ b/contrib/tools/png-fix-itxt.c
@@ -10,7 +10,7 @@
  *
  * Usage:
  *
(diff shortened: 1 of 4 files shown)


ctruta:
Thank you for your submission. I'm rejecting your proposed commit d338a7ee3bf9baca6e70900f08fc5e5b0bcb1236 because the underscore in window_bits is not strictly necessary. The comment may be in English prose (i.e. "the window bits") and is not necessarily the name of a variable. See also, for example, "the memory level".

ctruta:
Your other commit 8e69526a72e5fc68446939d1d6b6f8e9d12a287d was integrated, thank you.

PR #306, opened 2019-07-10
Fix a bug in png_set_compression_strategy.

tangyaofang:
Fix a bug in png_set_compression_strategy.

### DIFF
--- a/pngwrite.c
+++ b/pngwrite.c
@@ -1184,6 +1184,17 @@ png_set_compression_strategy(png_structrp png_ptr, int strategy)
    if (png_ptr == NULL)
       return;
 
+   if(strategy > 4)
+   {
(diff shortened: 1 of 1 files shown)


ctruta:
Thank you for your contribution, but the original code in `pngwrite.c` is not checking the validity of the zlib strategy, by design, because that is the job of zlib.

On one hand, libpng is supposed to be "zlib-agnostic" (except in peculiar places like `png_set_compression_window_bits` where the hard-coded limits are there to ensure the validity of a PNG-compliant implementation); and on the other hand, zlib may be, at any time, upgraded with new strategies that should automatically be available to the libpng developers. With your proposed change, any future zlib strategy would be unavailable, unless the libpng code would be, separately, updated -- and that is not desirable.

PR #308, opened 2019-07-11
Fix a bug in png_set_compression_level.

tangyaofang:
Fix a bug in png_set_compression_level.


ctruta:
I'm rejecting this change to `png_set_compression_level` for the same reason as the other change to `png_set_compression_strategy`. It would be possible to add new, higher, stronger compression levels to any future zlib version, and those levels would be expected to produce compliant deflate streams just as well.

PR #314, opened 2019-07-30
rename MACRO gv name to GV

fredgan:
When I read the code,  the following code really confuse me,

```c
static int
cppng(struct display *dp, const char *file, const char *gv dest)
```
I spend some time to find that gv is a macro.  Macro should be written in UPPER CASE.

### DIFF
--- a/contrib/tools/pngcp.c
+++ b/contrib/tools/pngcp.c
@@ -90,9 +90,9 @@
     * within the scope of a setjmp may be clobbered.  This hacks round the
     * problem (sometimes) without harming other compilers.
     */
-#  define gv volatile
+#  define GV volatile
 #else
-#  define gv
+#  define GV
 #endif
 
 /* 'CLOCK_PROCESS_CPUTIME_ID' is one of the clock timers for clock_gettime.  It
@@ -2275,7 +2275,7 @@ cp_one_file(struct display *dp, const char *filename, const char *destname)
 }
 
 static int
-cppng(struct display *dp, const char *file, const char *gv dest)
+cppng(struct display *dp, const char *file, const char * GV dest)
    /* Exists solely to isolate the setjmp clobbers which some versions of GCC
     * erroneously generate.
     */


ctruta:
I agree in principle with the proposed change, although, @jbowler, do we still need this GV hack at all?

I tried various versions of gcc, old and new, 32-bit and 64-bit, on Linux and on Windows, and all of them seemed to pass without any warning about longjmp clobbering function arguments. I am inclined to delete this hack altogether.

jbowler:
I don't use upper-case for macros which are type-like.  Libpng has no rules about the case of such macros because it doesn't define any - types are defined with typedef.   There really aren't any hard rules about macro case and there can't be because of the use of case in chunk types.

In this case, however, I agree with Cosmin's _second_ comment.   I just tested it with gcc-12.2.1 and gcc-11.3.1 (the only two I currently have) and, using -Wall -Wextra, neither produce warnings.   The 'volatile' in there is undesirable; I put it in because GCC versions kept on producing the error but it is unnecessary; the warning is incorrect.

I only tested by removing the 'volatile' - I didn't try eliminating the whole of cppng which was, I believe, the original implementation before the GCC nonsense.

ctruta:
> In this case, however, I agree with Cosmin's _second_ comment. I just tested it with gcc-12.2.1 and gcc-11.3.1 (the only two I currently have) and, using -Wall -Wextra, neither produce warnings. The 'volatile' in there is undesirable; I put it in because GCC versions kept on producing the error but it is unnecessary; the warning is incorrect.

Thank you. I, too, agree that the warning is incorrect, because it can (and should) be ruled out 100% at compile time.

> I only tested by removing the 'volatile' - I didn't try eliminating the whole of cppng which was, I believe, the original implementation before the GCC nonsense.

Yeah, `cppng` gotta go as well. @jbowler, will you do the honors, or will I do the honors?

While at it, take a look at commits like e2bb5e7512 (and others), where I deleted the "Last changed in libpng x.y.z" lines at the top of source files. Somehow I haven't gotten to all of contrib/, and especially to contrib/tools/ and scripts/. That old convention (to say when the file was modified) used to make sense ages ago, when we distributed tarballs and zipfiles only; but now it only gets in the way, especially in the way of commands like `git log -p` and `git diff`.

jbowler:
> Yeah, `cppng` gotta go as well.

I took a look at that before and I came to the conclusion that, regardless of the original motivation, the code is more clear with cppng there.  I also regard it as good practice to do "setjmp" in as simple a function as possible; the ANSI-C formalisation of setjmp revealed a lot of problems in existing code.  Indeed UNIX, particularly the AT&T version, only worked because the portable C compiler always wrote local variables to the stack before calling a function and that was unacceptable performance-wise on machines with larger register sets such as the ARM (10 32 bit registers) and, ironically, the Hobbit (256 bytes of registers located at the top of the stack).

It's a significant rewrite to undo cppng and, because "dp" is a local variable in main() it might throw up other instances of the GCC issue.  There's no perf overhead to the function call with a good compiler (it will inline it because it's static.)   With cppng present the GCC warning is so obviously bogus that it's reasonable to say "won't fix" to any bug reports if it reappears.

ctruta:
@jbowler are you good with removing this `#define gv volatile` workaround, entirely?

The old (buggy) gcc might have needed this workaround, but the present-day gcc and clang (and others) don't. I checked the old (gcc version 6 on Debian 9) and the new (gcc version 13 on Debian 12) on x86 and x86_64, with `-Wextra` enabled, unoptimized and optimized.

jbowler:
> @jbowler are you good with removing this `#define gv volatile` workaround, entirely?

Yes.  Do you want a pull request (it's only a 10-line delete plus the one line change on the definition of cppng())?

ctruta:
@jbowler yah, those changes, plus the update of your copyright year.

I just pushed a few unrelated commits, about 10min ago, and I'm closing this PR because it's no longer applicable.
Rebase, send me a new PR, thanks in advance.

PR #320, opened 2019-08-30
Fix CMake build with Emscripten

tronical:
When building with Emscripten, linkage to libm is neither required nor
does it will. As find_library will fail, M_LIBRARY will be set to
NOTFOUND and the build would fail later. The build works just fine
without libm linkage.


tronical:
Thank you very much :)

PR #321, opened 2019-09-16
Fix use after free.

eliteraspberries:


ctruta:
Looking further into this: there's no danger of use-after-free, because there is no continuation from png_error() at pngpread.c:502. If you found this error with a static analyzer, you need to show that analyzer, somehow, that png_error() is marked as PNG_NORETURN.

PR #328, opened 2020-01-14
Optimize png16 with loongson mmi for 64-bit os

XiWeiGu:
Add Optimizations for loongson platform(3A3000, 3A4000, etc.), make check passed.

### DIFF
--- a/contrib/oss-fuzz/libpng_read_fuzzer.cc
+++ b/contrib/oss-fuzz/libpng_read_fuzzer.cc
@@ -78,6 +78,22 @@ void user_read_data(png_structp png_ptr, png_bytep data, size_t length) {
   buf_state->data += length;
 }
 
+void* limited_malloc(png_structp, png_alloc_size_t size) {
+  // libpng may allocate large amounts of memory that the fuzzer reports as
+  // an error. In order to silence these errors, make libpng fail when trying
+  // to allocate a large amount. This allocator used to be in the Chromium
+  // version of this fuzzer.
+  // This number is chosen to match the default png_user_chunk_malloc_max.
+  if (size > 8000000)
+    return nullptr;
+
+  return malloc(size);
+}
+
+void default_free(png_structp, png_voidp ptr) {
+  return free(ptr);
+}
+
 static const int kPngHeaderSize = 8;
 
 // Entry point for LibFuzzer.
@@ -117,6 +133,9 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
     PNG_CLEANUP
     return 0;
   }
+
+  // Use a custom allocator that fails for large allocations to avoid OOM.
+  png_set_mem_fn(png_handler.png_ptr, nullptr, limited_malloc, default_free);
 
   png_set_crc_action(png_handler.png_ptr, PNG_CRC_QUIET_USE, PNG_CRC_QUIET_USE);
 #ifdef PNG_IGNORE_ADLER32


configure.ac
```diff
@@ -329,6 +332,11 @@ AC_ARG_ENABLE([hardware-optimizations],
               AC_DEFINE([PNG_ARM_NEON_OPT], [0],
                 [Enable ARM_NEON optimizations])
               ;;
+            mips64el*)
```
wzssyqa: This will force all mips64el* to use Loongson's MMI, and disable MSA.
This is not a good idea, even consider for Loongson only (and want to make trouble for generic MIPS).
since Loongson 3A 4000 supports MSA.

Why do you want to add mips64el here and  keep the bellow mipsel*|mips64el*)?
I don't think that MMI has such restriction.

configure.ac
```diff
@@ -437,6 +445,53 @@ AM_CONDITIONAL([PNG_MIPS_MSA],
...
+   [case "$enableval" in
+      no|off)
+         # disable the default enabling on __loongson_mmi systems:
+         AC_DEFINE([PNG_LOONGSON_MMI_OPT], [0],
+                   [Disable LOONGSON MMI optimizations])
+         # Prevent inclusion of the assembler files below:
+         enable_loongson_mmi=no;;
+      check)
+         AC_DEFINE([PNG_LOONGSON_MMI_CHECK_SUPPORTED], [],
+                   [Check for LOONGSON MMI support at run-time]);;
```
xen0n: There's no corresponding check implemented in `mips/mips_init.c`, like the MSA implementation did.

configure.ac
```diff
@@ -437,6 +445,53 @@ AM_CONDITIONAL([PNG_MIPS_MSA],
...
+      api)
+         AC_DEFINE([PNG_LOONGSON_MMI_API_SUPPORTED], [],
+                   [Turn on LOONGSON MMI optimizations at run-time]);;
```
xen0n: Similarly this. There's actually no run-time switch governing Loongson MMI code blocks, instead it seems all compile-time.

pngpriv.h
```diff
@@ -23,6 +23,7 @@
 #ifndef PNGPRIV_H
 #define PNGPRIV_H
 
+
```
xen0n: Diff damage.

pngpriv.h
```diff
@@ -195,6 +196,14 @@
 #  endif
 #endif
 
+#ifndef PNG_LOONGSON_MMI_OPT
+#  if defined(__loongson_mmi) && defined(_MIPS_ARCH_LOONGSON3A) && defined(PNG_ALIGNED_MEMORY_SUPPORTED)
```
xen0n: This won't match if `-march=(gs464|gs464e|gs264e)`, for which compiler support is already upstreamed. Actually you don't have to double-check `_MIPS_ARCH` here, if a processor implements Loongson MMI and environment is n64 you should be OK.

loongson-mmi/loongson_mmi_init.c
```diff
@@ -0,0 +1,57 @@
...
+   }
+   else if (bpp == 6)
+   {
+      pp->read_filter[PNG_FILTER_VALUE_SUB-1] = png_read_filter_row_sub6_mmi;
+   }
+   else if (bpp == 8)
+   {
+      pp->read_filter[PNG_FILTER_VALUE_SUB-1] = png_read_filter_row_sub8_mmi;
+   }
+   /* No need optimize PNG_FILTER_VALUE_UP.  The compiler should
```
xen0n: Fix English here.

configure.ac
```diff
@@ -437,6 +445,53 @@ AM_CONDITIONAL([PNG_MIPS_MSA],
...
+         AC_MSG_WARN([--enable-loongson-mmi: please specify 'check' or 'api', if]
+            [you want the optimizations unconditionally pass '-march=loongson3a']
+            [to the compiler.]);;
+      *)
+         AC_MSG_ERROR([--enable-loongson-mmi=${enable_loongson_mmi}: invalid value])
+   esac])
+
+# Add MIPS specific files to all builds where the host_cpu is mips ('mips*') or
+# where MIPS optimizations were explicitly requested (this allows a fallback if a
+# future host CPU does not match 'mips*')
```
xen0n: If a future host doesn't use the `mips*` triple, then perhaps modify the logic at that future date (should it REALLY happen). Otherwise this just complicates matters for others in the meantime, not to mention the supposed arch fork may well never happen.

loongson-mmi/filter_mmi_intrinsics.c
```diff
@@ -0,0 +1,590 @@
...
+   __asm__ volatile (
+       "1:                                          \n\t"
+       "ldc1   %[rp],       0x00(%[row])            \n\t"
+       "ldc1   %[pp],       0x00(%[prev_row])       \n\t"
+       "paddb  %[rp],       %[rp],            %[pp] \n\t"
+       "sdc1   %[rp],       0x00(%[row])            \n\t"
+
+       "daddiu %[row],      %[row],           0x08  \n\t"
+       "daddiu %[prev_row], %[prev_row],      0x08  \n\t"
+       "daddiu %[istop],    %[istop],        -0x08  \n\t"
```
xen0n: These are 64-bit-capable-ABIs-only (which excludes o32 evidently), yet neither here nor the place defining `PNG_LOONGSON_MMI_IMPLEMENTATION` checks for this, which will break o32 builds.

FlyGoat:
You're leaving CMake building scripts untouched.

configure.ac
gxw-loongson: thx, if "./configure --enable-hardware-optimizations", it will force all mips64el* to use Loongson's MMI. I will fix it.

configure.ac
gxw-loongson: thx, i wll add run-time check.

ctruta:
Hello, and thank you very much for your contribution. I delayed the integration of LOONGSON support into libpng, but I haven't forgotten about it.

If you are still interested in making this contribution, and are able to bring it up to date, I would highly appreciate it.

XiWeiGu:
Thank you for your reply. Recently, I'll take some time to update a version.

PR #330, opened 2020-02-04
Avoid -Wundef warnings for non-x86 architectures too

vadz:
Define this symbol as 0 instead of leaving it undefined when not using
ARM NEON optimizations.

No real changes, but just avoid a bunch of

"PNG_ARM_NEON_IMPLEMENTATION" is not defined, evaluates to 0 [-Wundef]

warnings when building the library.


ctruta:

PR #337, opened 2020-07-08
docs: Add spaces around '/*' and '*/'

XMuli:
Like most other Qt coding styles, the necessary spaces should be added
to the sides of the '/*' and '/*'


ctruta:

PR #338, opened 2020-07-28
Adding a check

klebertarcisio:
Adding a check as in the function we have in the update_display function


ctruta:
Patch applied. Thank you very much.

PR #339, opened 2020-07-28
Avoiding using a null reference

klebertarcisio:
Value may be null and it should be checked before dereferencing.


ctruta:
Hello, and thank you for your contribution.

You will find that early on, in the while loop, inside the `buffer_read` function, the value of `last` should not be null. That value is, nevertheless, checked, although such situation would be abnormal -- see the two `/*NOTREACHED*/` comments -- and the loop is exited.

The original code is, therefore, secure.

PR #340, opened 2020-07-28
Insufficient memory

klebertarcisio:
We should check all results returned by **png_voidcast** function

Regards.


benkasminbullock:
`png_malloc` already checks for NULL:

https://github.com/glennrp/libpng/blob/libpng16/pngmem.c#L171-L185

If there is no memory it produces an error:

https://github.com/glennrp/libpng/blob/libpng16/pngmem.c#L182

ctruta:
Thank you. As @benkasminbullock pointed out, `png_malloc` already checks for allocation errors inside, and it never returns a null pointer.

PR #351, opened 2020-12-11
Remove second call to write_eXIf

benkasminbullock:
The second call to write_eXIf doesn't check whether the first call
succeeded; no other chunks except tIME and the text chunks seem to be
doubled like this.

The second call causes the eXIf chunk to incorrectly be written twice.


ctruta:
Thank you for all the contributions, and especially for the EXIF fix.

lrosenth:
Bug seems still  to be in Version 1.6.37...

PR #354, opened 2020-12-14
cmake: Fix cross-compilation on macOS

orudge:
When cross-compiling on macOS (e.g., on Intel for arm64 or vice versa) using the CMAKE_OSX_ARCHITECTURES variable, the CMAKE_SYSTEM_PROCESSOR variable does not get updated with the target architecture; instead it uses the host CPU architecture (or, in some versions of cmake, the architecture of the `cmake` binary itself, which may be different).

This appears to be a cmake bug/design decision, and may be a consequence of the fact that you can specify multiple architectures in the CMAKE_OSX_ARCHITECTURES variable to generate universal/fat binaries. However, it does cause problems when trying to cross-compile libpng with hardware optimisations enabled, as the incorrect architecture-specific files or flags get included/enabled.

This pull request fixes cross-compilation (and native compilation, when using an older version of cmake on Apple Silicon) on macOS, and should have no effect on other platforms. Note that this will likely fail if the user tries to build a universal binary (e.g., `CMAKE_OSX_ARCHITECTURES=arm64;x86_64`), but that would have been the case before this patch too. The only way to build libpng as a universal binary with hardware optimisations would be to perform two separate builds then `lipo` them together.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,11 +65,22 @@ option(PNG_HARDWARE_OPTIMIZATIONS "Enable hardware optimizations" ON)
 set(PNG_PREFIX "" CACHE STRING "Prefix to add to the API function names")
 set(DFA_XTRA "" CACHE FILEPATH "File containing extra configuration settings")
 
+# CMake currently sets CMAKE_SYSTEM_PROCESSOR to one of x86_64 or arm64 on macOS,
+# based upon the OS architecture, not the target architecture. As such, we need
+# to check CMAKE_OSX_ARCHITECTURES to identify which hardware-specific flags to
+# enable. Note that this will fail if you attempt to build a universal binary in
+# a single cmake invocation.
+if (APPLE AND CMAKE_OSX_ARCHITECTURES)
+  set(TARGET_ARCH ${CMAKE_OSX_ARCHITECTURES})
+else()
+  set(TARGET_ARCH ${CMAKE_SYSTEM_PROCESSOR})
+endif()
+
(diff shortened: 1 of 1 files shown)


rdb:
I think a better fix would be to check if CMAKE_OSX_ARCHITECTURES *contains* arm64, and if necessary check via the preprocessor definitions whether ARM support is actually available for the current target.

vespakoen:
How about turning optimisations off when cross compiling:

```
set(TARGET_ARCH ${CMAKE_SYSTEM_PROCESSOR})
if(APPLE AND CMAKE_OSX_ARCHITECTURES)
  list(LENGTH CMAKE_OSX_ARCHITECTURES ARCHITECTURE_COUNT)
  # if we are cross compiling, we cannot do any optimizations unless support
  # is tested for at compile time, which is not possible at the moment
  if(ARCHITECTURE_COUNT GREATER 1)
    set(PNG_HARDWARE_OPTIMIZATIONS OFF)
  elseif(ARCHITECTURE_COUNT EQUAL 1)
    # override the architecture we are testing for with the CMAKE_OSX_ARCHITECTURES value
    set(TARGET_ARCH ${CMAKE_OSX_ARCHITECTURES})
  endif()
endif()
```

I don't think we can do any optimizations when cross compiling, I could be compiling arm64 on a x86_64 machine, and when testing for arm64 features it will just fail?

orudge:
In the case of macOS we are compiling for a known platform in terms of ARM64 capabilities. Of course, this may not be the case for other non-Apple platforms.

ctruta:
Integrated. Many thanks for your contribution.

PR #364, opened 2021-02-13
pdf document 'libpng-manual_1.6.37.pdf' created from 'libpng-manual.t…

fSQRL:
pdf document 'libpng-manual_1.6.37.pdf' created from 'libpng-manual.txt' ;
intermediate LibreOffice document 'libpng-manual_1.6.37.odt' put in new directory 'contrib/doc-src' for further editing


fSQRL:
pdf document not in final state (page breaks must be revised)
it's the 'pure' libpng-manual.txt reworked in LibreOffice
opening .odt file permits futher editing

ctruta:
Thanks, but we'd rather avoid storing binary artifacts (other than PNG files) in the repo, especially if they're documents that need to be opened in an external editor for corrections. Maintenance of such files would not be sustainable in the long term.

On the other hand, conversions of the libpng manual to a modern lightweight markup format like Markdown or Asciidoc, which can then be further converted automatically to PDF or XPS or ODT, etc., would be something worth considering.

PR #375, opened 2021-04-15
Update pngpread.c

mscarfo:


benkasminbullock:
This and the other pull request both look like just adding blank lines, did something go wrong?

PR #378, opened 2021-04-30
fix dangerous undefined behavior.

ihsinme:
unfortunately your construction is subject to undefined behavior in terms of the sequence of function calls. which can lead to incorrect work.
I propose to fix this point.

### DIFF
--- a/contrib/tools/pngfix.c
+++ b/contrib/tools/pngfix.c
@@ -1005,10 +1005,11 @@ file_end(struct file *file)
 
    if (file->out != NULL)
    {
-      /* NOTE: this is bitwise |, all the following functions must execute and
-       * must succeed.
-       */
-      if (ferror(file->out) | fflush(file->out) | fclose(file->out))
+      int ef = 0;
+      ef += ferror(file->out);
+      ef += fflush(file->out);
+      ef += fclose(file->out);
+      if (ef != 0)
       {
          perror(file->out_name);
          emit_error(file, READ_ERROR_CODE, "output write error");


benkasminbullock:
While I don't think that chaining function calls using a bitwise operator is at all a good practice, I think it would be better to give the commit a message such as "order of evaluation is undefined here". Also note that this is part of the contrib/tools directory not libpng itself.

ihsinme:
if I have to do something else. let me know.

ihsinme:
good day.
have you made a decision on this PR?

ValZapod:
> undefined behavior in

Unspecified behaviour. Also, with comma operator function calls are left from right, so why here it is lso like that?

ihsinme:
good day.
if you mean to use comma-separated function calls in a condition, then this will not allow you to trace the return value.

ValZapod:
Are there any side effects? I mean, is file->out modifed? If not, I think it will evaluate from left to right.

ihsinme:
there are no side effects when using the call `f1() ,f2(), f3()`.
but maybe I do not quite understand you correctly.
if you want to do so
`if (f1 (), f2 (), f3 ())`
then it is similar
`f1 ();`
`f2 ();`
`if (f3 ())`
`

ihsinme:
ping

ihsinme:
ping

landfillbaby:
fclose flushes so the fflush can be removed completely

ValZapod:
True, morover "Note that fflush() only flushes the user space buffers
provided by the C library. To ensure that the data is physically stored on
disk the kernel buffers must be flushed too, for example, with sync(2) or
fsync(2)."

jbowler (changes requested):
Defiantly one of my weirder pieces of code but the comment is correct: it doesn't matter what order the functions are executed in just so long as they all succeed.  Unfortunately while ferror/fflush don't matter if fclose is executed before one or other the "value of the pointer... is indeterminate".

The problem is that fclose must always happen and must happen at the end, so a sequence point is required between ferror/fflush and fclose.  The aim of the code was to do this in a neat way and the suggested patch isn't neat.

It's a constant and entertaining issue with the fclose semantics, obviously I got the solution wrong this time but I'd like to see the best solution, something people can copy, claim their own and feel proud of.

ValeZAA:
You cannot do ferror after fflush, and | operator does not short curcuit like || does, so it can be any order, gcc in fact has such a quirk to randomize the calls.

Also, ffclose DOES fflush internally.

jbowler:
> You cannot do ferror after fflush,

ferror works after fflush; fflush does not clear the 'error indicator'.  It's ambiguous whether or not it sets it.

|| is not applicable because it "shortcuts" - that's the point.  The problem is establishing a sequence point before fclose.

> Also, ffclose DOES fflush internally.

But it is not unambiguously defined to return an error code if the output flush fails; note the use of "or" in the standard before the "," in the discussion of the return from fclose.  That's always been the core of the problem and it reflects the prior implementations; had the committee been able to require fclose to return an error if there was an error in the flush they would have used the word "and" there but some existing implementations did not do that.

So this is what makes the problem so interesting; even though C was standardized the standard didn't make the code to handle close and error detection any easier and a lot of code gets it wrong, e.g. doing an fclose and relying on it to do an fflush and report any errors.

It's also not completely unambiguous whether fflush sets the error indicator, logically based on 5.9.1 it should but only fputc and fgetc document setting it!  Otherwise it would not be necessary to check the fflush return code.

Back to what I said; **come up with a compelling piece of code that everyone will use!**  I've not been able to do it (as evidenced by the piece of code in question) but I can certainly do it without a variable, or a ";" (think how many ways there are to establish a sequence point; || certainly isn't the only option here ;-)

The sequence point (before fclose) is, I believe, the key.  It has to be there.  ferror/fflush evalulation order does not matter; fflush unambiguously returns a failure (not zero) code if it fails.

My best attempt so far still requires two calls to fclose, so it's not compelling.

ValeZAA:
>The sequence point (before fclose) is, I believe, the key.

Yes, you are not allowed to do anything after pointer/ file handle is destroyed. Key issue.

>My best attempt so far still requires two calls to fclose, so it's not compelling.

Well, why not?

>some existing implementations did not do that.

Well, do you care about any implementation besides glibc? Maybe Apple and Microsoft?.

jbowler:
> Yes, you are not allowed to do anything after pointer/ file handle is destroyed.

I've already submitted a PR for @ctruta's consideration with my own fix however I'm interested in knowing why you make that statement, given that the libpng build is **required** to be done with a C89 (ISO-C90) compliant compiler.  I am not saying that, for example, calling ferror after fclose will result in anything useful but the standard doesn't seem to say what you are saying (perhaps you could provide your own interpretation, with justification of course) and the rationale has little to nothing to say on the matter.

landfillbaby:
#513 is better imo

jbowler:
> #513 is better imo

I disagree quite strongly, what are your problems?

(Cf Eliza)

jbowler:
> #513 is better imo

Yes, but why?  I can see several reasons, but I wrote PR513.

landfillbaby:
this should be closed now, since #513 was merged

PR #381, opened 2021-05-25
Add libpng:: namespace

h3ndrk:
This PR adds a `libpng::` namespace to `CMakeLists.txt`. I'm pasting/quoting the answer of https://stackoverflow.com/a/48526017:

The [cmake-developer documentation](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html?highlight=namespace) gives the following advice on namespaces:

> When providing imported targets, these should be namespaced (hence the `Foo::` prefix); CMake will recognize that values passed to `target_link_libraries()` that contain `::` in their name are supposed to be imported targets (rather than just library names), and will produce appropriate diagnostic messages if that target does not exist (see policy *CMP0028*).

And the [*CMP0028* policy documentation](https://cmake.org/cmake/help/latest/policy/CMP0028.html) says on the "common pattern" in the use of namespaces:

> The use of double-colons is a common pattern used to namespace `IMPORTED` targets and `ALIAS` targets. When computing the link dependencies of a target, the name of each dependency could either be a target, or a file on disk. Previously, if a target was not found with a matching name, the name was considered to refer to a file on disk. This can lead to confusing error messages if there is a typo in what should be a target name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@
 # Revised by Alex Gaynor, 2020
 # Revised by Owen Rudge, 2020
 # Revised by Gleb Mazovetskiy, 2021
+# Revised by Hendrik Sieck, 2021
 # Revised by Christopher Sean Morrison, 2022
 
 # This code is released under the libpng license.
@@ -1031,6 +1032,7 @@ endif()
 # Create an export file that CMake users can include() to import our targets.
 if(NOT SKIP_INSTALL_EXPORT AND NOT SKIP_INSTALL_ALL)
   install(EXPORT libpng
+          NAMESPACE "libpng::"
           DESTINATION lib/libpng
           FILE lib${PNG_LIB_NAME}.cmake)
 endif()


ctruta (commented):
Thank you, Hendrik, for your contribution.

Could you please add your full name at the top of the CMakeLists.txt file, near the names of the other contributors?

Could you please also amend your commit so that your full name appears in the Author and Committer field?

As a general rule, I would like all future libpng contributors to include their full names, for copyright purposes.

h3ndrk:
Regarding your new general rule, I'm not very happy that I am forced to expose my full name in public. Also I don't see any reason adding names to files because that's the whole reason why we have git. Nevertheless, since I would like to finally get this merged, I implemented your requested changes. In addition, I rebased to the latest `master`. I hope everything is covered from my side.

theta682:
@ctruta probably we can improve CMake packaging see #475 too. At the moment the standard [FindPNG Module](https://cmake.org/cmake/help/v3.1/module/FindPNG.html) can't properly find headers because it expects them in `include/libpng` instead of `include/libpng16`

ctruta:
@h3ndrk @theta682 I'm very sorry for not fully understanding the compatibility implications of your PR. I wish I knew more about CMake than I actually do.

I specifically could not understand what will happen if I changed the namespace, from the global one, to `libpng::`. Now that we are preparing for a new version line, post-1.6.x, I guess that the `PNG::` namespace could be added. (Please note: the namespace is `PNG::` not `libpng::`.

@bebuch FYI ^

h3ndrk:
@ctruta How shall we proceed? Should I rename the prefix to `::PNG` and rebase to the latest master or something else? Or did you already implement it in your 1.6 release preparations? Please give more concrete instructions, thanks! 😊

touraill-adobe:
I'm a bit confused: there is already a `PNGConfig.cmake` file exported, with a proper namespace (`PNG::`); what is the purpose of the _second_ config file `libpng16.cmake`?

PR #388, opened 2021-07-27
Fix an instance of -Wunused-but-set-variable.

pkasting:
Bug: chromium:1203071


ctruta:

PR #391, opened 2021-08-08
Fix shared build on Android

vkrause:
The ld version script is not generated on Android, so also don't try to use it there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -536,7 +536,7 @@ if(PNG_SHARED)
   endif()
   target_link_libraries(png ${ZLIB_LIBRARIES} ${M_LIBRARY})
 
-  if(UNIX AND AWK)
+  if(UNIX AND AWK AND NOT ANDROID)
     if(HAVE_LD_VERSION_SCRIPT)
       set_target_properties(png PROPERTIES
                             LINK_FLAGS "-Wl,--version-script='${CMAKE_CURRENT_BINARY_DIR}/libpng.vers'")


AJenbo:
@ctruta looks like Xcode is affected by the same issue.

ctruta (changes requested):
It is correct that the ld version script is not generated on Android, but this means that `HAVE_LD_VERSION_SCRIPT` was initialized incorrectly. The real error is a few lines up.
The correct fix would be to disable the auto-detection of the ld version script, and implicitly the initialization of `HAVE_LD_VERSION_SCRIPT`, under Android. See the comment:
```
# Check if LD supports linker scripts.
```
That section should be avoided, not only on `APPLE`, but also on `ANDROID`.

ctruta:
Considering that I want to publish a new libpng release in a few days, I reworked the fix and I pushed it to the master branch. Thank you @vkrause for your original contribution, I gave you credit in the commit message.
See https://github.com/glennrp/libpng/commit/dd8db8105c021637c7d7c7504c83c3c57c03f130

vkrause:
> Considering that I want to publish a new libpng release in a few days, I reworked the fix and I pushed it to the master branch. Thank you @vkrause for your original contribution, I gave you credit in the commit message. See [dd8db81](https://github.com/glennrp/libpng/commit/dd8db8105c021637c7d7c7504c83c3c57c03f130)

Thank you, and sorry for not getting to applying your suggested changes earlier. I tested this here meanwhile and it works as expected. A new release would of course be much appreciated :)

Issue #179, opened 2017-09-12
libpng configuration package is completely broken


rcdailey:
So right now, the CMake script attempts to install exports for libpng but they're not usable by `find_package()` because they are not following the proper rules [documented by CMake][1]. I'm happy to fix this for you, but I want to make you aware of certain changes that need to happen.

The fact that the project is named "libpng" instead of "png" will make things quirky on linux because the 'lib' part is usually implied on that platform. Since it doesn't seem reasonable to rename everything to just "png" (unless you are ok with this; it's a lot of refactoring but possible to do), the config package installation logic will be inconsistent.

In most cases, users will be expected to do this:

    find_package(png CONFIG)

This will trigger the config package search for libpng, as opposed to the "FindPNG.cmake" that comes with CMake (this is the module package structure which is the older way of doing it).

There are 2 main requirements that I think you want:

1. New installations of libpng should remain structurally backward compatible with module package (i.e. people that use FindPNG.cmake should still be able to do so)
2. New installations of libpng should install config and starget scripts in CMake's deterministic structure so that config package searches work (this will be a new feature since the current way it is being done is non-functional).

When users do `find_package(png CONFIG)`, CMake searches for a `png-config.cmake`, which will in turn do an `include()` to include `png-targets.cmake`. So the final install structure would look like:

Under `${CMAKE_INSTALL_PREFIX}/lib/png/[cmake/]`:

```
png-config.cmake
png-targets.cmake
```

You're already sort-of doing `png-config.cmake`, but the name is incorrect. You're installing the name as `libpng`, which will require users to do `find_package(libpng)`, which breaks interoperability with the module package style.

How much flexibility do I have in restructuring and renaming things from the installation logic standpoint? I want to fix all of this but I need to know what I can and can't do. Happy to go over detail items if you want.

There's a lot of weirdness with the library naming as well, but I recognize for backward compatibility reasons, you probably can't allow the naming convention for the libraries to change. Normally the VERSION and SOVERSION symlinking and management is done for you by CMake, if you set it up right. Right now, the script assumes responsibility for doing all of that by hand.

[1]: https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html

skirk:
I too would like to see the changes suggested by @rcdailey to get into repo.
As a solution for now I've added few lines to my (forked) CMakeLists:
```
#install Config file for find_package calls in user projects 
[...]

PR #396, opened 2021-09-30
CMake: Generate pkg-config files for all Windows builds

fanc999:
Hi,

This PR is opened to support generating pkg-config files also for native (non-Cygwin) Windows builds, in addition to MinGW/mingw-w64 builds, since the Meson build system (which uses pkg-config and CMake to find dependencies for most packages) is becoming more widespread for building items such as Cairo and the GNOME/GTK stack and the current CMake config files generated by the CMake builds are not usable by Meson as a consequence of #179, which should be fixed in a PR of its own.

With blessings, thank you!


ctruta:
Thank you, looks neat, but could you please fix the following build error? I tested it on macOS, and it fails with the Ninja generator and the Release target, as follows:
```
CMake Error at cmake_install.cmake:90 (file):
  file INSTALL cannot find
  "/Users/ctruta/src/libpng/build/libpng16-config": No such file
  or directory.

FAILED: CMakeFiles/install.util
```

BotellaA:
Any update on this? It will be very helpful!

fanc999:
Hi @ctruta,

Sorry for the delay, as I was away and this PR was out for some time... I rebased some items, so I think things should probably be remedied, hopefully, since I don't have a macOS devenv.

CMakeLists.txt
```diff
@@ -931,30 +931,42 @@ endif()
...
-                 @ONLY)
-  create_symlink(libpng.pc FILE ${PNGLIB_NAME}.pc)
   configure_file(${CMAKE_CURRENT_SOURCE_DIR}/libpng-config.in
                  ${CMAKE_CURRENT_BINARY_DIR}/${PNGLIB_NAME}-config
                  @ONLY)
   create_symlink(libpng-config FILE ${PNGLIB_NAME}-config)
+else()
+  STRING (TOLOWER "${CMAKE_BUILD_TYPE}" config_lower)
+  if (config_lower STREQUAL "debug")
+    set(LIBS    "${ZLIB_LIBRARY_DEBUG} ${M_LIBRARY}")
```
ctruta, line 947: I'm not an expert at how CMake libraries are being exported and imported. My understanding from other people is that there shouldn't be a `ZLIB_LIBRARY` or a `ZLIB_INCLUDE_DIR` (or a `ZLIB_LIBRARY_DEBUG`, etc.), but rather a symbol named `ZLIB_LIBRARIES` and another symbol named `ZLIB_INCLUDE_DIRS`.

For that reason, I am asking: are you sure that we shouldn't just have the following:
```
  set(LIBS "${ZLIB_LIBRARIES} ${M_LIBRARY}"
```

I can imagine that there are pros and cons to each side, but I would like to know your opinion also.

Moreover, take a look at line 598, further up in this CMake file:
```
  target_link_libraries(png ${ZLIB_LIBRARIES} ${M_LIBRARY})
```

I'd like to be sure that there won't be breakages due to this inconsistency.

CMakeLists.txt
```diff
@@ -931,30 +931,42 @@ endif()
...
 
 # Set up links.
 if(PNG_SHARED)
   set_target_properties(png PROPERTIES
     VERSION ${PNGLIB_SHARED_VERSION}
     SOVERSION ${PNGLIB_SHARED_SOVERSION}
     CLEAN_DIRECT_OUTPUT 1)
+  if(WIN32 AND NOT CYGWIN AND NOT MINGW)
+    foreach (pc_file ${CMAKE_CURRENT_BINARY_DIR}/${PNGLIB_NAME}.pc ${CMAKE_CURRENT_BINARY_DIR}/libpng.pc)
+      FILE(READ ${pc_file} png_pc_orig)
```
ctruta, line 964: Could you please use lowercase CMake commands (`file`, `string`, etc.), for the sake of stylistic consistency?

ctruta (changes requested):
Overall comment: this pull request appears to be carefully developed, and I like it a lot. There are just a few comments to be addressed.

CMakeLists.txt
fanc999: Hi,

Sadly, using `${ZLIB_LIBRARIES}` in the pkg-config file generating won't work, because you will have all the variants inside there.  Will look for a cleaner solution.

fanc999: Hi,

I updated the CMakeLists.txt to reflect on the suggestions that were raised purely on the part that this PR attempts to cover, so I did not replace the occurences of `ZLIB_LIBRARIES` with `ZLIB::ZLIB` in any way but instead grabbed the library names for our pkg-config files with `get_target_property(ZLIB_LIB ZLIB::ZLIB IMPORTED_LOCATION[_xxx])`.

ctruta:
Hi, here's a heads-up,

I am currently applying refactoring changes in order to clean up the CMakefile and to make Clang on Windows work correctly, without regressing other platforms. Your contributions from this pull request may or may not break because of these upcoming changes.

On the positive side of the news, this pull request is next on my to-do list.

PR #397, opened 2021-10-25
Enable win-arm64 platform

gaborkertesz-linaro:
These changes are targeting the enablement Windows on Arm (win-arm64) platform.


ctruta:
Integrated. Many thanks for your contribution!

PR #403, opened 2021-12-04
CMake: Correctly handle generated files

### LINKED ISSUE #344: Can't build with CMake nightly for Xcode
CMake nightly removes support for the old Xcode build system, and the CMake script for libpng now fails as shown below. 

A hacky workaround is to comment out the line that searches for awk.

```py
#find_program(AWK NAMES gawk awk)
```

```
CMake Error in CMakeLists.txt:
  The custom command generating

    /Users/nporcino/dev/usd-dev-install/build/libpng-1.6.37/pnglibconf.out

  is attached to multiple targets:

    genfiles
    gensym
    genvers

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".


CMake Generate step failed.  Build files cannot be regenerated correctly.
```

### LINKED ISSUE #394: libpngconf.c is attached to multiple targets
CMake Error in /Users/smallville7123/StudioProjects/Graphical-Tool-Kit/app/src/main/cpp/freetype-cmake/libpng/CMakeLists.txt:
  The custom command generating

    /Users/smallville7123/StudioProjects/Graphical-Tool-Kit/macOS/IMPORTED_CMAKE_PROJECTS/OpenGLApp_CMAKE_PROJECT/GRAPHICS/libpng/pnglibconf.c

  is attached to multiple targets:

    genfiles
    gensym
    genvers

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".

when generating cmake project for XCode

glebm:
Generated files depend on other generated files, and this previously resulted in the same custom command output being a dependency of multiple other custom commands without a shared custom targets.

Adds a top-level target for each generated file and ensures that commands that depend on generated files also depend on the corresponding custom targets.

Per [CMake documentation](https://cmake.org/cmake/help/latest/command/add_custom_command.html):

> Do not list the output in more than one independent target that may build in parallel or the two instances of the rule may conflict (instead use add_custom_target to drive the command and make the other targets depend on that one).

Fixes #344
Fixes #394

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -340,13 +340,18 @@ else()
   endfunction()
 
   # Copy file
-  function(generate_copy source destination)
-    add_custom_command(OUTPUT "${destination}"
+  # generate_copy(INPUT inputfile OUTPUT outputfile [DEPENDS dep1 [dep2...]])
+  function(generate_copy)
+    set(options)
+    set(oneValueArgs INPUT OUTPUT)
+    set(multiValueArgs DEPENDS)
+    cmake_parse_arguments(_GSO "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+    add_custom_command(OUTPUT "${_GSO_OUTPUT}"
                        COMMAND "${CMAKE_COMMAND}"
-                               -E remove "${destination}"
+                               -E remove "${_GSO_OUTPUT}"
                        COMMAND "${CMAKE_COMMAND}"
-                               -E copy "${source}" "${destination}"
-                       DEPENDS "${source}")
+                               -E copy "${_GSO_INPUT}" "${_GSO_OUTPUT}"
+                       DEPENDS "${source}" ${_GSO_DEPENDS})
   endfunction()
 
   # Generate scripts/pnglibconf.h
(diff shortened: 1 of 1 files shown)


glebm:
@glennrp This merges cleanly against both `master` and `libpng16` branches

cjhowedev:
I have to use @glebm's branch because the Xcode CMake build is broken. I'm not sure how iOS builds are going to work without this, as they require Xcode. I verified that @glebm's branch is working in Xcode 13.3 with CMake 3.22.3.

meshula:
Hi there, support for the "old" xcode build system is going away, and so landing this patch would be really helpful right now.

```

-- Configuring done
-- Generating done
CMake Error in CMakeLists.txt:
  The custom command generating

    /private/var/tmp/usd-pxr3b/build/libpng-1.6.29/pnglibconf.out

  is attached to multiple targets:

    gensym
    genfiles
    genvers

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".
```

ioquatix:
I would welcome your fix in https://github.com/libpng/libpng

ctruta:
Integrated. Thank you very much for your contribution.

PR #412, opened 2022-03-10
fix windows issue when sources got with git but built with wsl fail.

brlcad:
issue is that by default git for windows checks out text files with
CRLF line endings.  this is a problem for awk which is expecting
unix-style LF line endings.  so when an unsuspecting dev clones, but
then attempts to compile on WSL, Mingw, or Cygwin, they'll get an
error from awk.

fix used here is to leverage cmake's ability to configure a file and
perform eol conversions, so we copy the awk scripts from source to
binary. this portable method ensures they have LF endings and build
logic is updated to use the build dir version.

intentionally avoiding .gitattributes to avoid setting precedent.


ctruta:
Integrated. Many thanks for your contribution.

PR #413, opened 2022-04-05
Drop possible tail byte in hIST chunk

eustas:
When I feed png decoder with complete chunks I expect that nothing remains in internal buffer.

Here is the case I've met (via fuzzing): empty palette followed by hIST chunk with 1 byte of payload.
Decoder gives a warning and picks one byte into internal buffer.
For the next chunk it uses this byte -- this is unexpected for the embedder (as chunk start for embedder and png decoder is placed at different positions).

There is no public API to check how much is stored in internal buffer, so it is better to be consistent on processing of "tails" of chunks.

### DIFF
--- a/pngrutil.c
+++ b/pngrutil.c
@@ -2139,7 +2139,7 @@ png_handle_hIST(png_structrp png_ptr, png_inforp info_ptr, png_uint_32 length)
       readbuf[i] = png_get_uint_16(buf);
    }
 
-   if (png_crc_finish(png_ptr, 0) != 0)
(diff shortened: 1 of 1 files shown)


eustas:
Similar to https://github.com/glennrp/libpng/commit/9f2ad4928e47036cf1ac9b8fe45a491f15be2324

eustas:
Friendly ping.

ctruta:
Hello, and thank you for the report. I developed an alternative fix, which I pushed to the master branch. See https://github.com/glennrp/libpng/commit/62c027d4dfe166a81280e50eb8a6e67db63da695

Issue #417, opened 2022-06-28
CMake build breaks with PNG_ARM_NEON checks


superevr:
Per the discussion in #141, shouldn't PNG_ARM_NEON be unset in [CMakeLists.txt](../blob/61a9a054fd616946738a5952267434b6e5822bc7/CMakeLists.txt#L73-L100) (or set to on)?

On Apple Silicon, building via `./configure && make` works fine. But when using `cmake . && make`, I get:

```
libpng/arm/arm_init.c:49:4: error: "PNG_ARM_NEON_FILE undefined: no support for run-time ARM NEON checks"
#  error "PNG_ARM_NEON_FILE undefined: no support for run-time ARM NEON checks"
   ^
libpng/arm/arm_init.c:86:27: error: implicit declaration of function 'png_have_neon' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
               no_neon = !png_have_neon(pp);
                          ^
2 errors generated.
make[2]: *** [CMakeFiles/png.dir/arm/arm_init.c.o] Error 1
make[1]: *** [CMakeFiles/png.dir/all] Error 2
make: *** [all] Error 2
```

To pass the error it, I used `ccmake .` to set `PNG_ARM_NEON on` and rebuilt. But I've been having to do this for numerous other projects that use libpng, and think it may be a bug based on the previous closed pull request.

ioquatix:
I'm having the same problem.

ioquatix:
Setting the following bypasses the issue:

```
-DPNG_ARM_NEON=off
```

However, this should probably work automatically? It is probably an issue with the `CMakeLists.txt` file.

https://github.com/glennrp/libpng/blob/a37d4836519517bdce6cb9d956092321eca3e73b/CMakeLists.txt#L70-L95

It looks like "check" means "do run-time checks". But apparently this isn't supported. So... maybe it would be better to do a compile time check?

ctruta:
Fixed in https://github.com/glennrp/libpng/commit/103f8346e39779a9b8954e434b28bbdab36e90fb

PR #421, opened 2022-07-27
Add compile-time detection of neon instruction set.

ioquatix:
Fixes https://github.com/glennrp/libpng/issues/417 by using compile-time check which according to documentation should be reliable.


ioquatix:
@ctruta What do you think of this? This would make `libpng` easier to compile on Mac OS with M1 chip.

PR #433, opened 2022-09-19
rfc 2083 specification

musvaage:
https://www.rfc-editor.org/rfc/rfc2083.txt

### DIFF
--- a/README
+++ b/README
@@ -144,6 +144,7 @@ Files in this distribution:
       pngwrite.c    =>  High-level write functions
       pngwtran.c    =>  Write data transformations
       pngwutil.c    =>  Write utility functions
+      rfc2083.txt   =>  RFC 2083
       arm           =>  Contains optimized code for the ARM platform
       powerpc       =>  Contains optimized code for the PowerPC platform
       contrib       =>  Contributions
--- a/rfc2083.txt
+++ b/rfc2083.txt
@@ -1,0 +1,5715 @@
+
+
+
+
+
+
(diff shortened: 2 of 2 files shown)


musvaage:
The README should be updated if this PR is considered for a merge.

```diff
        pngwutil.c    =>  Write utility functions
+       rfc2083.txt   =>  RFC 2083
        arm           =>  Contains optimized code for the ARM platform
```

ctruta:
That RFC has long been superseded by the International Standard, which, BTW, is mentioned in the references section of the `png(5)` manual page. Copies of standard documents and drafts may be stored on the PNG home page, or perhaps in a separate repository in the PNG Group's account at https://github.com/pnggroup, but when it comes to storing them in the libpng source tree, they would be just deadweight.

musvaage:
> deadweight

Noted, I've seen elsewhere packagers clipping ChangeLog and NEWS files to respectively 1000 and 1500 lines.

> separate repository 

IIRC, you mentioned "the original PNG specification" in a comment in this PR's linked to Issue.

Possibly a "pasture" repository would suffice to store standard [historical] documents.

jbowler:
> Possibly a "pasture" repository would suffice to store standard [historical] documents.

libpng.org is currently up-to-date and does, indeed, include a list of the preceding specifications.  If you don't want to click through it's here:

http://libpng.org/pub/png/spec/iso/

Greg also provides some helpful notes at the top of the page.

PR #438, opened 2022-10-04
Rename gen{files,sym,prebuilt},check-symbols targets to have png_ prefix

jbms:
This avoids taget namespace pollution when including libpng as a sub-project.


ctruta:
Integrated in master. Thank you very much for your contribution.

Issue #371, opened 2021-03-20
Possible genout.cmake.in error when building with CMake


dxie25:
I am trying to include zlib and libpng as git submodules and build them with CMake. I set PNG_BUILD_ZLIB to ON, meaning that I did not use FindZLIB.

The CMakeLists.txt file includes and uses the genout.cmake script inside a generate_out function.
The scripts/genout.cmake.in file uses ZLIB_INCLUDE_DIR when including zlib when running this script. However, ZLIB_INCLUDE_DIR is never to be seen elsewhere.
Instead, the top-level CMakeLists.txt file uses ZLIB_INCLUDE_DIR**S** for its targets. In addition, CMake's documentation on [FindZLIB](https://cmake.org/cmake/help/latest/module/FindZLIB.html) states that ZLIB_INCLUDE_DIRS will be defined.

For example, I set ZLIB_INCLUDE_DIRS correct but I did not manually define ZLIB_INCLUDE_DIR. When I ran `cmake --build build` on a machine that has AWK, I got a CMake error:
```
Scanning dependencies of target genfiles
 Generating pnglibconf.c
 Generating pnglibconf.out
/my/path/build/extern/libpng/pnglibconf.c:34:11: fatal error: zlib.h: No such file or directory
   34 | # include <zlib.h>
      |           ^~~~~~~~
compilation terminated.
CMake Error at scripts/genout.cmake:78 (message):
  Failed to generate
  /my/path/build/extern/libpng/pnglibconf.out.tf1


make[2]: *** [extern/libpng/CMakeFiles/genfiles.dir/build.make:93: extern/libpng/pnglibconf.out] Error 1
make[1]: *** [CMakeFiles/Makefile2:410: extern/libpng/CMakeFiles/genfiles.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```
I fixed this by manually defining ZLIB_INCLUDE_DIR to the correct directory.

In short, I believe this is a possible error with the intention that genout.cmake.in should use `set(ZLIBINCDIR "@ZLIB_INCLUDE_DIRS@")` instead.

If this was actually intentional and has a better solution, I apologize. I'm still new to CMake.

See scripts/genout.cmake.in:
https://github.com/glennrp/libpng/blob/a37d4836519517bdce6cb9d956092321eca3e73b/scripts/genout.cmake.in#L23
See top-level CMakeLists.txt:
[...]

PR #439, opened 2022-10-05
Fix remaining instance of ZLIB_INCLUDE_DIR

fhanau:
I encountered the same issue as #371 when trying to build libpng with a custom zlib path specified using ZLIB_INCLUDE_DIRS. In #303, this name was changed from ZLIB_INCLUDE_DIR to ZLIB_INCLUDE_DIRS, but one instance in scripts/genout.cmake.in was overlooked as noted by @arsnyder16.

A workaround for #371 is to define both ZLIB_INCLUDE_DIR and ZLIB_INCLUDE_DIRS, but this way it won't be needed.


ctruta:
This is a nice fix, thank you.

PR #443, opened 2022-10-27
Enable intrinsics code for android arm64

halx99:


ctruta (commented):
Thank you for your contribution, but I don't agree with the changes this patch.

When it comes to ARM64, there should be no need for run-time checking, at all, neither on Android nor elsewhere. For this reason, `PNG_ARM_NEON_CHECK_SUPPORTED` is prohibited. Instead, it should be either a compile-time opt-in or a compile-time opt-out, and this is expected to work reliably.

Moreover, looking at the compiler that's shipped with the latest and greatest Android NDK, from `armv7a-linux-androideabi19-clang` to `armv7a-linux-androideabi33-clang`, they all have the `__ARM_NEON` symbol enabled, which means they all produce code for ARM Neon, unconditionally, out of the box. To me, this means that all of the supported Android/ARM systems (both on ARM32 and ARM64) have ARM Neon available, which, in turns, means that the run-time detection of this feature on Android would be essentially an overly complicated way to call a function that always returns true.

In conclusion, I contemplate retiring `contrib/arm-neon/android-ndk.c`, completely.

PR #444, opened 2022-10-30
Document the return values for simplified APIs.

uguu-org:
Document the return values for `png_image_begin_read_*`, `png_image_finish_read`, and `png_image_write_*` functions.


jbowler (approved):

jbowler (changes requested):
I don't think this is necessary and, as the edit shows, it is error prone; programmers will realise pretty quite that all the APIs return true if they succeed and false if they do not (unless there is an error in there, which would be a serious inconsistency; like the error on 3037 :-)

jbowler (commented):
Yeah :-)

jbowler:
Given later versions of C these functions would return bool; that's far better than comments in the header or documentation.  Perhaps libpng 1.8 should do:

#define bool int

when a lower ISO-C standard is used, along with appropriate changes to the return values of course.  At this point the libpng-manual.txt seems fine to me but I don't think any of the suggested comment changes in the code are productive; adding extra lines for something that is certainly going to be obvious when someone runs a program seems pointless.

uguu-org:
It seemed like it would save a few minutes reading time back then, but I agree the extra comments seem unnecessary now.

Also, because the pull request was left stale for a while, I am afraid I would mess something up if I push any changes from here, so I have closed this request.  Thanks for your time.

Issue #372, opened 2021-03-27
link error on Intel macOS when building Universal Binary (arm64) using cmake


seanm:
Trying to build current git master (a37d4836519517bdce6cb9d956092321eca3e73b) as a [Universal Binary](https://en.wikipedia.org/wiki/Universal_binary) (or even just arm64 only) on a Intel Mac fails to link:

```
cd ~
git clone https://github.com/glennrp/libpng.git
mkdir libpng-bin
cd libpng-bin
cmake -DCMAKE_OSX_ARCHITECTURES=arm64 ../libpng
make
```

result:

```
[ 77%] Built target png_static
Undefined symbols for architecture arm64:
  "_png_do_expand_palette_rgb8_neon", referenced from:
      _png_do_expand_palette in pngrtran.c.o
  "_png_do_expand_palette_rgba8_neon", referenced from:
      _png_do_expand_palette in pngrtran.c.o
  "_png_init_filter_functions_neon", referenced from:
      _png_init_filter_functions in pngrutil.c.o
  "_png_riffle_palette_neon", referenced from:
      _png_do_read_transformations in pngrtran.c.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This is because of https://github.com/glennrp/libpng/blob/a37d4836519517bdce6cb9d956092321eca3e73b/CMakeLists.txt#L71 which queries the compiling machine's CPU type.  This doesn't work for Universal Binaries, which are basically cross compilation, because the CPU ***doing*** the building doesn't match the CPU being built ***for***.

A [working hack](https://github.com/InsightSoftwareConsortium/ITK/pull/2450) is to add in `OR APPLE` so that those arm-only .c files are added to the list of files to be compiled. Compiling them on Intel seems harmless because their contents are guarded with [`#if PNG_ARM_NEON_OPT > 0`](https://github.com/glennrp/libpng/blob/a37d4836519517bdce6cb9d956092321eca3e73b/arm/filter_neon_intrinsics.c#L48)

benkasminbullock:
Is this related to these pull requests?

https://github.com/glennrp/libpng/pull/359
https://github.com/glennrp/libpng/pull/354

seanm:
#359 not really related

#354 definitely related.  I'll look into this one...

AswinParanji:
Any updates on this?
I would like to build universal binary for Mac Intel and Silicon
my make command is like this
[...]

PR #451, opened 2022-12-02
Fix Arm Neon optimization detection when building for iOS

TheMostDiligent:
When [cross-compiling for iOS](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-or-watchos), CMake does not define the `CMAKE_OSX_ARCHITECTURES` variable. As a result, [this branch](https://github.com/glennrp/libpng/blob/61bfdb0cb02a6f3a62c929dbc9e832894c0a8df2/CMakeLists.txt#L95-L123) of CMake script does not get executed, so `PNG_ARM_NEON_OPT` is not defined as well as source files for Neon are not added to the project.

The `PNG_ARM_NEON_OPT` then gets defined [here](https://github.com/glennrp/libpng/blob/61bfdb0cb02a6f3a62c929dbc9e832894c0a8df2/pngpriv.h#L132) which results in linker errors as source files are not in the project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,6 +164,8 @@ if(TARGET_ARCH MATCHES "^(ARM|arm|aarch)")
   else()
     add_definitions(-DPNG_ARM_NEON_OPT=0)
   endif()
+else()
+  add_definitions(-DPNG_ARM_NEON_OPT=0)
 endif()
 
 # Set definitions and sources for PowerPC.


TheMostDiligent:
Is it possible that you merge this minor change?
It fixes link errors in XCode

jbowler:
You have an incorrectly set up cmake toolchain file; the toolchain file is what is setting the Apple specific stuff.

TheMostDiligent:
> You have an incorrectly set up cmake toolchain file; the toolchain file is what is setting the Apple specific stuff.

We are not using any toolchain file. We follow [cross-compiling for iOS provided by CMake](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-visionos-or-watchos): `-DCMAKE_SYSTEM_NAME=iOS`. 
We do not define `-DCMAKE_OSX_ARCHITECTURES`, but this is not an error and the build should not fail.

Is there any concern with the proposed change? It makes the CMake build script more reliable in any case.

jbowler:
The change isn't necessary; it should be fixed by this PR #575

[csparker247](https://github.com/csparker247) may care to comment but please test #575 anyway; I don't know if Seth has an iOS test system and I certainly don't.  My understanding is that XCode and Android Studio both supply appropriate toolchains but the old cmake code required rewrites and edits for every different environment.  #575 should fix that.

TheMostDiligent:
@jbowler thanks! I will test https://github.com/pnggroup/libpng/pull/575

jbowler:
@ctruta: duplicate of #372, fixed by #575

jbowler:
@ctruta fixed in libpng-1.8 (#575 has been merged), no comments from @TheMostDiligent  May need #614 but I don't think it does and that is already merged.

I believe that the TARGET stuff is no longer used by the cmake build and that all the source files (i.e. for all architectures) are included in all builds but the original complaint that the optimizations weren't being detected (because CMakeLists.txt was disabling them) has now been fixed in all my builds.  I.e. when including #614 (#614 may have fixed an issue with the NEON palette optimizations.)

I suggest closing this as "fixed" unless there is more input.

TheMostDiligent:
@jbowler Sorry for delay.
I tried to build libpng at 76e5ec217fbdc882bacc40ae3f2276d28507c341 and I get the following link errors:

```
Undefined symbols for architecture arm64:
  "_png_do_expand_palette_rgb8_neon", referenced from:
      _png_do_expand_palette in libpng16d.a[9](pngrtran.o)
  "_png_do_expand_palette_rgba8_neon", referenced from:
      _png_do_expand_palette in libpng16d.a[9](pngrtran.o)
  "_png_init_filter_functions_neon", referenced from:
      _png_init_filter_functions in libpng16d.a[10](pngrutil.o)
  "_png_riffle_palette_neon", referenced from:
      _png_do_read_transformations in libpng16d.a[9](pngrtran.o)
ld: symbol(s) not found for architecture arm64
```

jbowler:
> @jbowler Sorry for delay. I tried to build libpng at [76e5ec2](https://github.com/pnggroup/libpng/commit/76e5ec217fbdc882bacc40ae3f2276d28507c341) and I get the following link errors:

The changes are only in libpng18.  Try HEAD (of libpng18) which is currently https://github.com/pnggroup/libpng/commit/d9f13d8d846e08b00f6530b7a5fe07039d48c78d  The actual changes end with this commit: 

https://github.com/pnggroup/libpng/commit/82f43f71b18d0bcac4f0ce83237491cab562bb95

TheMostDiligent:
How does versioning work? Looks like master=libpng16, what about 17 and 18? Why are they separate?

jbowler:
> How does versioning work? Looks like master=libpng16, what about 17 and 18? Why are they separate?

1.x is not ABI compatible with 1.(!x).   These are effectively _major_ releases because they break ABI compatibility.  The version numbers weren't used consistently - 1.2 -> 1.4 was a massive API change, 1,4 should really have been 2.0 (IMO).  1.x.patch is a patch release; it must be ABI compatible (there is dispute over the meaning of the word _compatible_) and API compatible (but some 1.6 jumps were not...)

1.5 and 1.7 were mainly done by me as "development" releases but that was sort-of-unofficial.  1.7 was never released (it's still in "alpha").  It was a major ABI change but retained almost complete API compatibility.

Each 1.x release maintains build compatibility up to a point.  This mainly depends on the maintainer but the idea at one time was that people who build operating system distributions shouldn't have to mess with the build options in the middle of a release.

Each 1.x release maintains source file compatibility up to a point; source files don't get moved around.  This was probably just a feature of the source control system used over a number of releases; it was basically done by #if's on the version number so things had to stay in the same place in the same files.  This may be moot now since git mostly handles renames ok.

1.8 is currently "alpha"; this is why build changes, ABI changes and at least some API changes are going in there.  The door on those closes when 1.8.0 is released.

TheMostDiligent:
Thanks for the explanation!
I think it is worth adding this to the readme because I'm certainly not the one who has this question.

Branch `libpng18` does work fine, and there were no API issues. So should we switch to 18 or stick to 16 until 18 is released?

jbowler:
> So should we switch to 18 or stick to 16 until 18 is released?

Using unreleased software is something only developers involved, directly or often indirectly, should do.  There's a reason it is unreleased.  I don't use unreleased software other than things I am actively working on.  I do use the latest-and-greatest; the latest  **released** versions of everything that my development operating system distribution (gentoo "~" - the development version) provides.

I think that is probably a "no" :-)

TheMostDiligent:
@jbowler Thanks!

> I suggest closing this as "fixed" unless there is more input.

I confirm it is fixed in libpng18, but since it is a dev branch, I think it makes sense to close this issue once the fix makes it into master.

PR #463, opened 2023-01-09
Fix for universal binaries on macOS

jonny-apple:
Fixes the issue of a `clang: error: no such file or directory: 'x86_64'` error on building universal binaries on macOS.

It is also necessary to specify `-DCMAKE_C_FLAGS="-DPNG_ARM_NEON_OPT=0"`

Example Cmake command:

```
cmake -S libpng -B build_libpng -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" -DCMAKE_C_FLAGS="-DPNG_ARM_NEON_OPT=0"
```


ctruta:
Integrated in trunk. Thank you for the fix.

PR #468, opened 2023-03-02
configure: allow to disable building of tools and tests.

carlo-bramini:
This PR adds two set of options to the configure script:
`--enable-png-tests`/`--disable-png-tests`
and
`--enable-png-tools`/`--disable-png-tools`

By using this feature, an user will be allowed to build only library if needed, which will be useful on platforms not able to build the tools and/or the tests. This PR leaves the existing behaviour as default setting, by building both the tools and the tests if the options are not used.

Actually, CMakeLists.txt already supports this feature with options `PNG_TESTS` and `PNG_EXECUTABLES`.
After this commit, Autotools will provide the same feature.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,13 +12,21 @@ PNGLIB_BASENAME= libpng@PNGLIB_MAJOR@@PNGLIB_MINOR@
 ACLOCAL_AMFLAGS = -I scripts
 
 # test programs - run on make check, make distcheck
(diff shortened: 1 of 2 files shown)


ctruta:
Applied. Later on, I renamed `--enable-png-tests/--disable-png-tests` to just `--png-tests/--png-tests`, etc. Regardless, this was a very useful contribution -- many thanks!

PR #470, opened 2023-03-10
pcal: avoid memory leak when allocation fails

catenacyber:
Sets `PNG_FREE_PCAL` as soon as `pcal_params` is allocated so that even if one of `pcal_params[i]` allocation fails, the array `pcal_params` still gets freed by `png_free_data`

Found by nallocfuzz cf https://github.com/google/oss-fuzz/pull/9902

### DIFF
--- a/pngset.c
+++ b/pngset.c
@@ -401,6 +401,7 @@ png_set_pCAL(png_const_structrp png_ptr, png_inforp info_ptr,
 
    memset(info_ptr->pcal_params, 0, ((unsigned int)nparams + 1) *
        (sizeof (png_charp)));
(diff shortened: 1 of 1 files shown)


ctruta:
> friendly ping @ctruta ?

Friendly pong :-)

Apologies for the long wait. I just returned to the project.

Thank you for the discovery and for the fix. I reworked your fix by moving the setting of the `PNG_FREE_PCAL` flag even earlier than that. I also looked around at the rest of the `PNG_FREE_...` flags inside that module, and I put them in a more consistent order.

catenacyber:
Thanks for the better fix :-)

PR #472, opened 2023-03-19
Fix bashisms

skitt:
makepngs.sh relies on a Bash feature in one of its case statements, ";;&"; this should be made explicit in the shebang.

intgamma.sh declares a function in a manner which may fail in non-Bash sh implementations, this patch uses the correct syntax.

Based on a patch by Roflcopter4:
https://github.com/joncampbell123/dosbox-x/pull/3850

### DIFF
--- a/contrib/testpngs/makepngs.sh
+++ b/contrib/testpngs/makepngs.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 # Make a set of test PNG files, MAKEPNG is the name of the makepng executable
 # built from contrib/libtests/makepng.c
--- a/contrib/tools/intgamma.sh
+++ b/contrib/tools/intgamma.sh
@@ -19,7 +19,7 @@
 
 # Function to print out a list of numbers as integers; the function truncates
 # the integers which must be one-per-line.
-function print(){
+print(){
    awk 'BEGIN{
       str = ""
    }


contrib/testpngs/makepngs.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
ctruta, line 1: Use `#!/usr/bin/env bash` instead of `#!/bin/bash`, to ensure portability across all Unix systems, including *BSD.

ctruta (changes requested):
This commit looks good to me in principle, but I'd like to know what @jbowler says also.
For example, maybe he prefers to rewrite the bashism in `makepngs.sh` using strict POSIX shell language.

jbowler:
The intgamma.sh change is correct; the "function" form is an unfortunate result of confounding the awk syntax with the sh syntax.

makepngs.sh is not used in the build and, anyway, relies on 'makepng' (from libtests) which doesn't build, so there's not much to worry about.  It's always possible to invoke it using "bash makepngs.sh", unfortunately I don't have a "sh" to see how it fails; "bash --posix" doesn't care about the extra syntax.

jbowler:
Hum... @ctruta requests an innocent change and @skitt is AWOL for 6 months.  Permit me to scream.

intgamma is really not part of libpng; it isn't used in the build either so far as I can remember, so all of it can be deleted and that will make everyone's life much better.

Leaving stuff in there which is just compost for bugs seems like a big part of the problem but fixing the whole thing requires a major release (unless you are well tanked) and so should be done.

skitt:
Sorry, I misinterpreted your comment about `makepngs.sh` as implying that the shebang wasn’t important. I’ve pushed an updated version.

ctruta (approved):
Approved. Rebase it, ship it, and thank you!

Issue #482, opened 2023-06-15
tRNS must follow PLTE to be valid


ProgramMax:
The PNG spec  version 1.2 says:
"When present, the tRNS chunk must precede the first IDAT chunk, and must follow the PLTE chunk, if any."

The default libpng config defined PNG_READ_OPT_PLTE_SUPPORTED and causes a tRNS chunk to be considered invalid and ignored if it comes before the PLTE chunk. However, png_get_valid(png, info, PNG_INFO_tRNS) returns a nonzero value. This could cause a user to allocate extra space, resulting in uninitialized memory.

wantehchang:
Chris: Thank you very much for filing this bug report. Coincidentally, I have been investigating this issue since last Friday.

This issue breaks some (but not all) applications that use the following example code in libpng-manual.txt and example.c:

```
    if (png_get_valid(png_ptr, info_ptr,
        PNG_INFO_tRNS)) png_set_tRNS_to_alpha(png_ptr);
```

Applications that assume the decoded output has alpha if this `png_set_tRNS_to_alpha()` call is made will have memory errors.

Here is a possible fix:

```
diff --git a/pngget.c b/pngget.c
index e44933c0d..e64ec2774 100644
--- a/pngget.c
+++ b/pngget.c
@@ -21,7 +21,18 @@ png_get_valid(png_const_structrp png_ptr, png_const_inforp info_ptr,
     png_uint_32 flag)
 {
    if (png_ptr != NULL && info_ptr != NULL)
+   {
+#ifdef PNG_READ_tRNS_SUPPORTED
+      /* png_handle_PLTE() may have canceled a valid tRNS chunk but left the
+       * 'valid' flag for the detection of duplicate chunks. Do not report a
+       * valid tRNS chunk in this case.
+       */
+      if (flag == PNG_INFO_tRNS && png_ptr->num_trans == 0)
+         return(0);
+#endif
+
       return(info_ptr->valid & flag);
+   }
 
    return(0);
 }
```

wantehchang:
I attached two PNG images useful for investigating this issue and testing a fix.

Both images are of color type 2 (PNG_COLOR_TYPE_RGB) and have a tRNS chunk and a PLTE chunk. The two images differ only in whether tRNS is before or after PLTE.

Both images have an opaque blue circle on a transparent green background.

circle-trns-before-plte

[...]

PR #484, opened 2023-06-23
pngrutil.c: Delete dead code in png_handle_PLTE()

wantehchang:
Delete unreachable code in png_handle_PLTE() related to the condition #ifndef PNG_READ_OPT_PLTE_SUPPORTED.

Suppose PNG_READ_OPT_PLTE_SUPPORTED is not defined. If png_ptr->color_type is not PNG_COLOR_TYPE_PALETTE, the function has returned earlier in the function. So if we reach here, png_ptr->color_type must be PNG_COLOR_TYPE_PALETTE, therefore the if-else statement:

```
   if (png_ptr->color_type == PNG_COLOR_TYPE_PALETTE)
   {
      A;
   }
   else if (blah) {
      B;
   }
```

is equivalent to:

```
   A;
```

On the other hand, suppose PNG_READ_OPT_PLTE_SUPPORTED is defined. Then "if (png_ptr->color_type == PNG_COLOR_TYPE_PALETTE)" and the entire "else if (blah) { B; }" are not compiled, leaving only "A;".

### DIFF
--- a/pngrutil.c
+++ b/pngrutil.c
@@ -1014,39 +1014,7 @@ png_handle_PLTE(png_structrp png_ptr, png_inforp info_ptr, png_uint_32 length)
     * have an RGB image, the PLTE can be considered ancillary, so
     * we will act as though it is.
     */
-#ifndef PNG_READ_OPT_PLTE_SUPPORTED
-   if (png_ptr->color_type == PNG_COLOR_TYPE_PALETTE)
-#endif
-   {
-      png_crc_finish(png_ptr, (png_uint_32) (length - (unsigned int)num * 3));
-   }
-
-#ifndef PNG_READ_OPT_PLTE_SUPPORTED
-   else if (png_crc_error(png_ptr) != 0)  /* Only if we have a CRC error */
-   {
-      /* If we don't want to use the data from an ancillary chunk,
-       * we have two options: an error abort, or a warning and we
-       * ignore the data in this chunk (which should be OK, since
-       * it's considered ancillary for a RGB or RGBA image).
-       *
-       * IMPLEMENTATION NOTE: this is only here because png_crc_finish uses the
-       * chunk type to determine whether to check the ancillary or the critical
-       * flags.
-       */
-      if ((png_ptr->flags & PNG_FLAG_CRC_ANCILLARY_USE) == 0)
-      {
-         if ((png_ptr->flags & PNG_FLAG_CRC_ANCILLARY_NOWARN) != 0)
-            return;
-
-         else
-            png_chunk_error(png_ptr, "CRC error");
-      }
-
-      /* Otherwise, we (optionally) emit a warning and use the chunk. */
-      else if ((png_ptr->flags & PNG_FLAG_CRC_ANCILLARY_NOWARN) == 0)
-         png_chunk_warning(png_ptr, "CRC error");
-   }
-#endif
+   png_crc_finish(png_ptr, (png_uint_32) (length - (unsigned int)num * 3));
 
    /* TODO: png_set_PLTE has the side effect of setting png_ptr->palette to its
     * own copy of the palette.  This has the side effect that when png_start_row


pngrutil.c
```diff
@@ -1014,39 +1013,7 @@ png_handle_PLTE(png_structrp png_ptr, png_inforp info_ptr, png_uint_32 length)
...
-         else
-            png_chunk_error(png_ptr, "CRC error");
-      }
-
-      /* Otherwise, we (optionally) emit a warning and use the chunk. */
-      else if ((png_ptr->flags & PNG_FLAG_CRC_ANCILLARY_NOWARN) == 0)
-         png_chunk_warning(png_ptr, "CRC error");
-   }
-#endif
+   png_crc_finish(png_ptr, (png_uint_32) (length - (unsigned int)num * 3));
```
wantehchang, line 1016: (1) Please check if the comment block at lines 1011-1015 should be updated.

(2) I assume there is no bug and therefore the unreachable code can be safely deleted. But perhaps there is a bug and the code should not be unreachable. So please check the code I deleted carefully.

Thanks!

wantehchang:
The early return from `png_handle_PLTE()` that I referred to is the folllowing:

```
 952 #ifndef PNG_READ_OPT_PLTE_SUPPORTED
 953    if (png_ptr->color_type != PNG_COLOR_TYPE_PALETTE)
 954    {
 955       png_crc_finish(png_ptr, length);
 956       return;
 957    }
 958 #endif
```

pngrutil.c
jbowler: They are both "#if**n**def".

Your logic seems to assume the second is "#ifdef"

wantehchang: I confirm they are both `#ifndef`. I did not assume the second is `#ifdef`.

Earlier in the function, at lines 952-958, we have:

```
#ifndef PNG_READ_OPT_PLTE_SUPPORTED
   if (png_ptr->color_type != PNG_COLOR_TYPE_PALETTE)
   {
      return;
   }
#endif
```

Suppose `#ifndef PNG_READ_OPT_PLTE_SUPPORTED` is true. If we reach here,  the `png_ptr->color_type == PNG_COLOR_TYPE_PALETTE` condition at line 1018 of the original code must be true, so the else if block at lines 1025-1048 of the original code will never be taken.

jbowler:
Are, that helps understanding.  Those lines are clearly wrong for the reasons stated in the lines you want to remove; if  READ_OPT_PLTE is **unset** (i.e. "don't read an optional palette") and the PLTE is optional (not a palette PNG) the code as written unconditionally errors out on a CRC error in the PLTE; pretty much the exact opposite of what is wanted.  Easiest fix is just to delete those lines.

The actual code in there dates back to 1996/7 and it looks like the support for disabling reading of PLTE in non-palette PNGs never worked correctly.

@wantehchang since you are using the option to disable the PLTE read do you see any issues if you delete the erroneous lines?  I.e. 952..958

wantehchang:
John: Thank you for taking a look. I am not familiar with libpng. (I remember I spotted this dead code while debugging https://github.com/glennrp/libpng/issues/482.) I am afraid that I don't know the answer to your question.

I work on the Chrome web browser. Chrome undefines PNG_READ_OPT_PLTE_SUPPORTED in its [third_party/libpng/pnglibconf.h](https://source.chromium.org/chromium/chromium/src/+/main:third_party/libpng/pnglibconf.h;l=153?q=PNG_READ_OPT_PLTE_SUPPORTED&ss=chromium) file. I don't know why.

pngrutil.c
wantehchang: I found that the comment block at lines 1012-1016 describes the original code. So the comment block should be deleted (or updated) in this pull request.

jbowler:
萬德 if my analysis is correct it's a bug anywhere.  Since you are reporting bugs in Chrome:

https://chromium-review.googlesource.com/c/chromium/src/+/4616010

I suggest you report the spurious/dangerous exit before the error handling there.  They should be able to test it and verify or refute my suggested change.

wantehchang:
John: I just wrote a test Chrome changelist (patch) of your suggested change:
https://chromium-review.googlesource.com/c/chromium/src/+/5005531

I sent the changelist to Chrome's commit queue dry run. I will report back if passes Chrome's test suite.

Did you suggest I report this bug in Chrome issue tracker? I can certainly do that, but isn't this a libpng bug? Or did you mean Chrome should not build libpng with PNG_READ_OPT_PLTE_SUPPORTED undefined?

jbowler:
> I sent the changelist to Chrome's commit queue dry run. I will report back if it passes Chrome's test suite.

Yes, that's what I wanted.  A full build test of libpng will, I'm fairly sure, just pass simply because it does not unset the option; so it proves nothing.

> Did you suggest I report this bug in Chrome issue tracker? I can certainly do that, but isn't this a libpng bug? Or did you mean Chrome should not build libpng with PNG_READ_OPT_PLTE_SUPPORTED undefined?

It certainly looks like a bug in libpng; READ_OPT_PLTE isn't working correctly so far as I can see.  I doubt it's an issue in Chrome because I believe Chrome turns off the CRC checking; that's a runtime option, "png_set_crc_action".  Unfortunately the tests seem to be failing and they are failing in precisely the area of change.

jbowler:
Oh, it's the other bug you mentioned.  Here's one of the results:

```
[ RUN      ] StaticPNGTests.ColorType2TrnsBeforePlte
../../third_party/blink/renderer/platform/image-decoders/png/png_image_decoder_test.cc:1137: Failure
Value of: frame->HasAlpha()
  Actual: false
Expected: true
```

So, in fact, removing the erroneous code has actually **fixed** the other bug, but your tests are expecting the unfixed result!

Hah.  Definitely an issue for Chrome :-)  Seems like you will need to fix the tests first (it's not a very good test to test for the wrong answer :-)

PR #488, opened 2023-07-30
building png fix and png fix itxt is now optional

TheNicker:
I would  rather to have the default behavior to not to build those, though I'm not sure about backward compatibility
If it can be off by default, please let me know so I could change it.


ctruta:

The building of the extra tools (previously named "executables") could be enabled or disabled via `PNG_EXECUTABLES`, which I have then renamed to `PNG_TOOLS` (but with a compatibility hook to keep the old name working). The configure script has also gotten this option: see `--enable-tools` and `--disable-tools`.

About enabling and disabling each tool, separately, via separate options, I would prefer not to do that. If this set of tools will eventually grow, the number of build options would grow accordingly, complicating the programming interface.

PR #501, opened 2023-11-19
Add iOS headless building support

slingthor:
Signed-off-by: Ingthor Hjalmarsson ihjalmarsson@apple.com

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,23 @@ else()
   set(TARGET_ARCH ${CMAKE_SYSTEM_PROCESSOR})
 endif()
 
+#If we are building for iOS, attempt to avoid code signing
+#This is a useful default for building in complex build chains
+if(IOS)
+  set(DISABLE_CODESIGNED_EXECUTABLES TRUE)
+endif()
+
+macro(add_executable_or_lib_if_codesign)
+  if(DISABLE_CODESIGNED_EXECUTABLES)
+    set(args "${ARGV}")
+    set(arg0 "${ARGV0}")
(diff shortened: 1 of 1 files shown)


ctruta:
Hello, and thank you very much for your contribution.

Although I agree with your idea in principle, I wonder if you could please refactor it in a way that does not complicate the current use of `add_executable`. Would it not be easier (both to implement and to understand) if you could please just surround the option definitions `PNG_TESTS` and `PNG_TOOLS` with:
`if(IOS)` then off by default (possibly with a brief comment as to why they're off by default on iOS) `else()` on by default, `endif()` ...?

I'm open to other simplifying ideas as well.

jbowler:
@ctruta: PNG_TOOLS already has that check.  It would make sense for PNG_TESTS too.  In both cases (i.e. Android and iOS) while running the tests is possible and _should_ happen we need someone to supply something to build an APK or the Apple equivalent to support the tests on those platforms.

My understanding is that the executables don't have to be signed; signing costs $99/year and that's an unreasonable requirement for a FOSS build.  Anyway signing only lasts three months (IRC).  Not signing apparently has some issues and the executables only last for 7 days but that should be fine for the tests.

jbowler:
@ctruta: I'm now a member of pnggroup but I can't add labels.  I admit the github user interface is not designed for github users like me; I think you have to be a total githead!  However, that said, I can't edit titles or, indeed, do anything.  @ProgramMax might care to comment; I don't think the list of labels (if I am understanding what that is) even includes "W3C PNG v3" or anything similar!

PR #512, opened 2023-12-28
Add runtime zlib version number check for Apples

jbowler:
On Apple systems libpng may be compiled against a newer version of zlib than the library eventually gets used on.  This may or may not work but in the specific case of Issue 187 it apparently results in a crash (a test case has not yet been established).

This change modifies pngrutil.c to verify that the compiled version number of zlib matches the installed version number and applies that verification to the problem call.

This is minimalist but safe; as such it is, in fact, a good check for all calls (on all operating systems) to zlib which have version number checks. The Linux Foundation man page for 'zlibVersion' (at least in what I found) recommends, "Applications should compare the value returned from zlibVersion() with the macro constant ZLIB_VERSION for compatibility."

I don't have the faintest idea how up-to-date that is.

I recommend that this patch be reviewed by the various parties contributing to Issue 187 and that at least one of them confirm that an application which crashes (at the relevant point) without the patch no longer crashes with it.

TESTED: gentoo-dev, gcc (Gentoo 13.2.1_p20231216 p11) 13.2.1 20231216 with 'cmake, make all, make test' and 'configure, makedistcheck'.

### DIFF
--- a/pngrutil.c
+++ b/pngrutil.c
@@ -423,7 +423,8 @@ png_inflate_claim(png_structrp png_ptr, png_uint_32 owner)
 
 #if ZLIB_VERNUM >= 0x1290 && \
    defined(PNG_SET_OPTION_SUPPORTED) && defined(PNG_IGNORE_ADLER32)
-      if (((png_ptr->options >> PNG_IGNORE_ADLER32) & 3) == PNG_OPTION_ON)
+      if (((png_ptr->options >> PNG_IGNORE_ADLER32) & 3) == PNG_OPTION_ON &&
+          inflateValidate != NULL)
          /* Turn off validation of the ADLER32 checksum in IDAT chunks */
          ret = inflateValidate(&png_ptr->zstream, 0);
 #endif


jbowler:
I think I said everything in the commit msg but it appears that there are no crash repros available for the original Issue.  Nevertheless I think this is a good commit; it just requires testing by Apple developers or, even better, some CI tests for Apple from those who use it!

vadz:
FWIW this will still leave the original warning, but now it could be just disabled, I guess.

pngrutil.c
```diff
@@ -332,6 +332,36 @@ png_read_buffer(png_structrp png_ptr, png_alloc_size_t new_size, int warn)
...
+ */
+static int png_zlib_version_ok(void) {
+   const char *zv = zlibVersion();
+   if (zc != NULL) {
+      const char *pv = ZLIB_VERSION;
+
+      while (*zv == *pv) {
+         if (*pv == 0) return 1;
+         ++zv, ++pv;
+      }
```
vojtechkral, line 356: I believe this is way too coarse - this will error out in all cases where the version isn't exactly the same, that is, even when the runtime zlib version is _newer_ in a compatible way (eg. patch number).

I think the fix should be in the libpng's build system, on Mac OS, it should decide based on deployment target whether `inflateValidate()` is available or not.

vojtechkral: In fact, I think if you tried running an app build against a newer Mac OS SDK on an older system, it would fail to start up due to linker not being able to find the `inflateValidate()` symbol, no?

So you wouldn't ever get to this check.

vojtechkral: > I think the fix should be in the libpng's build system; on Mac OS, it should decide based on deployment target

To clarify: This check:

```c
#if ZLIB_VERNUM >= 0x1281
```

would be replaced by something like 

```c
#if ZLIB_VERNUM_EFFECTIVE >= 0x1281
```

(name bikeshedable), where `ZLIB_VERNUM_EFFECTIVE` would be set to `ZLIB_VERNUM` iff the deployment target is not set/is default. If the deployment target is older than the SDK being built against, the `ZLIB_VERNUM_EFFECTIVE` would be set to an older one based on deployment target (you'd probably need some kind of table mapping SDK versions to zlib vernums).

jbowler: Replying to @vojtechkral:

>I think the fix should be in the libpng's build system; on Mac OS, it should decide based on deployment target whether inflateValidate() is available or not.

That fix was suggested in issues refererenced by the original (libpng) issue as, apparently, minimum and maximum targets are available at build time.  It would seem from what I understand that an Apple build defines the minimum (Apple OS) target available and the maximum **available** while allowing the app build to specific a minimum **required**.  Some relevant links:

https://github.com/wxWidgets/libpng/pull/1
https://github.com/wxWidgets/libpng/pull/5/commits/dfc331f0bd8165dd41c0c5b0a666d490e5dc8a54 (still in the runtime)

A fix based on this is far preferable in my opinion because it produces a consistent behavior in the _application_; either Adler32 checking is turned off on all supported OSes or it is left on in all supported OSes.

_Anyway I'll send a better patch that has only a couple of lines in it._

> In fact, I think if you tried running an app build against a newer Mac OS SDK on an older system, it would fail to start up due to linker not being able to find the `inflateValidate()` symbol, no?

Then it wouldn't crash, it would output a DLL load error.  This is exactly what happens on other (non Apple) systems when a libpng caller is matched against an older zlib, here's an example from a Windows app:

`https://github.com/Ryujinx/Ryujinx/issues/2931`

So Apple is handling DLLs with a function level delayed link; this is also an option on other hardware architectures but it has been a common approach on ARM perhaps because of the NS32016 background of the original (Acorn) devs (PANOS).  It is my favored approach too; it reduces startup time with large apps at the cost of overhead on the first call (practically zero on the NS32016 IRC :-)

Anyway what seems to be happening is that `png_set_option `is called to turn the Adler32 check _off_ then the app calls libpng to read a PNG, at which point the `inflateValidate `call happens and the Apple implementation of the delayed link causes the app to "crash".

There's no way for the app to discover that `png_set_option `has/will fail; that is part of the design of `png_set_option `because it is used for things that are optional like hardware acceleration.  The whole thing seems to be some kind of end-work-round for some undocumented bug/problem: see a106899a

vojtechkral: @jbowler Those patches are _compile-time_-evaluated  macros. They're not runtime.

> So Apple is handling DLLs with a function level delayed link
No, this happens upon executable load time, _before_ application gets to run any code. There's no lazy loading.

> There's no way for the app to discover that `png_set_option` has/will fail
The application doesn't _need_ to do any such discovery.
Either the application is built against zlib with this feature, or it is not. If someone needs their app to run on Mac OS before 10.13, they need to not link against that function, it's just a matter of propagating that information through libpng build system correctly.

jbowler:
This is actually based on:

https://github.com/wxWidgets/libpng/pull/5/commits/dfc331f0bd8165dd41c0c5b0a666d490e5dc8a54

however the code is now fail-safe; the above fix (either versions, the one in that commit or the one it replaced) would only disable inflateValidate if it could be proved that it was not available (double negative; inverting the boolean arithmetic in the original fix on line -427 helps understanding.)  This fix only enables it if the minimum OS libpng will run on is known to support inflateValidate.

BTW this is only an issue when an app or library ships a private DLL build using a system zlib build.  It would make more sense for the app to use a static build of zlib and libpng appropriately deconfigured.

jbowler:
@ctruta: on balance my latest "push" seems like the best fix but there are four approaches:

1) Do nothing.  This only happens with a _private_ build of libpng and it's trivial to statically link the private build against zlib, avoiding the problem entirely.  The Apple build systems even output a warning message about this.  In addition simply deconfiguring **png_set_option** in pngusr.dfa (remember this is a private build) avoids the issue; how many of these private builds require **png_set_option**?

2) Remove the **inflateValidate** call entirely; removes code, simple, mark the Adler32 OPTION deprecated.  Originally the code was meant to disable Adler32 along with CRC32, following the CRC32 flag, then Glenn removed it from everything except IDAT then he made it a png_set_option so, at that point, no one uses it so far as I have been able to discover.  (The inflateReset2 attempt is clearly broken BTW; it fails to skip the Adler32 at the end.)  So no one uses it, png_set_option makes no guarantees anyway (and the result can't be checked) so remove the code.  It's a quiet API change (libpng 1.8 maybe.)

3) Use the latest patch: it fails safe, it's compile time and it doesn't change the API.

4) Add it as a 'setting' in scripts/pnglibconf.dfa and default it to off.  Remember these are private builds of libpng so should be [de]configured anyway if they are shipped, hardly anyone (no one?) uses png_set_option at all and an app that uses a vendor version of zlib can't relay on png_set_option doing anything in this scenario.  This is too complex IMO without an actual example of a real app which calls png_set_option and crashes.  A real app using a system libpng will not have a problem: solution (1) or (3).

vojtechkral:
> This is actually based on:
> 
> [wxWidgets@dfc331f](https://github.com/wxWidgets/libpng/commit/dfc331f0bd8165dd41c0c5b0a666d490e5dc8a54)

Right, but I don't think `MAC_OS_VERSION_MIN_REQUIRED` is defined anywhere.
Not sure what the right macro is, the linked Wx patch uses `MAC_OS_VERSION_MIN_REQUIRED`, but that seems old? Not sure that's still a thing these days.
It might be `__MAC_OS_X_VERSION_MIN_REQUIRED` instead these days, but I'm not sure. Some digging / testing for that might be needed. I'm not at a Mac box right now.

jbowler:
> Right, but I don't think `MAC_OS_VERSION_MIN_REQUIRED` is defined anywhere.

It's discussed in the other issues; Apple changed it at some point.  This is a big part of the issue; the fix needs to be failsafe because all these #define tests are apparently subject to change with new build systems.  I'm assuming that __APPLE__ is a sufficiently reliable test to detect that the issue exists, but it means that anything using Apple macros requires a maintenance commitment.

pngrutil.c
jbowler: > @jbowler Those patches are _compile-time_-evaluated macros. They're not runtime.

Sorry that wasn't clear; I meant that the fix is in the runtime source code not the build mechanism code; not by setting something when building libpng (including modifying pngusr.dfa).  (Same usage as in the pull request; change the libpng source rather than a change to the build system.)

>If someone needs their app to run on Mac OS before 10.13, they need to not link against that function,

Then why does @vadz's merged patch from https://github.com/wxWidgets/libpng/pull/5 call an Apple specific OS version check function to prevent a call to inflateValidate()?  I.e. in the original issue https://github.com/glennrp/libpng/issues/187 (which also mentions the "build control" patch below):

https://github.com/vadz/wxWidgets/commit/28348b578a05d50ef3a9a0a62e4c18c774e9aeab

I'm talking about these exact lines from the commit:

```
#  if PNG_USE_ZLIB_INFLATE_VALIDATE == 2
         if (__builtin_available(macOS 10.13, iOS 11.0, *))
#  endif
            /* Turn off validation of the ADLER32 checksum in IDAT chunks */
            ret = inflateValidate(&png_ptr->zstream, 0);
```

So that is preventing the call at runtime (correct use here; when libpng is running) if the OS version isn't available and furthermore @sgowdev asserts the same change works here (well, this only has the MACOS part):

https://github.com/glennrp/libpng/issues/187#issuecomment-342633982

>it's just a matter of propagating that information through libpng build system correctly.

This patch does that:

https://github.com/glennrp/libpng/pull/261/commits/d020132c02b796c0259b8963076fcef9e60bdf84

but the pull request, https://github.com/glennrp/libpng/pull/261, was closed without comment by the requestor, https://github.com/mrserb (and apparently not merged).  The change is incomplete; scripts/pnglibconf.dfa needs support for the new setting, but it gave build control, i.e. "configuration control" (as opposed to, I guess, 'source level' control  - what I called 'runtime' control).  It's also not specific to Apple and a generally useful configuration option (since very few private builds will need the functionality.)

You, @vadz and @sgowdev need to sort out what works and what doesn't!  If what you say is correct the only option is to remove the `inflateValidate `call in a private build which targets an older OS and my patch (solution [1], current one in this pull request) does that without needing @mrserb's configuration option or any build time changes just so long as someone can work out the correct #defines to use.

In fact a configuration change is easier to maintain but slightly less safe unless the option is off by default.  It just requires the new setting to be turned on/off in pngusr.dfa (pnglibconf.h.prebuilt is a machine generated file that is distributed solely to support old 'makefile' builds, DFA_XTRA can be used to but I haven't checked that with cmake.)

vojtechkral: Ah ok, so basically the linker [sets the symbol to `NULL`](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html) if it's not present in the system's library _and_ the import is marked weak. That's cool.
So why not just lift that patch? Looks like a good solution as far as I can tell, though I can't vouch for the version macros being used.

jbowler: > Ah ok, so basically the linker [sets the symbol to `NULL`]

That's a non-OS-specific patch.  I'll revise the pull request.

jbowler:
Ok, the "fix" is now down to 1 line and is clearly safe (since a pointer to a function can be evaluated validly before a function call in ANSI-C - this is absolutely ANSI-C conformant!)  This replaces (3) above.

vadz:
Sorry, I don't think this works at all. If you always link with this function, the program won't start under the OS version without it.

The whole point of `MAC_OS_VERSION_MIN_REQUIRED` check was to avoid linking with it if we need to support old OS.

jbowler:
@vojtechkral:
>Ah ok, so basically the linker [sets the symbol to NULL](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html) if it's not present in the system's library and the import is marked weak.

@vadz:
>If you always link with this function, the program won't start under the OS version without it.

Either it's weak or it isn't: please clarify.  I don't have access the Apple system headers which quite possibly are hacked versions of Mark's.

Anyway, @vadz, you still haven't clarified what on earth you expect the runtime check to do.  The suggested wxWidget fix didn't include it but it wasn't merged and there didn't seem to be any relevant comments as to why.

jbowler:
https://developer.apple.com/forums/thread/723983

So there is an Apple supported build of libpng

So this is only about people who, for some reason, are building libpng and building it against the Apple supported build of zlib.

Don't do that: if you build libpng build zlib too.  If you do it right the whole shebang will be a lot smaller than a complete build of libpng.

vadz:
I'm honestly not sure what we're discussing any more nor what the goal is, so let me just explain what I know and if you have any questions, please ask them.

1. If you link with zlib in the usual way, you link with it normally, there is no run-time magic involved and if you reference a symbol not present in the library, the program doesn't start.
2. You can use `__builtin_available` (a.k.a. `@available`) Apple-specific construct to restrict some part of code to only the OS version specified. This would allow, in principle, only call `inflateValidate` on the macOS version supporting it in the system library.
3. But [wxWidgets patch](https://github.com/glennrp/libpng/commit/3ffeff7877598d3236cc09a6d3f478073eb33f35) doesn't do this, but does something much simpler and simply disables this call at _compile_ time when targetting version of macOS earlier than 10.13.

Maybe I should also explain how `MAC_OS_X_VERSION_MIN_REQUIRED` works. When you compile on macOS you may use `-mmacosx-version-min=x.y` compiler/linker option which says that you want to support code targetting all OS versions starting from `x.y`. In this case, linker will give a warning if the final binary uses any symbols not present in the libraries in this OS version.

Also, final note: libpng is not included in any macOS version, but libz is, so it makes perfect sense to compile yourself the former but not the latter (both are available from 3rd party repositories such as Homebrew, but this absolutely doesn't mean that this is somehow supported by Apple).

Again, I hope this helps but I'm really lost as to what you're trying to do. AFAICS there is a simple, trivial fix for the problem which you clearly don't want to apply for the reasons you've never explained. Luckily, macOS < 10.13 almost doesn't exist anyhow any longer (the web browser stats I could find indicate that they account for ~3% of Mac users), so this is not very important, but I still can't help wondering what's wrong with just applying the trivial patch and not spending any more time on this.

jbowler:
> there is a simple, trivial fix for the problem which you clearly don't want to apply for the reasons you've never explained.

Eh?  You didn't read the previous version of this pull request.

There are umpteen variations of this check on wxWidgets, but no one seems to have submitted a "pull" request for a single one.  So far as I could deduce from #187 the runtime check was the "best solution".  I don't like that _because_ it includes the runtime check, which is apparently completely unnecessary, and it's approaching 60 lines long.

>But [wxWidgets patch](https://github.com/glennrp/libpng/commit/3ffeff7877598d3236cc09a6d3f478073eb33f35) doesn't do this, but does something much simpler and simply disables this call at compile time when targetting version of macOS earlier than 10.13.

Exactly.  That was the last but-one version, indeed I even referenced that commit (indirectly; I couldn't find the actual commit but I included exactly those lines with the pull request!)

But then @vojtechkral observes:

>Right, but I don't think MAC_OS_VERSION_MIN_REQUIRED is defined anywhere.

As I summarized in my message to @ctruta (who is the maintainer and the only person who can merge a pull request) before @vojtechkral made that comment it's one of four possible solutions, number (3) to be precise.  I'm happy to resurrect that version but you guys need to make up your minds.

**The solution does need to be failsafe**.  I said that and explained it when I explained why I implemented (3) the way I did.

vojtechkral:
I was mostly commenting on individual bits as I know/remember them, just "my two cents" type of comments, because I was in the original issue.

The `MAC_OS_VERSION_MIN_REQUIRED` was I think a typo when copying from the other patch? Note the distinction `MAC_OS` vs `MAC_OS_X` ... as for the the one with the `X` in it, I don't know whether that one is defined or not. I only know about the `-mmacosx-version-min` flag that Vadim also mentioned, but what macros that turns on I'm not exactly sure and documentation is scarce.

As for the runtime check, I was assuming the zlib that's part of Mac OS SDK would have its functions annotated with the right compat attributes to make the check work, but I don't have a Mac OS box available at the moment so can't make sure that's really the case.

> but you guys need to make up your minds

I don't think I can help much further, probably what you need at this point is someone with Mac OS & dev setup to test.

pngrutil.c
```diff
@@ -423,7 +423,8 @@ png_inflate_claim(png_structrp png_ptr, png_uint_32 owner)
 
 #if ZLIB_VERNUM >= 0x1290 && \
    defined(PNG_SET_OPTION_SUPPORTED) && defined(PNG_IGNORE_ADLER32)
-      if (((png_ptr->options >> PNG_IGNORE_ADLER32) & 3) == PNG_OPTION_ON)
+      if (((png_ptr->options >> PNG_IGNORE_ADLER32) & 3) == PNG_OPTION_ON &&
+          inflateValidate != NULL)
```
ctruta, line 427: Suggestion: considering how weird this workaround is (in my opinion, at least), I'd like to see the comment from the commit message added to the actual code. We shouldn't lose track on what we did and why.

I'm thinking about something like this:
```
   if (((png_ptr->options >> PNG_IGNORE_ADLER32) & 3) == PNG_OPTION_ON)
   {
      /* On some Apple systems, zlib.h claims that inflateValidate is available
       * when, in fact, it isn't. On those systems, inflateValidate is NULL.
       */
      if (inflateValidate != NULL)
      {
         /* Turn off validation of the ADLER32 checksum in IDAT chunks */
         ret = inflateValidate(&png_ptr->zstream, 0);
      }
   }
```

vadz:
> > there is a simple, trivial fix for the problem which you clearly don't want to apply for the reasons you've never explained.
> 
> Eh? You didn't read the previous version of this pull request.

Maybe I missed something, sorry. As I said, I'm quite lost here.
 
> There are umpteen variations of this check on wxWidgets, but no one seems to have submitted a "pull" request for a single one.

But I *had* proposed submitting one and this was completely ignored without any explanation.

> But then @vojtechkral observes:
> 
> > Right, but I don't think MAC_OS_VERSION_MIN_REQUIRED is defined anywhere.

It should be `__MAC_OS_X_VERSION_MIN_REQUIRED` and it is defined by Apple headers when `-mmacosx-version-min` is used.
 
> **The solution does need to be failsafe**.

I am not sure what do you mean by this. How can testing for `__MAC_OS_X_VERSION_MIN_REQUIRED` "fail" in any way that would be worse than the current situation?

pngrutil.c
vadz: Sorry, no amount of comments will help with this code, and this comment is just wrong. 

Again, sorry if this went unnoticed above, but if you link with this function and use it like this (not guarded by any `__builtin_available` checks), the program won't even start, so this check is completely useless.

ctruta:
Stupid side question:
`inflateValidate` is marked as undocumented inside zlib.h, so it may be changed or even removed at any point, unless @madler says otherwise. Is this whole code that we're fixing here actually necessary?

pngrutil.c
ctruta: @vadz got it.

jbowler:
Dang!  I renamed libpng16 to jbowler/libpng:PR512.  All the variations on this PR are here:

https://github.com/jbowler/libpng/branches

The #define one is PR512-2

jbowler:
> Is this whole code that we're fixing here actually necessary?

Not a stupid question at all.  No; the code is not necessary at all, it's something that _might_ be available and can be turned on by png_set_option but there is no way (at present) of finding out if it _is_ available and turning it on fails silently if it is not available.

I can't remember exactly when this mess started but I think I was still working on libpng at the time.  All the same these may be false memories; reading commit messages and using git blame does that to me sometimes.  Bearing that in mind:

libpng has had, for ever, a flag to turn off "CRC" checking.  This is widely used.  The flag, however, did not turn off the Adler32 checking because it could not.  I remember (false memory?) spending a lot of time inside the code that runs the zlib inflate loop looking for a way to skip the Adler32 after zlib code had delivered the last byte.  IRC it was impossible if zlib::avail_in was more than 0 when the last deflate byte was delivered.  Not only did zlib say that the stream wasn't finished but also it proceeded to decode following bytes even though zlib::avail_out was zero.  If there were four more bytes it returned a stream error.

Calling inflateReset with a negative window bits was the first attempt.  For some reason that didn't work, I was just guessing when I said that it maybe didn't read the Adler32 causing the infamous warning about excess data in the stream.

At some point inflateValidate was added, maybe Glenn asked Mark directly; that's quite likely.

For libpng1.6 dropping the call entirely is only a minor API change and, well, pngtest should still pass :-)  (pngtest contains the only call I know of to turn the checking off).  This is my "option 2".

The most likely client that does set the option is Chrome, but they have their very own patched libpng (the APNG version).   They can just revert the patch which deletes the call.

jbowler:
> As for the runtime check, I was assuming the zlib that's part of Mac OS SDK would have its functions annotated with the right compat attributes to make the check work, but I don't have a Mac OS box available at the moment so can't make sure that's really the case.

That seems reasonable; if it were me doing it I would mark inflateValidate and all the 2() APIs weak (in the header) and it does explain why the original issues (some of them) use the word "crash".

One of the nice things about the check is that it works with a static zlib build too.

jbowler:
@ctruta: I can recreate this if you wish, all three of the possibilities are in my repo and I can add "delete it all" very easily :-)

My mistake: I repeatedly forget to create a new branch for github pull requests in my own repo.

pngrutil.c
vojtechkral: > Again, sorry if this went unnoticed above, but if you link with this function and use it like this (not guarded by any `__builtin_available` checks), the program won't even start, so this check is completely useless.

Just for my understanding of Mac OS, according to the [docs](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html), the check should work _iff_ the `inflateValidate()` symbol is imported as weak, no?
ie.
```c
extern int inflateValidate(z_streamp, int)  __attribute__((weak_import));
```
_provided_ that the symbol is also marked for weak linking in the SDK.

vadz: I guess this could work if `inflateValidate()` were declared as weak, but it isn't.

Consider this example (called `testz.c`):

```c
#include <zlib.h>

#include <stdio.h>

int main() {
    printf("inflateValidate=%p\n", inflateValidate);
}
```

then

```console
% cc -Wall -mmacosx-version-min=10.10 testz.c
testz.c:6:36: warning: 'inflateValidate' is only available on macOS 10.13 or newer [-Wunguarded-availability-new]
    printf("inflateValidate=%p\n", inflateValidate);
                                   ^~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/zlib.h:1917:32: note: 'inflateValidate' has been marked as being introduced in macOS 10.13 here, but the deployment target is macOS 10.10.0
ZEXTERN int            ZEXPORT inflateValidate OF((z_streamp, int))
                               ^
testz.c:6:36: note: enclose 'inflateValidate' in a __builtin_available check to silence this warning
    printf("inflateValidate=%p\n", inflateValidate);
                                   ^~~~~~~~~~~~~~~
1 warning generated.
```

Looking at the header mentioned in the error message shows that the full declaration is

```c
ZEXTERN int            ZEXPORT inflateValidate OF((z_streamp, int)) __API_AVAILABLE(macos(10.13),ios(11.0));
```

and there is no weak import anywhere.

BTW, to fix both the warning and the actual behaviour you could use the availability test:

```c
#include <zlib.h>

#include <stdio.h>

int main() {
    if (__builtin_available(macOS 10.13, *))
        printf("inflateValidate=%p\n", inflateValidate);
    else
        printf("There is no inflateValidate in this macOS version!\n");
}
```

Again, this is not what I had originally proposed because I didn't want to avoid the complications, but as it's too late for this by now, I can once again reiterate my readiness to produce a patch implementing either compile-time or run-time check, if the maintainer(s) would be interested in it.

vojtechkral: > and there is no weak import anywhere.

Well, chances are the `__API_AVAILABLE` thing, which seems to expand to `__attribute__((availability(...)))`, implies weak linking under the hood.
How does your fix work? You're still using `inflateValidate` in the source code, so either it's weak linking too, or does it rely on `if (__builtin_available(macOS 10.13, *))` being a const expression and the call to `inflateValidate` being optimized out?

vadz: Yes, you're right, quoting https://clang.llvm.org/docs/AttributeReference.html#availability

<blockquote>
A declaration can typically be used even when deploying back to a platform version prior to when the declaration was introduced. When this happens, the declaration is weakly linked, as if the weak_import attribute were added to the declaration. A weakly-linked declaration may or may not be present a run-time, and a program can determine whether the declaration is present by checking whether the address of that declaration is non-NULL.
</blockquote>

I'm really sorry, I hadn't realized this but it looks like just testing for the function pointer does work, after all (and the failures to start I remember may have been due to using headers without the proper availability attribute). It would still emit warnings during compilation, however, so using the proper test as above would be still better.

To summarize, my order of preferences would be:

1. Just a compile-time test: it's the simplest and I don't see any real drawbacks to not compiling this code.
2. A proper run-time test, using `__builtin_available`.
3. A simple run-time test for the function pointer not being null with Apple-specific pragmas suppressing the warning about using it.

As before, I can produce a PR implementing either of those if you'd like.

jbowler:
I'm just waiting on @ctruta's response.

He seemed to be leaning towards the "remove it everywhere" approach, that wasn't quite in my list but it's easy to do, safe, solves the problem everywhere with complete API compatibility etc.

I may also try something that's closer to @sgowdev's very original proposal and a couple of the other variations, adding a new SUPPORTED option and setting it to *off* by default; so the API doesn't change but pnglibconf.h contains something like  PNG_DISABLE_ADLER32_CHECK_SUPPORTED (by default undefined).  This is also easy and it might smoke out any people who really are using the functionality (in addition to any of the big guys, who will most likely deal with any change without telling us.)

The PR512 variants are here:

https://github.com/jbowler/libpng/branches

512-2 still has the wrong macro name.

pngrutil.c
vojtechkral: @vadz Thanks, and yeah that's a good assesment. Good point about still having to use Apple-specific macros to supress the warning.

PR #513, opened 2023-12-30
pngfix: ensure fclose happens last on file close

jbowler:
This reverts the code to a variant of my old style of doing an 'fclose' correctly and adds comments explaining why it is so difficult.  Thanks to @ihsinme for pointing out the error on github.com

### DIFF
--- a/contrib/tools/pngfix.c
+++ b/contrib/tools/pngfix.c
@@ -1005,10 +1005,18 @@ file_end(struct file *file)
 
    if (file->out != NULL)
    {
-      /* NOTE: this is bitwise |, all the following functions must execute and
-       * must succeed.
+      /* On some systems 'fclose' deletes the FILE struct (making it
+       * inaccessbile).  There is no guarantee that fclose returns an error
+       * code from fflush or, indeed, from the FILE error indicator.  There is
+       * also no explicit (or clear) guarantee in the standard that anything
+       * other than a read or write operation sets the error indicator; fflush
+       * is not a read or write operation, so both conditions must be checked
+       * to ensure the close succeeded and in ANSI-C conformant code they must
+       * be checked before the fclose call.
        */
-      if (ferror(file->out) | fflush(file->out) | fclose(file->out))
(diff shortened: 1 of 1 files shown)


ValeZAA:
>On some systems 'fclose' deletes the FILE struct (making it inaccessbile).

We are talking Windows file handles system here?

jbowler:
> > On some systems 'fclose' deletes the FILE struct (making it inaccessbile).
> 
> We are talking Windows file handles system here?

No.  Those are at the OS level, i.e. similar to POSIX 1003.1 file system descriptors (which are "small" integers which index into the kernel u-area in kernel space in the original UNIX implementations).  So far as I know NT follows the traditional 2-level model and implements FILE as an in-process buffer with a separate object with almost identical functionality living in the OS memory (not necessarily inside the micro-kernel itself.)

I was stating my understanding of existing implementations and of possible C89 conformant implementations.  The actual definition of fclose does say the same thing:

> [24] **4.9.5.1 The fclose Funtion**
> [32] [* * *] If the associated buffer was automatically allocated, it is deallocated.

The other general definition is in 4.9.3 line 26 and 27, "The value of a pointer to a **FILE** object is indeterminate after the associated file is closed[.]"  A more than slightly confusing sentence but my understanding is that the noun is "object", not "pointer" :-)  So even if the FILE is not on the heap the contents are "indeterminate" after fclose.  For example on a multi-threaded application the FILE might represent the result of an intervening fopen.  (Of course pngfix is not MT.)

ValeZAA:
>The value of a pointer to a FILE object is indeterminate after the associated file is closed.

No, the noun is value of a pointer, i.e. memory address. 

It is just it's undefined behavior, of course just like with use-after-free, the memory address will still be there and can be casted to void *0 (NULL macros).

J.2 Undefined behavior:

— The value of a pointer to a FILE object is used after the associated file is closed (7.21.3).

ctruta:
Thanks!

jbowler:
@ctruta: the merge addresses the issue raised in https://github.com/glennrp/libpng/pull/378

PR #514, opened 2023-12-30
Do not build unused Adler32 code

jbowler:
This removes the default build of an undocumented feature to disable Adler32 checksums on those systems where it was the default.

The PR is motived by github #187 however it fixes a much more general problem (#187 is limited to an issue where libpng "crashes" on some manufacturer systems).  The fix is based on a suggestion by @sgowdev who is the originator of the issue.

When libpng disables the checking of Adler32 checksums it does so by an undocumented and therefore possibly unsupported call to a zlib function which does not exist in some versions of zlib.

Fortunately libpng only does this if the caller of libpng explicitly asks for it to happen.  Unfortunately the call to the undocumented function is still in the compiled and built libpng and this means that on some systems (as identified in #187) libpng can fail to load or maybe even crash.

The libpng authors are currently unaware of any program or system that uses this feature and none has been identified by the contributors to #187.

In this fix an option is added to *enable* the code so that by default the code is *disabled* - this is a simple generalization of the suggestion by @sgowdev.

BENEFITS: the problem is eliminated, users of the functionality, if any, are idenfified, the functionality can be implemented correctly in the future or it can be removed.  Hardly anyone complains.

COSTS: someone will complain that they have to enable an option in a libpng build to use a feature that never worked consistently in the first place.

This patch has been tested both with the option enabled and with it disabled via pngusr.dfa.  Tests, checks pass with cmake and configure, make distcheck passes on configure.

### DIFF
--- a/png.h
+++ b/png.h
@@ -3204,9 +3204,12 @@ PNG_EXPORT(245, int, png_image_write_to_memory, (png_imagep image, void *memory,
 #ifdef PNG_MIPS_MSA_API_SUPPORTED
 #  define PNG_MIPS_MSA   6 /* HARDWARE: MIPS Msa SIMD instructions supported */
 #endif
-#define PNG_IGNORE_ADLER32 8
+#ifdef PNG_DISABLE_ADLER32_CHECK_SUPPORTED
+#  define PNG_IGNORE_ADLER32 8 /* SOFTWARE: disable Adler32 check on IDAT */
+#endif
 #ifdef PNG_POWERPC_VSX_API_SUPPORTED
-#  define PNG_POWERPC_VSX   10 /* HARDWARE: PowerPC VSX SIMD instructions supported */
+#  define PNG_POWERPC_VSX   10 /* HARDWARE: PowerPC VSX SIMD instructions
+                                * supported */
 #endif
 #define PNG_OPTION_NEXT  12 /* Next option - numbers must be even */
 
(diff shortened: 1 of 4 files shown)


jbowler:
Tested with and without the code enabled i.e. with and without:

`option DISABLE_ADLER32_CHECK on`

in `pngusr.dfa`

cmake all; cmake test
configure; make all; make check; make clean; make distcheck

This is the only one of all the solutions that I am happy with.  I'd go for the "delete it all" approach too if necessary, but the rest of them are a bug farm.

0-wiz-0:
I found one: [pngcrush](http://www.ki.nu/pkgsrc/reports/current/NetBSD-9.0/20240128.2318/pngcrush-1.8.13/build.log):
```
pngcrush.c: In function 'main':
pngcrush.c:5523:46: error: 'PNG_IGNORE_ADLER32' undeclared (first use in this function); did you mean 'PNG_NORETURN'?
                     png_set_option(read_ptr, PNG_IGNORE_ADLER32,
                                              ^~~~~~~~~~~~~~~~~~
                                              PNG_NORETURN
pngcrush.c:5523:46: note: each undeclared identifier is reported only once for each function it appears in
*** Error code 1
```

jbowler:
Yeah, well it was written by Glenn :-)  If he was the only person to ever use it we are on pretty safe ground.

https://github.com/glennrp/pmt/pull/2

The same fix has already been made on gentoo, it's probably better to take that one if the maintainer pushed it upstream.

PR #517, opened 2024-01-15
pngcp: remove GNU setjmp warning workround

jbowler:
Prior versons of the GCC warned about the 'dest' parameter of contrib/tools/pngcp.c not being volatile, which isn't necessary because it isn't modified.  This removes the GCC specific fixup.

The function which calls setjmp, cppng() also relied on undefined behavior because it assigned the result of setjmp() to a variable; this is not one of the four uses of setjmp permitted by ANSI-C.  This passes the result previously returned by longjmp via (struct display).  It's very very unlikely that any compiler could have got the code wrong but it is technically undefined.

### DIFF
--- a/contrib/tools/pngcp.c
+++ b/contrib/tools/pngcp.c
@@ -1,6 +1,6 @@
 /* pngcp.c
  *
- * Copyright (c) 2016,2022 John Cunningham Bowler
+ * Copyright (c) 2016,2022,2024 John Cunningham Bowler
  *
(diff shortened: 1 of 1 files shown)


jbowler:
The same issues exist in png_safeexecute() although the case of undefined behavior is even less likely to cause problems.  I'll get rid of the now spurious volatiles and correct the undefined behavior.

ctruta:
Thank you.

PR #518, opened 2024-01-15
png_safe_execute setjmp and GCC warning fixes

jbowler:
png_safe_execute called setjmp in a context where the result is undefined (an assignment statement).  This corrects the code.

The GCC diagnostic -Wmaybe-uninitialized switches on warnings of possible problems in functions calling setjmp and in code which passes a pointer to an uninitialized object to a function taking a (const*) pointer.

At present png_safe_execute seems not to suffer from the later warnings even when the spurious (volatile auto) declarations are removed.  They have been removed (a likely performance benefit because spurious stack variables have been removed) and every auto/register variable in extent at the setjmp call has been marked (const), guaranteeing no modification in the function after or, indeed, before the setjmp call.

The known existing case where -Wmaybe-uninitialized causes spurious warnings is caused by the number formating support in png_format_number; it takes a (const) pointer to the start of the buffer which is never initialized; the buffer is accessed via the 'end' (non-const) pointer and 'start' is, effectively an end point.  (This is an unfortunate feature of Arabic numbers; in LtoR languages they are written backwards...)

The change disables the warning around the code in question with GCC specific pragmas, as used elsewhere for similar problems.

configure; make; make check now succeed with:

CFLAGS="-Wall -Werror" ../configure -enable-werror

and cmake builds pass make; make test.

### DIFF
--- a/png.c
+++ b/png.c
@@ -732,6 +732,7 @@ png_convert_to_rfc1123_buffer(char out[29], png_const_timep ptime)
    {
       size_t pos = 0;
       char number_buf[5]; /* enough for a four-digit year */
+      /* INITIALIZED by PNG_FORMAT_NUMBER below: */
 
 #     define APPEND_STRING(string) pos = png_safecat(out, 29, pos, (string))
 #     define APPEND_NUMBER(format, value)\
--- a/pngerror.c
+++ b/pngerror.c
@@ -255,7 +255,7 @@ void
 png_warning_parameter_unsigned(png_warning_parameters p, int number, int format,
     png_alloc_size_t value)
 {
-   char buffer[PNG_NUMBER_BUFFER_SIZE];
+   char buffer[PNG_NUMBER_BUFFER_SIZE]; /* CORRECT: initialized below: */
    png_warning_parameter(p, number, PNG_FORMAT_NUMBER(buffer, format, value));
 }
 
@@ -272,7 +272,7 @@ png_warning_parameter_signed(png_warning_parameters p, int number, int format,
    if (value < 0)
       u = ~u + 1;
 
-   str = PNG_FORMAT_NUMBER(buffer, format, u);
+   str = PNG_FORMAT_NUMBER(buffer, format, u); /*CORRECT: initializes buffer */
 
    if (value < 0 && str > buffer)
       *--str = '-';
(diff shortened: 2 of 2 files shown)


png.c
```diff
@@ -758,6 +758,8 @@ png_convert_to_rfc1123_buffer(char out[29], png_const_timep ptime)
          APPEND_STRING(PNG_FORMAT_NUMBER(number_buf, format, (value)))
 #     define APPEND(ch) if (pos < 28) out[pos++] = (ch)
 
+#pragma GCC diagnostic push
```
ctruta, line 761: All gcc-specific pragmas should be guarded by `#ifdef __GNUC__`. Let's avoid warnings like "unknown pragma" from compilers that don't know about gcc.

ctruta: One more thing I'm thinking about: does this solution need to be this complicated? If it's just gcc we're wandering about, then I'd rather have a proper alternative solution based on `snprintf`, guarded as follows:
```
#if (defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L)) || defined(__GNUC__)
/* TODO new simple solution, based on standard snprintf */
#else
/* TODO old complicated solution, hand-made */
#endif
```

jbowler: They're all just warnings; only a problem with -Werror.  I'll update the patch to remove them all.

'snprintf' is part of an enormous library and is not ANSI-C so your suggestion introduces two code paths to fix one spurious GCC warning (which can be ignored by -Wno-maybe-initialized).  I only added these things because Glenn insisted on warning-free compilation with -Wall (+GCC, which is why all these hacks worked!)  Clearly time to eliminate them.

ctruta:
Going from the original ideas from @jbowler I fixed these issues but with a different approach.

In commit 7dacc4d5aa8253c1746f8de046492694b4b1e173 I fixed the warnings by memzero-ing the arrays. If an the optimizing compiler is smart enough to see that the initialization is redundant, then it's going to ignore that initialization. Otherwise, a compiler like GCC (notably not Clang, not Microsoft C, etc.) will use that initialization, "for safety", in code that is not performance-critical.

And then, in commit 2a4f0f5aee8b78d8b7afeb91bf9b6ad7f6e0131e I did pretty much what John did, except that I didn't put in those redundant consts. They really aren't necessary, not even for readability. The code is clean and rather obvious, not to mention that it now compiles cleanly, so we should be good.

jbowler:
The (const) thing is a core part of my code style.  I mark autos that I assume will not be changed as (const) for documentation; this applies in particular to CSEs that I hand-craft for code clarity but also in cases where the compiler can't see them.  So you will see extensive use of (const) in the libtests, pngcp, makepng, pngfix; the things so marked are things I assume won't be changed, things not so marked are not assumed to be constant.  (So in png_safe_execute 'int result' is not const even though it doesn't change because it might be changed, as in the previous version of the code which, ironically, was probably wrong because of the change.)

It's such a core part of my style that I do it automatically; *not* doing it is difficult so it just happens (since it is of itself harmless and doesn't alter the compiled code in any way.)

I've been saved from serious errors by this technique; coming back to code I wrote years ago and accidentally modifying something without realising something else in the same function assumes it doesn't change.

PR #520, opened 2024-01-15
Remove GCC7.1 arithmetic overflow fixup

jbowler:
This removes pragmas and the controlling code that quelled warnings generated by GCC7.1 (only) with -Wstrict-overflow=3 and possibly other levels.  Tested with GCC13.2, GCC7.1 is no longer the current version of GCC7 (GCC7.5) and GCC7.1 was replaced by GCC7.2 on August 14, 2017.


jbowler:
The whole block of code is, by code examination, fluff; it's protected by an #if which checks for GCC 7.1 which hasn't been around for ever...

jbowler:
Almost 1%!

PR #225, opened 2018-05-17
LOONGSON MIPS MMI SIMD optimized filter functions on 64 bits OS

zhanglixia-hf:
Implement SIMD vector support for LOONGSON MIPS MMI SIMD equivalent to libpng support for SIMD optimizations for Intel SSE, ARM Neon, MIPS MSA and PowerPC VSX. Achieve speedup appropriate for LOONGSON vector width and processor pipeline in 64 bits OS.

PR #521, opened 2024-01-16
Optimize png16 with loongson mmi for 64-bit os

XiWeiGu:
The author of PR #225  is no longer maintaining it. I have updated this portion of the code and made several fixes on top of it.

### DIFF
--- a/pngpriv.h
+++ b/pngpriv.h
@@ -197,6 +197,16 @@
 #  endif
 #endif
 
+#ifndef PNG_MIPS_MMI_OPT
+#  ifdef PNG_MIPS_MMI
+#    if defined(__mips_loongson_mmi) && (_MIPS_SIM == _ABI64) && defined(PNG_ALIGNED_MEMORY_SUPPORTED)
+#       define PNG_MIPS_MMI_OPT 1
+#    else
+#       define PNG_MIPS_MMI_OPT 0
+#    endif
+#  endif
+#endif
+
 #ifndef PNG_POWERPC_VSX_OPT
 #  if defined(__PPC64__) && defined(__ALTIVEC__) && defined(__VSX__)
 #     define PNG_POWERPC_VSX_OPT 2
@@ -248,7 +258,7 @@
 #endif
 
 #if PNG_MIPS_MSA_OPT > 0
-#  define PNG_FILTER_OPTIMIZATIONS png_init_filter_functions_msa
+#  define PNG_FILTER_OPTIMIZATIONS png_init_filter_functions_mips
 #  ifndef PNG_MIPS_MSA_IMPLEMENTATION
 #     if defined(__mips_msa)
 #        if defined(__clang__)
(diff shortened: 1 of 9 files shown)


XiWeiGu:
@jinboson，please help me review the code.

ctruta (approved):
Looks good to me.

jbowler:
For the record I feel that **all** hardware optimisations should require an explicit "switch on" in the build.  All of them, without exception.  I can test loongson compiles (I haven't tested this one) and I can test the compiles of, I believe, all the other hardware opts that are in there but I have no way of verifying that they work when run on the required very wide variety of hardware.

There's too much danger of whack-a-mole here with the wide range of different hardware available; fix one instance, break another.  I had previously thought that compiler checks would be sufficient but anything that uses a run-time check of any form has the very real chance of failing on some hardware; this has happened in the past with Intel runtime checks.

ctruta:
> For the record I feel that all hardware optimisations should require an explicit "switch on" in the build.

I concur. This is what I want us to do in "libpng-ng".

But for libpng-1.6.x, that ship has sailed. The CMake file is rather atrocious, and it doesn't need to be. The configure script is even worse. The simplest of the makefiles (e.g. `scripts/makefile.gcc`) require that `NOHWOPT` thing (take a look!)

At the time of libpng-1.0, the makefiles were the second-simplest thing to use after
```
ls png*.c | xargs gcc
```

Now, of course, I understand the need of everything that came after that. And yet, grabbing all source files and writing a shell command or two on top of those should work intuitively, without requiring anything like `-DPNG_OBSCURE_THING=1`

The files `scripts/makefile.*` were broken for a very long time during the life of libpng-1.6.x, while nobody (myself included) was looking.

And, let's fix that.

jbowler:
Agreed.  But if you start with just *.[ch] you end up with the same mess.  It's necessary to delete most of the code too.  Much of it is never used and most of it is hardly used.

I guess I might try on my gentoo system with a pngusr.dfa that starts "everything=off".  The big players have their own, heavily hacked, version of libpng and the small players, real people, don't need most of it.

pngpriv.h
```diff
@@ -269,6 +279,22 @@
...
+#     else /* !defined __mips_loongson_mmi  || _MIPS_SIM != _ABI64 */
+#        define PNG_MIPS_MMI_IMPLEMENTATION 0
+#     endif /* __mips_loongson_mmi  && _MIPS_SIM == _ABI64 */
+#  endif /* !PNG_MIPS_MMI_IMPLEMENTATION */
+
+#   if PNG_MIPS_MMI_IMPLEMENTATION > 0
+#      define PNG_FILTER_OPTIMIZATIONS png_init_filter_functions_mips
+#   endif
+#else
+#   define PNG_INTEL_SSE_IMPLEMENTATION 0
```
vsautin, line 295: Seems like this should be `#define PNG_MIPS_MMI_IMPLEMENTATION 0`.

XiWeiGu: Thank you for taking the time to review my code.

ctruta: It actually did make a difference, according to Travis CI, but now it's all fixed and all good. Thank you very much to everyone who contributed!

jbowler: Check for "-Wundef" with GCC; it basically makes the compiler non-ANSI if warnings are treated as errors (-Werror).  It isn't enabled by -Wall or -Wextra in GCC 13 but I'm fairly sure it was in earlier versions of the suite.  It's clearly erroneous to turn it on except for code-correctness.

ctruta: IIRC, `-Wundef` was never enabled by `-Wall` or `-Wextra` in any present or past version of gcc, and yet, it is very useful in early catching real defects introduced by typos or bad copy+pasta, etc. Bonus points should be awarded to this catch, which Travis detected only at run time, while it was obvious to me at compile time.

https://app.travis-ci.com/github/ctruta/libpng/builds/268383302

> Check for "-Wundef" with GCC; it basically makes the compiler non-ANSI if warnings are treated as errors (-Werror).

To be pedantic, it's `-Werror` that makes the compiler non-standard. (Not that I would have any problem with `-Werror`, because I absolutely don't.) It's not `-Wundef`, and it's not any other warning enabled by `-Wextra`, for example.

And to be even more pedantic, people make mistakes, and software engineers are people, and standard committees are people, too. Nothing is perfect, everything is perfectible... have you seen ISO C23 yet? ;-)

> It's clearly erroneous to turn it on except for code-correctness.

Code correctness is the whole point of every single compiler warning that has the purpose to flag code smells of any kind. Included, but not limited to, almost the entire `-Wextra`.

ctruta: Oops, I sent the wrong Travis link; but anyway, I would like to move this debate offline, because I'd like to not spam this review thread any further.

PR #524, opened 2024-01-17
Correctly damage palette index test PNG

jbowler:
The PNG IDAT did not include a '255' entry, the highest entry is '254', this corrects the test PNG to have a palette with only 254 entries so that it triggers the palette index checks.


ctruta:
I integrated your change (thanks!) but I rephrased the subject line to make it less scary and more obvious to the casual reader. The "trick" was to use "quotes" with the word "correctly" 😝

On an unrelated note: could you @jbowler please keep your fixes coming, preferably (although not necessarily) into https://github.com/pnggroup/libpng? I want to release libpng-1.6.41, either today or in a day or two, and then I want to mark that other one as the official one, going forward.

jbowler:
It rebased ok.

There is one very important fix missing; the revert of the commit which broke pngtrans.c png_do_check_palette_indexes.  Without this the "sample-*" bad PNGs don't report the error.  The test is just:

`./pngtest contrib/testpngs/badpal/small-palette-1.png foo.png 2>&1 | egrep 'Read palette index exceeding num_palette'`

Which should succeed but fails for `small-palette-[124].png`  It succeeds (i.e. pngtest outputs the error) for the other five cases including small-palette-8.png because there's yet another bug - signed/unsigned compare on line 784 of pngtrans.c which makes a mess of `num_palette_max `when it starts out -1 (but why is it -1?)

ctruta:
> ```
> ./pngtest contrib/testpngs/badpal/small-palette-1.png foo.png 2>&1 | egrep 'Read palette index exceeding num_palette'
> ```

I'm not seeing what you're seeing.
The only message I'm seeing on my development machine is the usual pngtest complaint
```
Files contrib/testpngs/badpal/small-palette-1.png and foo.png are different
```

jbowler:
That is definitely wrong and I fully synced to glennrp/libpng16 before I posted that.  I'll post the full output of the current HEAD from my own dev system.

jbowler:
Precise repro:

```
git clean -dfx # warning; try git clean -dfxn first!
mkdir build.conf
cd build.conf
../configure
make check # succeeds and builds pngtest
./pngtest ../contrib/testpngs/badpal/small-palette-1.png foo.png
```

This does *not* produce the read IDAT error, even though it should!  I might have mistyped the egrep, since I did it by hand.  I'll append it in a couple of minutes:

PR #526, opened 2024-01-18
Loongson-mips: Fixed typo

XiWeiGu:


jinboson:
Looks good to me.
Sorry for this typo. Maybe this could be avoided, if I took some time to make a detailed review . :(

ctruta:
> Sorry for this typo. Maybe this could be avoided [...]

Nah, don't beat yourself too much. And in any case, I had to review it myself, also.

Unfortunately, these things are easy to miss in review; but fortunately, I was able to catch this in my testing.

Most fortunately of all, this little "accident" has convinced me to update the build so that these typos like this one can be caught easily in the future. So... thank you! 😉

PR #529, opened 2024-01-18
Fix unused platform check and configuration for macOS

zixu-w:
In a similar manner as zlib (https://github.com/madler/zlib/pull/895), libpng contains a header configuration that's no longer valid and hasn't been exercised for the macOS target.

- The target OS conditional macros are misused. Specifically `TARGET_OS_MAC` covers all Apple targets, including iOS, and it should not be checked with `#if defined` as they would always be defined (to either 1 or 0) on Apple platforms.
- `#include <fp.h>` no longer works for the macOS target and results in a compilation failure. macOS ships all required functions in `math.h`, and clients should use `math.h` instead.

This problem has not been noticed until a recent extension in clang (https://github.com/llvm/llvm-project/pull/74676) exposed the issue and broke libpng builds on Apple platforms. The failure can be reproduced now by adding `#include <TargetConditionals.h>` before the block.

### DIFF
--- a/pngpriv.h
+++ b/pngpriv.h
@@ -556,18 +556,8 @@
     */
 #  include <float.h>
 
-#  if (defined(__MWERKS__) && defined(macintosh)) || defined(applec) || \
-    defined(THINK_C) || defined(__SC__) || defined(TARGET_OS_MAC)
-   /* We need to check that <math.h> hasn't already been included earlier
-    * as it seems it doesn't agree with <fp.h>, yet we should really use
-    * <fp.h> if possible.
-    */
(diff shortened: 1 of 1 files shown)


ctruta (approved):
LGTM

ctruta:
Thank you very much for making the kind of contribution that I personally like the most: deleting code that should no longer be in there 😛

I landed your commit with your name added to the AUTHORS file, but may I please kindly ask for the native spelling of your name as well? Although that file is in ASCII right now, I want to convert it to Markdown and Unicode, eventually.

zixu-w:
Thanks for the quick review!

FYI: 王子旭

PR #531, opened 2024-01-19
Relax pngtest-all message testing

jbowler:
Previously the 'make check' test pngtest-all looked for given messages at EOL.  The match failed with Windows/MSYS2 because of the Windows <cr><lf> line endings output by pngtest.  This changes the test to look for the message anywhere in a line; this might give false matches but the specific messages being searched for are not likely to cause this problem.


jbowler:
This is a more relaxed version of my previous suggestion.  It's probably better to use this if the previous suggestion works given that the error messages are so specific and this is likely to be more reliable.

PR #533, opened 2024-01-26
CMakeLists: Enable pkg-config on Windows

jonaski:
In libpng, it's hard-coded to exclude pkg-config on Windows unless it's MinGW or Cygwin. I use pkg-config with Visual Studio 2022 and I have to patch this out for CMake to install the pc file. I'm not sure what the original intention was here, installing the .pc file does no harm even if you do not use pkg-config, but if someone really needs to disable it, it should be a CMake option, not hard-coded like this. create_symlink() also works on Windows. Out of over 50 libraries I build on Windows, libpng is the only one that have this behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1037,23 +1037,27 @@ endif()
 
 # Create pkgconfig files.
 # We use the same files like ./configure, so we have to set its vars.
-# Only do this on Windows for Cygwin - the files don't make much sense
-# outside of a UNIX look-alike.
-if(NOT WIN32 OR CYGWIN OR MINGW)
-  set(prefix      ${CMAKE_INSTALL_PREFIX})
-  set(exec_prefix ${CMAKE_INSTALL_PREFIX})
-  set(libdir      ${CMAKE_INSTALL_FULL_LIBDIR})
-  set(includedir  ${CMAKE_INSTALL_FULL_INCLUDEDIR})
-  set(LIBS        "-lz -lm")
-  configure_file(${CMAKE_CURRENT_SOURCE_DIR}/libpng.pc.in
-                 ${CMAKE_CURRENT_BINARY_DIR}/libpng${PNGLIB_ABI_VERSION}.pc
-                 @ONLY)
-  create_symlink(libpng.pc FILE libpng${PNGLIB_ABI_VERSION}.pc)
-  configure_file(${CMAKE_CURRENT_SOURCE_DIR}/libpng-config.in
-                 ${CMAKE_CURRENT_BINARY_DIR}/libpng${PNGLIB_ABI_VERSION}-config
-                 @ONLY)
-  create_symlink(libpng-config FILE libpng${PNGLIB_ABI_VERSION}-config)
+set(prefix      ${CMAKE_INSTALL_PREFIX})
+set(exec_prefix "\${prefix}")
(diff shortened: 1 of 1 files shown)


jonaski:
The patch is tested to work, see https://github.com/strawberrymusicplayer/strawberry-msvc-dependencies/actions/runs/7660160539/job/20876855934

CMakeLists.txt
```diff
@@ -983,23 +983,19 @@ endif()
...
-  create_symlink(libpng.pc FILE libpng${PNGLIB_ABI_VERSION}.pc)
-  configure_file(${CMAKE_CURRENT_SOURCE_DIR}/libpng-config.in
-                 ${CMAKE_CURRENT_BINARY_DIR}/libpng${PNGLIB_ABI_VERSION}-config
-                 @ONLY)
-  create_symlink(libpng-config FILE libpng${PNGLIB_ABI_VERSION}-config)
-endif()
+set(prefix      ${CMAKE_INSTALL_PREFIX})
+set(exec_prefix ${CMAKE_INSTALL_PREFIX})
+set(libdir      ${CMAKE_INSTALL_FULL_LIBDIR})
+set(includedir  ${CMAKE_INSTALL_FULL_INCLUDEDIR})
```
kmilos, line 989:
```suggestion
set(exec_prefix "\${prefix}")
if(IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
    set(libdir "${CMAKE_INSTALL_LIBDIR}")
else()
    set(libdir "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
endif()
if(IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
    set(includedir "${CMAKE_INSTALL_INCLUDEDIR}")
else()
    set(includedir "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
endif()
```

This ensures the package is relocatable on platforms that allow it (most of them), or uses absolute paths on platforms that don't (like NixOS).

kmilos:
> create_symlink() also works on Windows

There were some issues w/ it before: https://github.com/strukturag/libheif/pull/996

CMakeLists.txt
jonaski: Should the set's be indented 2 spaces not 4 here?

kmilos (approved):
LGTM

I saw the `create_symlink()` custom function handles Windows specially.

CMakeLists.txt
```diff
@@ -983,23 +983,27 @@ endif()
...
+    set(libdir "${CMAKE_INSTALL_LIBDIR}")
+else()
+    set(libdir "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
+endif()
+if(IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
+    set(includedir "${CMAKE_INSTALL_INCLUDEDIR}")
+else()
+    set(includedir "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
 endif()
+set(LIBS        "-lz -lm")
```
kmilos: `-lz` can actually be removed here, there is already `Requires.private: zlib` in `libpng.pc.in`...

jonaski: Removed

jonaski:
Ping?

jbowler:
@ctruta?

PR #535, opened 2024-01-30
Fix contrib/conftest/pngcp.dfa

jbowler:
This was broken by the corrections to the 'palette max' handling; if that is disabled the test of num_palette_max must be removed in pngread.c


jbowler:
CI tests for contrib/conftest/*.dfa would be useful; it's just a matter of copying each file to pngusr.dfa (overwrite it) and running the "make" check.  Some of the tests are failing at present, I'll look at fixing those (the tests need to check the #defines and not do things that won't work!)

jbowler:
Wrong branch

PR #536, opened 2024-01-30
Fix contrib/conftest/pngcp.dfa

jbowler:
This was broken by the corrections to the 'palette max' handling; if that is disabled the test of num_palette_max must be removed in pnread.c


jbowler:
There should be only *ONE* change in there; new lines in pngread.c  Last time I managed to include other changes.

PR #539, opened 2024-01-31
Simplify width check in png_check_IHDR.

cjacek:
This fixes a warning on Clang with MSVC target. In this case, clang defines `_MSC_VER`, not `__GNUC__`, but it's still a subject to the same warning. I could change the condition of `png_gt` to something like `defined(__GNUC__) || defined(__clang__)`, but we could also simplify things by adding a pointer size check. This way, the problematic tautological gt expression is entirely skipped by the compiler by constant folding the first expression.

I noticed it in a build of the version bundled in Wine, the patch avoids the warning both on GCC and Clang (both in GNUC and MSVC modes).

### DIFF
--- a/png.c
+++ b/png.c
@@ -2511,17 +2511,6 @@ png_colorspace_set_rgb_coefficients(png_structrp png_ptr)
 
 #endif /* COLORSPACE */
 
-#ifdef __GNUC__
-/* This exists solely to work round a warning from GNU C. */
-static int /* PRIVATE */
-png_gt(size_t a, size_t b)
-{
-   return a > b;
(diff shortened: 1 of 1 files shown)


jbowler:
@ctruta: I'll supply a PR which removes the offensive (compilist) code and works without the enormous complexity of the suggested change.

jbowler:
Fix is in #540; same code but make the mask a (png_alloc_size_t) not an (unsigned) so that PR does actually fix a bug on 16-bit systems.

cjacek:
Thanks!

PR #545, opened 2024-02-15
add CMake config file that follows CMake conventions

bebuch:
New functionality:

- add a CMake config file to the install, so an installed libpng can be found via `find_package(libpng)`
- mark API header includes as SYSTEM headers in install, so warnings in these headers can be disabled

Changed behavior:

- export the target `libpng` as `libpng::libpng` which follows the CMake convention for exported targets (previous it was just `libpng`)
- install CMake files to `lib/cmake/libpng` instead of `lib/libpng` to follow the CMake convention

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1128,6 +1128,30 @@ if(NOT SKIP_INSTALL_EXPORT AND NOT SKIP_INSTALL_ALL)
           FILE libpng${PNGLIB_ABI_VERSION}.cmake)
 endif()
 
+# Create a CMake Config File that can be used via find_package(PNG CONFIG)
+if(NOT SKIP_INSTALL_CONFIG_FILE AND NOT SKIP_INSTALL_ALL)
+  install(TARGETS ${PNG_LIBRARY_TARGETS}
+          EXPORT PNGTargets
+          RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+          LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+          ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+          FRAMEWORK DESTINATION ${CMAKE_INSTALL_LIBDIR})
+
+  include(CMakePackageConfigHelpers)
+  write_basic_package_version_file(PNGConfigVersion.cmake
+                                   VERSION ${PNGLIB_VERSION}
+                                   COMPATIBILITY SameMinorVersion)
+
+  install(EXPORT PNGTargets
+          FILE PNGTargets.cmake
+          NAMESPACE PNG::
+          DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/PNG)
+
+  install(FILES PNGConfig.cmake
+                ${CMAKE_CURRENT_BINARY_DIR}/PNGConfigVersion.cmake
+          DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/PNG)
+endif()
+
 # TODO: Create MSVC import lib for MinGW-compiled shared lib.
 # pexports libpng.dll > libpng.def
 # lib /def:libpng.def /machine:x86
--- a/PNGConfig.cmake
+++ b/PNGConfig.cmake
@@ -1,0 +1,15 @@
+include(CMakeFindDependencyMacro)
+
+find_dependency(ZLIB REQUIRED)
+
+include("${CMAKE_CURRENT_LIST_DIR}/PNGTargets.cmake")
+
+if(NOT TARGET PNG::PNG)
+  if(TARGET PNG::png_shared)
+    add_library(PNG::PNG INTERFACE IMPORTED)
+    target_link_libraries(PNG::PNG INTERFACE PNG::png_shared)
+  elseif(TARGET PNG::png_static)
+    add_library(PNG::PNG INTERFACE IMPORTED)
+    target_link_libraries(PNG::PNG INTERFACE PNG::png_static)
+  endif()
+endif()


CMakeLists.txt
```diff
@@ -1104,8 +1104,12 @@ endif()
 
 # Create an export file that CMake users can include() to import our targets.
 if(NOT SKIP_INSTALL_EXPORT AND NOT SKIP_INSTALL_ALL)
+  configure_file(libpng-config.cmake.in libpng-config.cmake @ONLY)
+  install(FILES "${CMAKE_CURRENT_BINARY_DIR}/libpng-config.cmake"
+          DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/libpng)
   install(EXPORT libpng
-          DESTINATION ${CMAKE_INSTALL_LIBDIR}/libpng
+          NAMESPACE "libpng::"
```
ctruta: I would prefer the namespace `PNG::` instead of `libpng::`, consistently with the rest of the libpng API. (Cf. the names `png_*` for functions, and `PNG_*` for macros, and `png.h` for the header file, etc.)

Moreover, considering that the CMake module name is `FindPNG`, and similar to `FindZLIB`, I would, again, prefer the name `PNG::` in uppercase, instead of the name `png::` in lowercase.

ctruta (changes requested):
The code looks good overall -- but see my comment about the `libpng` vs. `PNG` namespace.

I would also like to have one commit per one task, submitted on top of the tree, as a general rule. We are sticking to a linear history, going forward, and we reserve the branches for the time when we fork new libpng versions. Could you please squash all commits, then rebase?

Last, if you could please update `scripts/cmake/AUTHORS.md` 🙂

CMakeLists.txt
bebuch: Thanks for the fast reply, I will look into it tomorrow.

It would be great to have consistency with the PNG module. Currently I need an additional wrapper to get this.

To get full consistency by default I would need to do 3 changes:

1. Namespace PNG::
2. Target name export as PNG (so we get PNG::PNG from find_package; the target name during build can remain libpng)
3. Change the install directory to {CMAKE_INSTALL_LIBDIR}/cmake/png

If this looks good to you, I will do it tomorrow. (Alongside with the Authors addition and the squash.)

ctruta: @bebuch any confusion is on us. There are historical reasons, which may as well be called historical excuses, and they date back to the earliest days when the format name was PNG and the library name was _pnglib_ (akin to _zlib_). It was renamed from _pnglib_ to _libpng_ before the release of version 1.0 back in 1995, but everything else named `png_foo` inside remained as-is until this very day. I mean, the short library name was changed; but the long library name (i.e. _The PNG Reference Library_) remained in place.

The `PNG` and `png` namespaces are to be used as follows:
 * C programming: `png_foo` for functions and friends, `PNG_BAR` for macros.
 * C source file names: `pngfoo.c` and `pngbar.h`.
 * CMake names: `PNG_FOO` -- except for the `PNGLIB` version names at the top of the main CMake lists file.
 * Pretty much anything else other than directory names, archive names and pkgconfig file names (or so I think).

The `libpng` namespace is to be used as follows:
 * Directory names: `libpng/`, `include/libpng16/`
 * Library file names: `libpng.a`, `libpng16.a`, `libpng16.so`, and friends.
 * Pkgconfig file names: `libpng.pc`, `libpng16.pc`, `libpng-config`, `libpng16-config`
 * Archive file names: `libpng-1.6.42.zip`, containing `libpng-1.6.42/*`

I hope these rules will at least in part justify path names like `/usr/include/libpng16/pnglibconf.h` 🤨

bebuch: Sorry for the delay and thank you for your fast and detailed explanation! :-)

I created a new PR #550 with only the System Header part.

I will update this PR in the next days. My aim is to provide a CMake Config file which can be used with [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html). My secondary goal is to become directly backwards compatible with [`FindPNG`](https://cmake.org/cmake/help/latest/module/FindPNG.html).

Currently other projects include libpng via CMake's `FindPNG` module:

```cmake
find_package(PNG)
```

I configure my 3rdparty libraries with `CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON`, so they prefer CMake Config Files over CMake Modules. I provide a CMake Config File myself, which mimics the FindPNG Module by creating a Target `PNG::PNG`.

`install_root/lib/cmake/png/PNGConfig.cmake`:

```cmake
if(NOT TARGET PNG::PNG)
    include(CMakeFindDependencyMacro)
    find_dependency(libpng)
    add_library(PNG::PNG INTERFACE IMPORTED)
    target_link_libraries(PNG::PNG INTERFACE libpng::png_shared)
endif()
```

The `find_dependency(libpng)` resolves to the CMake Config File of libpng, that my current PR here provides.

If I understand you correctly, you want the CMake namespace for all targets to always be `PNG::`.

- `PNG::png_shared`
- `PNG::png_static`
- ...

In this case, however, the name of the config file and the directory in which it is installed should also be changed to `PNG`. Otherwise, `find_package(libpng)` would create targets in the namespace `PNG::`. If the file and directory name are also adjusted, `find_package(PNG)` would be used and the targets in the namespace `PNG::` would be obtained as expected.

I think that makes perfect sense. In this case, I would simply create the target `PNG::PNG` and have it forwarded to `PNG::libpng_shared` as shown before. This would give us the same functionality in the config file as in the old `FindPNG` module.

All projects that currently use `find_package(PNG)` would continue to function unchanged, but could simply use the new PNG config file via `CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON` and thus benefit from its advantages.

New projects could explicitly request the config file with `find_package(PNG CONFIG)` and thus build directly against `PNG::libpng_shared` or `PNG::libpng_static` at any time.

ctruta: Thank you. Smaller pull requests that are easier to review and test are always extra-appreciated.

I want to let you know a very badly kept secret: a new libpng version is coming up, tentatively called "libpng-ng". If you want to clean up things and change conventions in the sense of making them healthier, even if that means breaking compatibility, then go for it -- now's the time! (Seriously!!)

I will not apply such changes to the 1.6.x line, but that should be ok for those libpng users who don't want anything changed (because IIABDFI). Many libpng users haven't updated libpng in years, which is fine for them and fine for us, too; even if we do fix bugs in the meantime, they seem to not care unless our new fixes happen to solve their old defects.

On the other hand, if you can bring more small changes for the 1.6.x line, in the sense of them being non-disruptive net benefits, like the one in #550, then, sure, please bring!

> Sorry for the delay and thank you for your fast and detailed explanation! :-)

Please do not worry about delaying. I am, too, sometimes fast but other times very very slow to respond, depending on what I happen to be focusing at any given moment.

CMakeLists.txt
```diff
@@ -698,9 +698,11 @@ if(PNG_SHARED)
   endif()
   target_include_directories(png_shared
                              PUBLIC
-                               $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/libpng${PNGLIB_ABI_VERSION}>
                                $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
                              PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+  target_include_directories(png_shared
+                             SYSTEM PUBLIC
```
sergiud: As a minor remark, the `SYSTEM` modifier is not only redundant but in fact unnecessarily restrictive as include directories of imported targets are brought as system include directories anyway unless [`NO_SYSTEM_FROM_IMPORTED`](https://cmake.org/cmake/help/latest/prop_tgt/NO_SYSTEM_FROM_IMPORTED.html) is set. See also the CMake reference documentation for the [`SYSTEM`](https://cmake.org/cmake/help/latest/prop_tgt/SYSTEM.html#prop_tgt:SYSTEM) property.

ctruta: Thank you for your observation, @sergiud.

One thing that baffles me about the SYSTEM modifier is that it is said to be introduced in CMake 3.25. And yet, the verification of commit 7b888092b9f5d249e7baa7bd43ee14f560cd0ae1 with older versions of CMake was successful. Moreover, I just had a test run on a very old Linux install in a Docker container, running CMake 3.7. That old CMake was happy to take the SYSTEM modifier without even breaking a sweat.

I like it when things work, but not so much when they work without me understanding why...

bebuch: @sergiud Thanks for pointing out, I didn't know that!

CMakeLists.txt
bebuch: That sounds great! A more modern version would be great, even if it takes a long time for it to catch on. On the other hand, [zlib-ng](https://github.com/zlib-ng/zlib-ng) has demonstrated that it's really worth it!

As far as user updates are concerned, we also supply software to other companies. Some of these companies now run security testers on every piece of software and they complain accordingly if outdated 3rd party libs are used. It could well be that such tools will change things in the near future ;-)

bebuch:
I have now changed the PR so that it creates the config files I wanted and does not touch anything else. Ready to merge from my side! 😸

bebuch:
@ctruta If you find some time to work on libpng, it would be nice if you could merge this. That way we can use 1.6.44 without a patch once that is tagged. 🙂

bebuch:
@ctruta Since I just saw that you are active again, it would be nice if you could merge this ;-)

ctruta:
> @ctruta Since I just saw that you are active again, it would be nice if you could merge this ;-)

My apologies for taking so long, and my appreciations for your patience. This PR is next in line.

ctruta:
@bebuch Integrated in master, and many thanks. This is super-useful.

Please don't mind me for modifying your commit directly, instead of asking you to do it in a review. After so much wait, now a lot of people want me to push a new release, sooner instead of later. I hope that you accept my modification -- BUT if you do have objections, feel free to submit another PR.

PR #550, opened 2024-02-21
mark installed PNG headers as system headers

bebuch:
Most modern compilers can disable warnings from system headers. This change enables them to do so.


ctruta:
Integrated (even though GitHub doesn't seem to notice).
Many thanks!

PR #554, opened 2024-04-17
CMake: Properly declare target include directories for generated includes

Green-Sky:
Previously the non targeted `include_directories()` was used, which had issue when using the `png_static` target in a submodule.

I came across this issue why trying to link `sdl3_image` with **static** vendored libs, which links against png_static.
For some reason `png_shared` exported the binary include path just fine, but `png_static` not, resulting in `pnglibconf.h` not being found.
So I cleaned up the cmake a little and explicitly added the include dirs to the targets directly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -363,8 +363,6 @@ else()
   message(STATUS "Could not find an AWK-compatible program")
 endif()
 
-include_directories(${CMAKE_CURRENT_BINARY_DIR})
-
 if(NOT AWK OR ANDROID OR IOS)
   # No awk available to generate sources; use pre-built pnglibconf.h
   configure_file(${CMAKE_CURRENT_SOURCE_DIR}/scripts/pnglibconf.h.prebuilt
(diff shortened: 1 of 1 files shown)


ctruta:
This is a very useful fix.

Integrated in master. Danke schön.

ctruta:
Reopening. Verification failed, unfortunately, and I had to revert.
https://app.travis-ci.com/github/ctruta/libpng/builds/272260184

Green-Sky:
Oh yea, I did not test this on mac. I *think* all we need is the same change a couple of lines down in the "framework" section. But I am not sure.

ctruta:
> Oh yea, I did not test this on mac.

Same here -- but hey, the verification bots did.

Green-Sky:
Is there a way we can try running the ci in the pr? I don't like to make the changes blind.

ctruta:
> Is there a way we can try running the ci in the pr? I don't like to make the changes blind.

I understand. Sorry I haven't set up a GitHub action for verification of pull requests, because of a bunch of reasons that are too long to explain here.

One way for you to go is that you submit, then I'll verify it offline (i.e. run it on my own Mac machine), and let you know if the problem is fixed.

Another way for you would be to set up your own free Travis CI account and link it to your own libpng clone. And then, the automation should automatically pick up your commit, without any further action from my side. In my opinion, that's a bit too involved for just one contributed commit.

My advice: do make the changes blind, even if you don't have a Mac machine available. And then, Trust Me™ 😝

Green-Sky:
I did the change blind. Go ahead and test it :)

ctruta:
> I did the change blind. Go ahead and test it :)

Confirmed, not only by my machine, but also by Travis CI.
https://app.travis-ci.com/github/ctruta/libpng/builds/272261835

Meanwhile, AppVeyor CI is still pending, but I have a feeling it's gonna to work. (Famous last words.)
https://ci.appveyor.com/project/ctruta/libpng/builds/50576846

PR #556, opened 2024-04-26
Added markdown version of libpng-manual

neacsum:
Markdown version of libpng-maual.txt. Suggest replacing text version with this one.

### DIFF
--- a/contrib/docs/libpng-manual.md
+++ b/contrib/docs/libpng-manual.md
@@ -1,0 +1,3379 @@
+libpng-manual - A description on how to use and modify libpng
+
+ Copyright (c) 2018-2022 Cosmin Truta
+ Copyright (c) 1998-2018 Glenn Randers-Pehrson
+
+ This document is released under the libpng license.
+ For conditions of distribution and use, see the disclaimer
+ and license in png.h
+
+ Based on:
+
+ libpng version 1.6.36, December 2018, through 1.6.38 - September 2022
+ Updated and distributed by Cosmin Truta
+ Copyright (c) 2018-2022 Cosmin Truta
+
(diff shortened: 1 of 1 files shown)


jbowler:
@ctruta: maintainer required, maybe @neacsum is a candidate?

ctruta:
If @neacsum will agree to maintain it, that'd be great. But if not, I can maintain it myself also, along with the rest of libpng manuals. One of my plans (of lesser priority) to do for libpng-2.0 is to have all manuals in Markdown format, or in Asciidoc format, or in any other format that serves the double-duty of being readable both in plain-text form and rich-text form.

So I won't integrate this for now, but I'll definitely come back to it, later.

Also, FYI @neacsum, my preference for the markdown'ed manual would be to put it in the `manuals/` directory, replacing the current `libpng-manual.txt`:
https://github.com/pnggroup/libpng/tree/libpng18/manuals/

neacsum:
Sure, I can maintain it. I’ll change location per your suggestion. On Sep 21, 2024, at 03:23, Cosmin Truta ***@***.***> wrote:﻿
If @neacsum will agree to maintain it, that'd be great. But if not, I can maintain it myself also, along with the rest of libpng manuals. One of my plans (of lesser priority) to do for libpng-2.0 is to have all manuals in Markdown format, or in Asciidoc format, or in any other format that serves the double-duty of being readable both in plain-text form and rich-text form.
So I won't integrate this for now, but I'll definitely come back to it, later.
Also, FYI @neacsum, my preference for the markdown'ed manual would be to put it in the manuals/ directory, replacing the current libpng-manual.txt:
https://github.com/pnggroup/libpng/tree/libpng18/manuals/

PR #557, opened 2024-04-26
Restore STDERR in pngtest.c

### LINKED ISSUE #560: Need to restore STDERR in pngtest.c
In [test: Add consistency checks for the PNG_LIBPNG_VER* number](https://github.com/pnggroup/libpng/commit/cc8006c48d90cca8bf380fa69469b08f4edb00c5) the `STDERR` macro was moved from outside an `#ifdef` to inside an `#ifdef`. This broke the code in the else of this `#ifdef` which also uses the `STDERR` macro. It appears the `#define STDERR` should be moved back to where it was to avoid compile errors in the `#else` case.

This can be reproduced by making a change similar to [what the oss-fuzz build currently does](https://github.com/pnggroup/libpng/blob/f1848a3b560ddcad065242268433af475948461e/contrib/oss-fuzz/build.sh#L29C11-L29C76) setting `option WRITE disabled` and then building.

bungeman:
In "test: Add consistency checks for the PNG_LIBPNG_VER* number" [0] the `STDERR` macro was moved from outside an `ifdef` to inside an `ifdef`. This broke the code in the `else` of this `ifdef` which also uses the `STDERR` macro. Move `STDERR` back to where it was to avoid compile errors in the `else` case.

[0] https://github.com/pnggroup/libpng/commit/cc8006c48d90cca8bf380fa69469b08f4edb00c5

Fixes: https://github.com/pnggroup/libpng/issues/560

### DIFF
--- a/pngtest.c
+++ b/pngtest.c
@@ -45,6 +45,11 @@
 
 #include "png.h"
 
+/* This hack was introduced for historical reasons, and we are
+ * still keeping it in libpng-1.6.x for compatibility reasons.
+ */
+#define STDERR stdout
+
 /* Generate a compiler error if there is an old png.h in the search path. */
 typedef png_libpng_version_1_6_44_git Your_png_h_is_not_version_1_6_44_git;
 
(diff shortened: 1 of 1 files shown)


jbowler:
This should be an issue with a repro; i.e. report a bug, don't suggest a fix.

jbowler:
@ctruta: I've now verified this against the repro in #560  It's an issue for libpng16 because it breaks real world configurations.  In particular it breaks configurations which don't have the read support pngtest.c requires - "make" step fails when it compiles pngtest.c

The change fixes this.

ctruta (approved):

ctruta:
Integrated in the main branch -- many thanks.

PR #562, opened 2024-05-30
ARM: disable native assembler optimisations

jbowler:
This change disables assembly of assembler language implementations of
ARM Neon optimisations.  This only affects older GCC versions;
everything else either did nothing or used the GCC "intrinsics".

The change can be explicitly bypassed by defined
PNG_ARM_NEON_IMPLEMENTATION or, of course, by just reverting the two
lines that have material changes!

### DIFF
--- a/pngpriv.h
+++ b/pngpriv.h
@@ -148,6 +148,9 @@
     *
     *    1  The intrinsics code (the default with __ARM_NEON__)
     *    2  The hand coded assembler (the default without __ARM_NEON__)
+    *       NOTE: This is reported to cause security issues so has been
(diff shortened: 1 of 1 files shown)


jbowler:
@billatarm, @rossburton: please review.

billatarm:
I think the right approach here is to remove the `.S` file from compilation when neon intrinsics are natively supported. Give me some time to look at it more.

jbowler:
Yes; this change is insufficient.  See my comments in:

https://github.com/pnggroup/libpng/issues/505

PR #566, opened 2024-07-03
oss-fuzz: Add fuzzing targets for simplified READ API

tyler92:
New target added to libpng_read_fuzzer.cc for simplified READ API.

Although simplified APIs work via low-level API, I hope including the simplified API in fuzzing testing makes sense too.


contrib/oss-fuzz/.build.sh.swp
ctruta, line 1: Oops!...

tyler92: Sorry, force-pushed without this file

Issue #254, opened 2018-08-27
libpng.vers generation/parse bug/patch


ladar:
When I tried to combile the libpng with debug symbols, it barfed (both v1.6.35 and the repo HEAD). If you'd care to reproduce, run the following commands:

```bash
git clone https://github.com/glennrp/libpng.git && cd libpng/
export CFLAGS="-g3"; export CPPFLAGS="-g3"; export CXXFLAGS="-g3"
./configure && make
```
Which will result the following [relevant] error message:
```
/usr/bin/ld:libpng.vers:1: syntax error in VERSION script
```
After I managed to reproduce this error on CentOS 6.10 and CentOS 7.5, I decided to procrastinate, and a take a short safari, whereby I tracked down the anthropod to its nest inside the `vers.tf1` file. When libpng is compiled using the `-g3` flag, it seems the following symbol decides to make an appearance:
```C
#define PNG_EXPORTA(ordinal,type,name,args,attributes) PNG_DFN " @" SYMBOL_PREFIX "@@" name "@;"
```
It seems, the `PNG_EXPORTA` goesn't normally appear. However, when the `-g` flag is used, this elusive beast is coaxed toward the watering hole where it wreaks mayhem on all it encounters.

This errant symbol, causes a problem when `awk` (or if you prefer, `gawk`) is used to transform the `vers.tf1` file into the `vers.rf2` file, using the arcane `dfn.awk` script. This process results in spurious, and as it turns out, lethal, text, at the top of the `vers.tf2`. The first few lines of `vers.tf2` end up looking like:
```C
 SYMBOL_PREFIXname;
PNG16_0 {global:
 png_access_version_number;
 png_set_sig_bytes;
```
The "SYMBOL_PREFIXname;" text is what causes the problem, as `vers.tf2` is renamed to `vers.out` which then gets duplicated, so it ends up with `libpng.vers` name, or, the file indicated in the original error message above. From here it's academic. The `libppng.vers` file is what gives libtool/gcc indigestion. And just like that, my safari was over.

I created a small patch to fix this issue, although it might not be an ideal fix. I simply added extra logic to the `dfn.awk` script, which causes it to skip over the offending PNG symbols while writing out the `vers.tf2` file. Ideally someone who is more familiar with what's happening might want to either fix the vers.c file output, so this issue won't occur, or update the `dfn.awk` script so it can parse, and either transform, or omit, symbols which use the syntax above. 

My kludge is available here:

[...]

Issue #240, opened 2018-07-13
Cross Compilation for Android while having AWK seems to fail


strattist:
Using the git as an external module built with CMake there is some strange condition in the `CMakeLists.txt` (starting line 261):

```
find_program(AWK NAMES gawk awk)

include_directories(${CMAKE_CURRENT_BINARY_DIR})

if(NOT AWK OR ANDROID)
  # Compilation for Android
else()
  # Any other compilation 
endif()
```

Followed by (starting line 519):

```
if(PNG_SHARED)
  # Some stuff

  # No more Android related condition while libpng.vers is not generated if Android compilation
  if(UNIX AND AWK) 
    if(HAVE_LD_VERSION_SCRIPT)
      if(HAVE_LD_VERSION_SCRIPT)
        set_target_properties(png PROPERTIES LINK_FLAGS
          "-Wl,--version-script='${CMAKE_CURRENT_BINARY_DIR}/libpng.vers'")
      elseif(HAVE_SOLARIS_LD_VERSION_SCRIPT)
        set_target_properties(png PROPERTIES LINK_FLAGS
          "-Wl,-M -Wl,'${CMAKE_CURRENT_BINARY_DIR}/libpng.vers'")
      endif()
  endif()
endif()
```

I don't know if I use it correctly. Maybe I shouldn't compile libpng as a shared library for Android ? I found those conditions strange anyway so I just wanted to put my remark here.

markand:
I had the same issue trying to build libpng for Android using CMake, I get the following error:

```
[ 34%] Linking C shared library libpng16.so
/home/markand/android/ndk-bundle/toolchains/x86-4.9/prebuilt/linux-x86_64/lib/gcc/i686-linux-android/4.9.x/../../../../i686-linux-android/bin/ld: error: cannot open /tmp/libpng-1.6.35/build/libpng.vers: No such file or directory
/home/markand/android/ndk-bundle/toolchains/x86-4.9/prebuilt/linux-x86_64/lib/gcc/i686-linux-android/4.9.x/../../../../i686-linux-android/bin/ld: fatal error: unable to parse version script file /tmp/libpng-1.6.35/build/libpng.vers
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [CMakeFiles/png.dir/build.make:296: libpng16.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:396: CMakeFiles/png.dir/all] Error 2
[...]

PR #569, opened 2024-07-22
Proper solution for #254 and #240:

icegood:
gathering of versions of symbols gathers also macro definition itself

Happens also under synology spksrc toochain with GCC_DEBUG_INFO = 1 (have no idea why in all issues debugging matters)

Also added escaped \r in dfn.awk for better maintanability (otherwise some editor went mad)


icegood:
doesn't help. will do better

PR #572, opened 2024-08-06
Add additional warning flags and PNGLIB_FALLTHROUGH

r-barnes:


jbowler:
Since the ISO C standard has defined how to do this across all compilers and all the compilers I have already support it it should be done in the standard way.  Doing it like all the other __attribute settings in pngconf.h might be acceptable in 1.6 but existing systems that recognize /*FALLTHROUGH*/ and don't support __has_attribute need to be handled.

It's moot for 1.8 because I'm about to send a PR to use __has_c_attribute in place of all the current stuff.  In particular the current VC support cannot implement `PNG_USE_RESULT`, `[[nodiscard]]`.

https://learn.microsoft.com/en-us/cpp/code-quality/c26819?view=msvc-170

jbowler:
@ctruta: fixed by #590

jbowler:
@ctruta: fixed by @610

banesbyte:
@jbowler Doubt it was me :)

PR #575, opened 2024-08-09
cmake: Let pngpriv.h control most architecture optimizations

### LINKED ISSUE #372: link error on Intel macOS when building Universal Binary (arm64) using cmake

csparker247:
An attempt to fix #372. Based on the conversation there, removes option logic surrounding ARM optimization.

Alternatively, could restore `PNG_ARM_NEON` option with only `check` and `off` values (or `on` and `off` values).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,182 +137,50 @@ endif()
 if(PNG_HARDWARE_OPTIMIZATIONS)
 
 # Set definitions and sources for ARM.
-if(PNG_TARGET_ARCHITECTURE MATCHES "^(arm|aarch)")
-  if(PNG_TARGET_ARCHITECTURE MATCHES "^(arm64|aarch64)")
-    set(PNG_ARM_NEON_POSSIBLE_VALUES on off)
-    set(PNG_ARM_NEON "on"
-        CACHE STRING "Enable ARM NEON optimizations: on|off; on is default")
-  else()
-    set(PNG_ARM_NEON_POSSIBLE_VALUES check on off)
-    set(PNG_ARM_NEON "off"
-        CACHE STRING "Enable ARM NEON optimizations: check|on|off; off is default")
-  endif()
-  set_property(CACHE PNG_ARM_NEON
-               PROPERTY STRINGS ${PNG_ARM_NEON_POSSIBLE_VALUES})
-  list(FIND PNG_ARM_NEON_POSSIBLE_VALUES ${PNG_ARM_NEON} index)
-  if(index EQUAL -1)
-    message(FATAL_ERROR "PNG_ARM_NEON must be one of [${PNG_ARM_NEON_POSSIBLE_VALUES}]")
-  elseif(NOT PNG_ARM_NEON STREQUAL "off")
-    set(libpng_arm_sources
-        arm/arm_init.c
-        arm/filter_neon_intrinsics.c
-        arm/palette_neon_intrinsics.c)
-    if(PNG_ARM_NEON STREQUAL "on")
-      add_definitions(-DPNG_ARM_NEON_OPT=2)
-    elseif(PNG_ARM_NEON STREQUAL "check")
-      add_definitions(-DPNG_ARM_NEON_CHECK_SUPPORTED)
-    endif()
-  else()
-    add_definitions(-DPNG_ARM_NEON_OPT=0)
-  endif()
-endif()
+set(libpng_arm_sources
+    arm/arm_init.c
+    arm/filter_neon_intrinsics.c
+    arm/palette_neon_intrinsics.c)
 
 # Set definitions and sources for PowerPC.
-if(PNG_TARGET_ARCHITECTURE MATCHES "^(powerpc|ppc64)")
-  set(PNG_POWERPC_VSX_POSSIBLE_VALUES on off)
-  set(PNG_POWERPC_VSX "on"
-      CACHE STRING "Enable POWERPC VSX optimizations: on|off; on is default")
-  set_property(CACHE PNG_POWERPC_VSX
-               PROPERTY STRINGS ${PNG_POWERPC_VSX_POSSIBLE_VALUES})
-  list(FIND PNG_POWERPC_VSX_POSSIBLE_VALUES ${PNG_POWERPC_VSX} index)
-  if(index EQUAL -1)
-    message(FATAL_ERROR "PNG_POWERPC_VSX must be one of [${PNG_POWERPC_VSX_POSSIBLE_VALUES}]")
-  elseif(NOT PNG_POWERPC_VSX STREQUAL "off")
-    set(libpng_powerpc_sources
-        powerpc/powerpc_init.c
-        powerpc/filter_vsx_intrinsics.c)
-    if(PNG_POWERPC_VSX STREQUAL "on")
-      add_definitions(-DPNG_POWERPC_VSX_OPT=2)
-    endif()
-  else()
-    add_definitions(-DPNG_POWERPC_VSX_OPT=0)
-  endif()
-endif()
+set(libpng_powerpc_sources
+    powerpc/powerpc_init.c
+    powerpc/filter_vsx_intrinsics.c)
 
 # Set definitions and sources for Intel.
-if(PNG_TARGET_ARCHITECTURE MATCHES "^(i[3-6]86|x86|amd64)")
-  set(PNG_INTEL_SSE_POSSIBLE_VALUES on off)
-  set(PNG_INTEL_SSE "on"
-      CACHE STRING "Enable INTEL_SSE optimizations: on|off; on is default")
-  set_property(CACHE PNG_INTEL_SSE
-               PROPERTY STRINGS ${PNG_INTEL_SSE_POSSIBLE_VALUES})
-  list(FIND PNG_INTEL_SSE_POSSIBLE_VALUES ${PNG_INTEL_SSE} index)
-  if(index EQUAL -1)
-    message(FATAL_ERROR "PNG_INTEL_SSE must be one of [${PNG_INTEL_SSE_POSSIBLE_VALUES}]")
-  elseif(NOT PNG_INTEL_SSE STREQUAL "off")
-    set(libpng_intel_sources
-        intel/intel_init.c
-        intel/filter_sse2_intrinsics.c)
-    if(PNG_INTEL_SSE STREQUAL "on")
-      add_definitions(-DPNG_INTEL_SSE_OPT=1)
-    endif()
-  else()
-    add_definitions(-DPNG_INTEL_SSE_OPT=0)
-  endif()
-endif()
+set(libpng_intel_sources
+    intel/intel_init.c
+    intel/filter_sse2_intrinsics.c)
 
 # Set definitions and sources for MIPS.
-if(PNG_TARGET_ARCHITECTURE MATCHES "^(mipsel|mips64el)")
-  set(PNG_MIPS_MSA_POSSIBLE_VALUES on off)
-  set(PNG_MIPS_MSA "on"
-      CACHE STRING "Enable MIPS_MSA optimizations: on|off; on is default")
-  set_property(CACHE PNG_MIPS_MSA
-               PROPERTY STRINGS ${PNG_MIPS_MSA_POSSIBLE_VALUES})
-  list(FIND PNG_MIPS_MSA_POSSIBLE_VALUES ${PNG_MIPS_MSA} index_msa)
-  if(index_msa EQUAL -1)
-    message(FATAL_ERROR "PNG_MIPS_MSA must be one of [${PNG_MIPS_MSA_POSSIBLE_VALUES}]")
-  endif()
-
-  set(PNG_MIPS_MMI_POSSIBLE_VALUES on off)
-  set(PNG_MIPS_MMI "on"
-      CACHE STRING "Enable MIPS_MMI optimizations: on|off; on is default")
-  set_property(CACHE PNG_MIPS_MMI
-               PROPERTY STRINGS ${PNG_MIPS_MMI_POSSIBLE_VALUES})
-  list(FIND PNG_MIPS_MMI_POSSIBLE_VALUES ${PNG_MIPS_MMI} index_mmi)
-  if(index_mmi EQUAL -1)
-    message(FATAL_ERROR "PNG_MIPS_MMI must be one of [${PNG_MIPS_MMI_POSSIBLE_VALUES}]")
-  endif()
-
-  if(PNG_MIPS_MSA STREQUAL "on" AND PNG_MIPS_MMI STREQUAL "on")
-    set(libpng_mips_sources
-        mips/mips_init.c
-        mips/filter_msa_intrinsics.c
-        mips/filter_mmi_inline_assembly.c)
-    add_definitions(-DPNG_MIPS_MSA_OPT=2)
-    add_definitions(-DPNG_MIPS_MMI_OPT=1)
-  elseif(PNG_MIPS_MSA STREQUAL "on")
-    set(libpng_mips_sources
-        mips/mips_init.c
-        mips/filter_msa_intrinsics.c)
-    add_definitions(-DPNG_MIPS_MSA_OPT=2)
-    add_definitions(-DPNG_MIPS_MMI_OPT=0)
-  elseif(PNG_MIPS_MMI STREQUAL "on")
-    set(libpng_mips_sources
-        mips/mips_init.c
-        mips/filter_mmi_inline_assembly.c)
-    add_definitions(-DPNG_MIPS_MSA_OPT=0)
-    add_definitions(-DPNG_MIPS_MMI_OPT=1)
-    else()
-    add_definitions(-DPNG_MIPS_MSA_OPT=0)
-    add_definitions(-DPNG_MIPS_MMI_OPT=0)
-    endif()
-endif()
+set(libpng_mips_sources
+    mips/mips_init.c
+    mips/filter_msa_intrinsics.c
+    mips/filter_mmi_inline_assembly.c)
 
 # Set definitions and sources for LoongArch.
-if(PNG_TARGET_ARCHITECTURE MATCHES "^(loongarch)")
-  include(CheckCCompilerFlag)
-  set(PNG_LOONGARCH_LSX_POSSIBLE_VALUES on off)
-  set(PNG_LOONGARCH_LSX "on"
-      CACHE STRING "Enable LOONGARCH_LSX optimizations: on|off; on is default")
-  set_property(CACHE PNG_LOONGARCH_LSX
-               PROPERTY STRINGS ${PNG_LOONGARCH_LSX_POSSIBLE_VALUES})
-  list(FIND PNG_LOONGARCH_LSX_POSSIBLE_VALUES ${PNG_LOONGARCH_LSX} index)
-  if(index EQUAL -1)
-    message(FATAL_ERROR "PNG_LOONGARCH_LSX must be one of [${PNG_LOONGARCH_LSX_POSSIBLE_VALUES}]")
-  elseif(NOT PNG_LOONGARCH_LSX STREQUAL "off")
-    CHECK_C_COMPILER_FLAG("-mlsx" COMPILER_SUPPORTS_LSX)
-    if(COMPILER_SUPPORTS_LSX)
-      set(libpng_loongarch_sources
-          loongarch/loongarch_lsx_init.c
-          loongarch/filter_lsx_intrinsics.c)
-      set_source_files_properties(${libpng_loongarch_sources}
-                                  PROPERTIES
-                                  COMPILE_FLAGS "-mlsx")
-      add_definitions(-DPNG_LOONGARCH_LSX_OPT=1)
-    else()
-      message(FATAL_ERROR "Compiler does not support -mlsx option")
-    endif()
-  else()
-    add_definitions(-DPNG_LOONGARCH_LSX_OPT=0)
-  endif()
+set(libpng_loongarch_sources
+    loongarch/loongarch_lsx_init.c
+    loongarch/filter_lsx_intrinsics.c)
+include(CheckCCompilerFlag)
+check_c_compiler_flag("-mlsx" COMPILER_SUPPORTS_LSX)
+if(COMPILER_SUPPORTS_LSX)
+  set_source_files_properties(${libpng_loongarch_sources}
+                              PROPERTIES
+                              COMPILE_FLAGS "-mlsx")
 endif()
 
 else(PNG_HARDWARE_OPTIMIZATIONS)
 
-# Set definitions and sources for ARM.
-if(PNG_TARGET_ARCHITECTURE MATCHES "^(arm|aarch)")
-  add_definitions(-DPNG_ARM_NEON_OPT=0)
-endif()
-
-# Set definitions and sources for PowerPC.
-if(PNG_TARGET_ARCHITECTURE MATCHES "^(powerpc|ppc64)")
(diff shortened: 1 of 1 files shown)


jbowler (commented):
The specific commit is OK as a minimal fix for your specific problem but it does not address the same issue WRT Intel, MIPS and the other architectures.

I guess I'd only approve something that showed that you are willing to become a maintainer; otherwise the whole project just gets trapped in half fixes that break more than half of the other cases.

csparker247:
Understood. This was just meant as a starting point, and I hadn't realized that the same problems were affecting the other architectures. Given the conversation in the original issue, I think I understand what needs to happen. I'll revise this early next week and see if we can't get things in better shape.

csparker247:
Updated so that CMake now defers to `pngpriv.h` for most architectures when `PNG_HARDWARE_OPTIMIZATION=ON`. Only the loongarch optimizations remain separately conditional inside CMake.

Notes: 
- I need to squash and rebase this.
- This is an unrelated by important CMake PR: #545

CMakeLists.txt
```diff
@@ -133,126 +133,26 @@ endif()
...
-    add_definitions(-DPNG_MIPS_MMI_OPT=0)
-    endif()
-endif()
+set(libpng_mips_sources
+    mips/mips_init.c
+    mips/filter_msa_intrinsics.c
+    mips/filter_mmi_inline_assembly.c)
 
 # Set definitions and sources for LoongArch.
 if(TARGET_ARCH MATCHES "^(loongarch)")
```
csparker247: This could also be removed in favor of the sources being added only if `COMPILER_SUPPORTS_LSX`.

csparker247: Now implemented as described. Easy enough to rollback.

jbowler: > This could also be removed in favor of the sources being added only if `COMPILER_SUPPORTS_LSX`.

I'm against that; building the same files on all architectures reduces the possibility of bugs.  COMPILER_SUPPORTS_LSX is, itself, a bug since it will break a Universal Binary including Loongarch.  Well, to be exact, the build couldn't have the hardware support.

An approach is to punt the problem to the Loongarch maintainers and require that the compiler flags be set to enable the code.  In other words remove the auto-detect completely so CFLAGS has to include -DCOMPILER_SUPPORTS_LSX or, better, they fix the Loongarch compiler!  That's a separate issue.

jbowler (changes requested):
The COMPILER_SUPPORTS_LSX stuff should not affect the files that are compiled; it should be a minimal fix to work round the lack of functionality in the Loongarch compiler.

CMakeLists.txt
csparker247: Always adding the files is easy enough, and I can at least confirm that this doesn't break my universal library builds on macOS. I can't test support for loongarch, though, because I don't have the appropriate toolchain. 

I would still suggest using the `check_c_compiler_flag` block to add the `-mlsx` flag, though. This is idiomatic CMake for checking whether a compiler flag is supported and matches the `configure` behavior (which tries to compile test code with the flag). I don't currently print a warning when the test fails, like `configure` does, but I could. By default, this is logged when running cmake:
```
-- Performing Test COMPILER_SUPPORTS_LSX
-- Performing Test COMPILER_SUPPORTS_LSX - Failed
```

From what I can tell, `-mlsx` is currently required by at least Clang and GCC. And if the compilers ever remove `-mlsx` in favor of setting `__loongarch_sx` directly when supported, then the code in `pngpriv.h` works as is and the extra flag doesn't get added. It's also just nice for downstream users to not have to remember to run `CFLAGS=-mlsx cmake ...` when it's pretty easy for us to test for that.

csparker247:
I've reverted the `pngpriv.h` changes from your patch in this PR. I can open another PR to fix that or you can. Doesn't matter to me. This should now be exclusively focused on the CMake issues.

jbowler:
The Loongarch issue, which is also separate, may be moot:

https://gcc.gnu.org/onlinedocs/gcc/LoongArch-Options.html

I find it very difficult to believe that the GCC authors did not include some _PREPROCESSOR directive to indicate that LSX SIMD support was being used.  I'll try to investigate that and it needs to be fixed (when it can be).

csparker247:
There's `__loongarch_sx`, but it seems like `-mlsx` has to be passed for it to be set?

jbowler:
> There's `__loongarch_sx`, but it seems like `-mlsx` has to be passed for it to be set?

As I read the latest documentation; https://gcc.gnu.org/onlinedocs/gcc/LoongArch-Options.html there are specific arch types which include LSX and, for that matter, the 256-bit LASX.  There is also -msimd=lsx (or lasx which apparently enables LSX but also turns on vectorisation).  I'll need to build a cross toolchain to test, this takes time.  I'll also try to build a GCC multi-lib compiler with as much "multi" as I can.

The situation with Loongarch is probably identical to that with 32-bit ARM NEON; the compiler has to be told which instruction set to generate and even then -mfpu=neon (or a coproc variant which includes NEON) may have to be set if some CPUs can be built without NEON.  It's a bit like the instruction set; it can be 64-bit, 32-bit or 16-bit (the true RISC instruction set :-)  -mthumb gets 16-bit (which can be mixed with 32 or 64 bit) but the 32/64 switch is apparently an arch switch

The model for NEON is to require the setting to be enabled in the compiler, either by -mfpu..., -march=... on a single target or by building a multi-lib GCC where one option or more options include NEON.  The same thing should happen for Loongarch.

Still, it's a separate issue; not this PR.

csparker247:
Makes sense. Do you want me to remove the auto-flag set then?

jbowler:
> Do you want me to remove the auto-flag set then?

No: a minimal change so that the sources are always added and `PNG_LOONGARCH_LSX_OPT=1` is not set (as with the other architectures) is best.  Some other cleanup will be desirable (e.g. the API and CHECK options for NEON) but that can happen separately.

csparker247:
Alright, this is squashed and ready then, I think.

jbowler:
Ok, I'll gh pr pull it and do my basic tests probably later today my time (it's noon here).  I don't think I'll have time to set up a Loongarch system but I don't think that's an issue.

jbowler (approved):
Perfect.  Verified with AMD64, AArch64 and ARMv7 (various flavours).  cmake with no options now does the same as configure with no options and correctly builds or does not build the NEON hardware support.   Tested with gcc-14.

I haven't tested with multi-lib; I only have single-target GCC builds at present.

jbowler:
@ctruta; I'm very much in favour of having this go in.  It isn't Apple specific but Apple are using cmake build systems so the fact that cmake doesn't work on what might be termed "official" Apple builds (multi-arch) has left me suggesting "you must use configure" several times.

The Apple approach is also well engineered; the only alternatives I know for multi-arch support are based on interpreters (MicroPython, CircuitPython, Arduino, Lua, Forth etc.)

ctruta:
> @ctruta; I'm very much in favour of having this go in.

I concur. And yet...

In my local testing, the resulting build on both Linux/amd64 and FreeBSD/amd64 did not contain the SSE2 code. I looked into `filter_sse2_intrinsics.c.o` and `intel_init.c.o`, and they were empty.

Repro:
```
export CI_CMAKE_VARS="-DPNG_HARDWARE_OPTIMIZATIONS=ON"
export CI_NO_TEST=1
export CI_NO_CLEAN=1
./ci/ci_verify_cmake.sh
```

I think the missing `add_definitions(-DPNG_INTEL_SSE_OPT=1)` and friends might have something to do with it.

In any case, this PR needs to be not only updated, but also tested to confirm that the SIMD optimizations, are indeed, in the compiled binary.

ctruta (changes requested):
Reviewing: yes, in principle; but no, in practice, because it failed in my testing. (See my other comment.)

ctruta:
@csparker247 tell you what: if you fix and resubmit, that'd be great; but in the meantime, I will push my own CMake fix for FreeBSD/amd64. You will end up deleting it later on, which is to be expected. Just make sure you rebase before you start fixing.

csparker247:
@ctruta I'm on ARM (Apple Silicon) and can't really test SSE symbol creation, but it seems like #576 fails because it doesn't have my changes and this PR might be failing because it doesn't have @jbowler's changes. Easiest thing is to just cherry-pick and rebase one into the other, but that will make the final PR an edit of CMake AND Makefile AND pngpriv.h. Does that bother you?

ctruta:
Thank you @csparker247 for the clarification.

There is a peculiar downside with @jbowler's commit cbf24df9f2c7d8b231b9a0ea98020a28b5d41d0a, meaning that it might break peculiar kinds of custom libpng builds at the users' ends in peculiar use cases.

So I'd like to give it a try tomorrow, and see if I can overcome that downside somehow. And if I cannot, then, ok, I will apply your change (on top of a revert of the FreeBSD fix which just went in) anyway, because, really, I do see the point in what you're trying to solve in here.

ctruta (dismissed):
Changing my review from "request changes" to "approve".
(Conflicts with the recent commit 33ef48b6d65ebf47bc0c60988a430706f586321b notwithstanding.)

csparker247:
@ctruta At a glance, it looks like the changes to pngpriv.h (removing the extra definition checks around SSE detection) are all that would be required to make these CMake changes work as expected. I can't say what effect that alone would have on the Makefile configuration, though.

jbowler:
> > @ctruta; I'm very much in favour of having this go in.
> 
> I concur. And yet...
> 
> In my local testing, the resulting build on both Linux/amd64 and FreeBSD/amd64 did not contain the SSE2 code. I looked into `filter_sse2_intrinsics.c.o` and `intel_init.c.o`, and they were empty.

Yes; this is #576  As explained in the discussions Glenn quite deliberately disabled the Intel SSE support; the person who did the libpng build was obliged to add this obscure #define.  The more recent cmake implementation broke this by bypassing the check, this puts things back to the way Glenn intended and #576 then changes the behavior in both cmake and configure so that Intel SSE behaves like everything else.

There are two pull requests because #576 (building Intel SSE by default) is completely separate from the Universal Build issue which is fixed here.

#576 just removes the gatekeeper.

jbowler:
> There is a peculiar downside with @jbowler's commit [cbf24df](https://github.com/pnggroup/libpng/commit/cbf24df9f2c7d8b231b9a0ea98020a28b5d41d0a), meaning that it might break peculiar kinds of custom libpng builds at the users' ends in peculiar use cases.

No because `PNG_INTEL_SSE` was never set **except** in custom builds.  The change in #576 now makes that setting spurious because the `pngpriv.h` stuff that **is now** compiled is the stuff that **was** compiled only when `PNG_INTEL_SSE` was set.

The change in #576 is that it turns on SSE optimizations in `configure` builds.  `cmake` builds do not change in this regard with 575+576.

_Either there is some reason for not building SSE by default, in which case `cmake` should not be doing it or it's fine to build by default in which case configure should be doing it too._  Given that no one has yet complained about `cmake` building it by default it seems likely that it is fine so enabling Intel SSE by default on real OS distributions of libpng (which use `configure`) seems fine too.

jbowler:
> @ctruta At a glance, it looks like the changes to pngpriv.h (removing the extra definition checks around SSE detection) are all that would be required to make these CMake changes work as expected. I can't say what effect that alone would have on the Makefile configuration, though.

The problem is that `scripts/makefile.*` are pretty much unmaintained.  makefile.clang, makefile.darwin, makefile.gcc makefile.linux and makefile.std all explicitly turn off the hardware optimizations.  makefile.vcwin-arm64 adds support for the ARM Neon optimizations (i.e. it sets everything up)   Many of the other makefiles are for systems that don't have hardware optimizations.

One way round this is to disable all hardware optimizations in pngpriv.h then switching them on by default in pnglibconf.h but not in scripts/pnglibconf.h.prebuilt.  So, with #575, the default is "check" in cmake and configure but "off" in scripts/makefile.* unless turned on.  Since only scripts/makefile.vcwin-arm64 builds the hardware optimizations it would switch them on but, indeed, then could be simplified to -DPNG_HW_SUPPORTED

The current arrangement in the makefiles is a mess because every new hardware optimization has to be accompanied by changes to all the "general purpose" scripts/makefiles

jbowler:
@csparker247 I'm trying to build a script to test this properly; I think my previous tests were completely bogus and just used the Intel behavior.  This only worked because PNG_INTEL_SSE was *not* defined by cmake so when I did a cross-build I did, in fact, get a successful build.

It looks like @ctruta's recent changes have broken this; my test script now fails on teh aarch64 cmake cross build cases **without** this PR.  armv7a still pass but that is probably because I haven't used -mfpu=neon so the compile time checks fail.

Looking at aarch64 I can see that it builds the Intel code, not the ARM code.  Looking at cmake documentation it seems that there is no cross compilation support in libpng (so my tests were completely pointless.)

Yet, how on earth do Universal Binaries work?  The whole thing is cross compilation multiple times.  I'm getting the feeling that only a person with an Apple developer setup can test that.

At present I will drop cmake cross-compiles since they clearly don't work in my environment.  It looks like @ctruta is making massive changes in this area so maybe this will be fixed.

csparker247:
Yes, on newer Apple Clang versions, providing the compiler flags `-arch x86_64 -arch arm64` will build "universal binaries" ([Apple's definition](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) which is a special binary format containing symbols for both architectures. My understanding is that this would not work with LLVM Clang on macOS, as it's specifically Apple Clang which automatically cross compiles for the two targets and packages up the resulting binary objects. 

Unless something has changed recently, CMake does not natively support multiarchitecture builds in general (I vaguely remember seeing some post on the CMake forums about the conceptual difficulty in implementing general multiarch builds with a single CMake cache or something). However it does provide `CMAKE_OSX_ARCHITECTURES` on macOS which essentially just adds corresponding arch flags to the compiler commands. I'm short, CMake only supports it because Apple Clang makes it easy.

I will also say that I've had difficulty getting universal libraries to build on GitHub's Intel macOS runners, so all of my universal builds happen on Apple Silicon. What I don't know is whether optimizations like SSE will be enabled when compiling the Intel symbols on an Apple Silicon device. My assumption is that they won't be, but I suppose it's possible since Apple controls their hardware. All of this to say, I don't know what to expect in terms of optimizations in a universal library being compiled the apple way. 

I can run tests around this next week, but I'm not sure what will be useful for your purposes.

csparker247:
Also, I don't know how you were testing, but when you want to cross compile with CMake, you need to provide a toolchain:  https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html

I don't think this is something you build _into_ a project's CMakeLists.txt, but rather you expect the toolchain to be provided for the target architecture at configure time. So if this isn't working with the current CMake setup, that's something that probably needs to be ironed out.

jbowler:
@ctruta: all my cross-builds now work **but only with this fix.**  All my previous tests were, in fact, failing because the tests in `CMakeLists.txt` on `PNG_TARGET_ARCH` were not working (they always hit Intel).  With this change, and **only with this change**, the architecture is irrelevant to both `cmake` and `configure`; the build just does-the-right-thing.

> Also, I don't know how you were testing, but when you want to cross compile with CMake, you need to provide a toolchain

Well, I have installed toolchains (based on `gcc`) for a number of architectures and I can just use those with `configure` and, now, with `cmake`:

`CC=aarch64-linux-gnu-gcc "CFLAGS=-Wall -Wextra -Wundef -Werror -O2 -g" cmake ..`

I am now testing that way with just a single clone of pnggroup/master; with your change it seems to work correctly (because the build no longer depends on `PNG_TARGET_ARCH` or, for configure, `HOST_ARCH)`.  I can't run the tests in a cross-build and I haven't written scripts to automate copying the tests over to my RPis.  Instead my test, assuming the build succeeds, is:

`size CMakefiles/png_*.dir/{intel,arm}/*.o`

@ctruta: the above is a quick single command test to verify that the correct hardware is built and, indeed, that incorrect hardware is not built.  Even quicker:

`size CMakefiles/png_*.dir/*/*_init.c.o`

But the size difference between and empty `.o` and a compiled one is different.  `nm` on the same files may be more clear.

ctruta:
Heads-up: today should have been the day in which I publish libpng-1.6.44 and fork off the new branch in which we can incorporate this. But it wasn't... Please give me one more day.

jbowler:
@ctruta: I've removed both of my pending pull requests.

I'm seeing merge conflicts that are meaningless (why delete PNGARG, it does no harm!)   I'll wait until you have merged @csparker247's fix.

jbowler:
Merge conflicts currently resolved but I have to assume a merge order.  Specifically I'm assuming #575 gets merged first then the changes that I have which I've tested on Intel and cross-built for ARM32+neon ARM32-neon (armv7a) and MIPS (no special options).  After that I have the ISO-standard attribute patch.

jbowler:
@csparker247 can you change the base on this from libpng16 to libpng18 (go to the top of the PR, at the right click "edit" then click on the new "base:libpng16" button and chose libpng18.)

Merge conflicts may arise; I'm actually fixing them on my own clones and they are in #589 and descendant (#590).  #589 is built off your changes (i.e. requires and, in fact, incorporates your merged changes).  Keep a separate branch with the original code!  (Always, it's git and it really is a git; you can always force-push the original when everything breaks.)

csparker247:
@jbowler Done.

jbowler:
@ctruta: this fixes #606  With this fix on the libpng18 HEAD I can build everything except armv7a-neon (I'll post a minimal fix for that.)

ctruta:
Hi, please don't mind me, but I changed the state to this PR to "draft".

I just wanted to apply the commit to the 'libpng18' branch as-is, and then I realized that the configure build and the CMake build need to go hand in hand. What one does, the other must do, also: you'll need to change `Makefile.am` just as you changed `CMakeLists.txt`.

Speaking about the CMake build -- please don't mind me -- I was planning to make some changes that will break your patch, unless you beat me to it. If you get to submit your commit in which you modify both the CMake build and the configure build, then I'll apply your commit, and then I'll rework my upcoming CMake modifications. Otherwise, I'll apply my upcoming CMake modifications, and then you'll need to rework your patch in order to fit those (but, hopefully, it won't be too difficult).

I'm saying this explicitly because I don't want you (or anyone) to think that I ain't interested in this build modification. I am, in fact, very interested to apply this build modification.

ctruta (changes requested):
As I wrote in my previous comment: both `CMakeLists.txt` and `Makefile.in` need to be modified in the same manner. Regardless if you do it in the same commit, or in two consecutive commits, I would like to apply these two modifications in lockstep.

jbowler:
Both commits were in place before but it's taken so long that mine just got unmaintainable; merge after merge after merge.

jbowler:
@ctruta: exactly what configure changes are you referring to.  You already checked changes in to libpng16 to remove PNG_INTEL_SSE from pngpriv.h and apart from that configure always met the MacOS requirements and still does.

ctruta:
> @ctruta: exactly what configure changes are you referring to. You already checked changes in to libpng16 to remove PNG_INTEL_SSE from pngpriv.h and apart from that configure always met the MacOS requirements and still does.

There are two issues here, one of which is about meeting the macOS build requirements, and the other, about letting pngpriv.h control the arch-specific optimizations. I was talking about the latter.

From looking at this PR as it is today (i.e. commit 8474066c0542f934a6b0d187a77d83418accf261) I can tell that all of the CMake logic about PNG_ARM_NEON and PNG_INTEL_SEE and friends, being set to `on|off|check`, goes away. (Which I like.) But the same should disappear from `./configure` as well. I had a superficial attempt to make those changes, myself, in configure.ac, but it got hairy, and I abandoned it.

To conclude my answer to your question, it was _this_ `configure.ac` (and `Makefile.in`) change I was referring to.

> Both commits were in place before but it's taken so long that mine just got unmaintainable; merge after merge after merge.

Apologies. Next time I do that for too long, please ping me sooner. I do integrate changes (...eventually) but I admit I'm doing it rather chaotically. I think I can do a better job if I'm more aware of your own priorities, for example.

jbowler:
I think I have the configure.ac change still.  It's not difficult but I don't recall exactly what Seth did to CMakeLists.txt.  IRC he removed some bits completely, which is fine by me, but my original configure.ac changes retained 'off' for compatibility with 1.6.  Given the gentoo bug (the one about removing filter_neon.S) it's probably just a lot safer to remove the options and that is part of what my more extensive changes for 1.8 did.

ctruta:
> I think I have the configure.ac change still.

Ok then. Please, do point me to your changes.

jbowler:
> Ok then. Please, do point me to your changes.

I have to separate them out of the rewrite I did of the hardware handling; they all got merged together.

ctruta:
I see. But are they ready to land? If yes, then let us "Ship It (tm)". And if not, then I -- only suggest -- separate the build changes from the source changes, so that I can integrate the build changes now (so that I won't break them any further) and then, land the source changes later, when they are ready to ship.

jbowler:
I believe I might be able to do that.  I've given up on the merge and am just replicating @csparker247's cmake changes into configure.ac/Makefile.am.  Those changes are fairly straightforward.

csparker247:
@jbowler @ctruta seems like this is in hand, but let me know if I need to do anything (like pushing those Makefile changes to this branch).

jbowler:
@ctruta: #613 contains both the cmake changes (this PR) and the configure ones.  I can't change this PR.

@csparker: the only things I did that are relevant to your commit are to remove the libpng 1.6 check on PNG_INTEL_SSE which prevented use of SSE optimizations by default (so they were always off with your change) and to remove the -lsx compiler check stuff.  I verified that, in fact, it isn't necessary at all; gcc with the relevant options sets compiler flags in exactly the same way as it does for all the other architectures.

ctruta:
So I integrated #613 and now I'm closing this PR. Thanks again!

csparker:
I don't think you meant to email me
________________________________

ctruta:
> I don't think you meant to email me

Indeed. Thanks for letting me know. My message was intended for @csparker247.

PR #579, opened 2024-09-04
Remove cHRM check to accomodate ACES AP1

jbowler:
ACES AP1 has a red endpoint with a negative Z, this triggers the checks
in libpng that ensure that x, y and z (chromaticities) are all >=0.
This removes the checks on the sign of the chromaticities since it is
valid to use negative values for any of them and converts the "internal"
error code return to external (because the internal cases correspond to
negative x, y or z.)

### DIFF
--- a/png.c
+++ b/png.c
@@ -1257,20 +1257,6 @@ png_XYZ_from_xy(png_XYZ *XYZ, const png_xy *xy)
    png_fixed_point red_inverse, green_inverse, blue_scale;
    png_fixed_point left, right, denominator;
 
-   /* Check xy and, implicitly, z.  Note that wide gamut color spaces typically
-    * have end points with 0 tristimulus values (these are impossible end
-    * points, but they are used to cover the possible colors).  We check
-    * xy->whitey against 5, not 0, to avoid a possible integer overflow.
-    */
-   if (xy->redx   < 0 || xy->redx > PNG_FP_1) return 1;
-   if (xy->redy   < 0 || xy->redy > PNG_FP_1-xy->redx) return 1;
-   if (xy->greenx < 0 || xy->greenx > PNG_FP_1) return 1;
-   if (xy->greeny < 0 || xy->greeny > PNG_FP_1-xy->greenx) return 1;
-   if (xy->bluex  < 0 || xy->bluex > PNG_FP_1) return 1;
-   if (xy->bluey  < 0 || xy->bluey > PNG_FP_1-xy->bluex) return 1;
-   if (xy->whitex < 0 || xy->whitex > PNG_FP_1) return 1;
-   if (xy->whitey < 5 || xy->whitey > PNG_FP_1-xy->whitex) return 1;
-
    /* The reverse calculation is more difficult because the original tristimulus
     * value had 9 independent values (red,green,blue)x(X,Y,Z) however only 8
     * derived values were recorded in the cHRM chunk;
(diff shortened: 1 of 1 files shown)


jbowler:
REVIEW REQUIRED: @mbechard

jbowler:
@fuzzers: please review this with fuzzed cHRM values.

jbowler:
@mm2 the original (way, way back; 10, 15 years) issue was produced by a PNG with a cHRM which had colinear (IRC) rgb xy points.  This was before any of the checks and caused issues somewhere (it was a long time ago), possibly CIEXYZ with NaN values.  Somehow the result was a crash in lcms (possibly the author of the bug passed signalling NaN values in; I don't remember lcms supporting endpoints specified as chromaticities.)

This change should retain the check on colinear values but will allow cHRM chunks which result in CIEXYZ triples (to use the Windows terminology) with negative Z values, at least for the ACES AP1 and AP0 colour spaces.

jbowler:
If verified this closes https://github.com/pnggroup/libpng/issues/578

mm2:
> @mm2 the original (way, way back; 10, 15 years) issue was produced by a PNG with a cHRM which had colinear (IRC) rgb xy points. This was before any of the checks and caused issues somewhere (it was a long time ago), possibly CIEXYZ with NaN values. Somehow the result was a crash in lcms (possibly the author of the bug passed signalling NaN values in; I don't remember lcms supporting endpoints specified as chromaticities.)

Thanks for let me know. FYI, ICC addressed use of negative XYZ many years ago
[https://color.org/technotes/Guidelines_on_the_use_of_negative_PCSXYZ_values.pdf](https://color.org/technotes/Guidelines_on_the_use_of_negative_PCSXYZ_values.pdf)

mbechard:
Yep, looks good to me and I'm able to write->read a PNG using AP1 primaries.

jbowler:
> ICC addressed use of negative XYZ many years ago 

Yes; negative XYZ were always meaningful.  Unfortunately the original PNG spec defined the chromaticity values as "unsigned".  This doesn't arise in the adaptation case because the negatives come from the adaptation and it doesn't arise with ACES AP1 (that was just my over-enthusiastic error checks).  It does arise in PNG with AP0 because the red y value is negative and significantly so; the rounding hack I suggested to @mbechard won't help.

@ctruta; this really needs to go into libpng16; it's an unexploded bomb that only hasn't been reported before because no one (apparently) has generated PNG files with the ACES chromaticities.  That's going to change as a side effect of the W3C initiatives; these will make PNG more appropriate to video workflows and will, indirectly, encourage the use of wide gamut colour spaces.

It is a bug fix, nope, not security but then surely neither are changes to parts of the build system that have never worked?

I suspect @mbechard might support me in pushing for this.

mbechard:
Yes for sure. Although I'm currently defaulting to ICC profiles if they are available, I think that will go away for the more common color spaces (ACES/DCI-P3/Rec2020) as they are easily defined and manipulated with matrices, which allows them to be done on the GPU. This will be important for high-res image sequence playback where I suspect running the image through an ICC transform on the CPU will end up being the bottleneck for playback performance. Being able to use the cHRM instead of ICC will be optimal in the future, when it's appropriate.

ctruta (approved):
👍

ctruta:
> @ctruta; this really needs to go into libpng16

Done.
@jbowler: off-topic: if you believe that other pending PRs need to go into libpng16 also, please let me know before we turn it into a maintenance-only branch.

jbowler:
> if you believe that other pending PRs need to go into libpng16 also, please let me know before we turn it into a maintenance-only branch.

The only one that I'm aware of which you haven't checked in to libpng16 is #557

I'll keep looking.

ctruta:
> The only one that I'm aware of which you haven't checked in to libpng16 is #557

Ha! Good catch.

> I'll keep looking.

Thank you @jbowler.

PR #582, opened 2024-09-07
Use ISO standard attributes

jbowler:
This removes the compiler specific code in favour of the newly
standardized __has_c_attribute.  This is supported in the principle
compilers (gcc, clang, vc) and, now that it is standard, should be
supported in other compilers too, for example icc and, presumably if
they are maintained, Watcom and Borland.

Passes multiple tests with gcc and clang including cross-compilation
with gcc.


jbowler:
Merge conflict disaster area.

PR #584, opened 2024-09-08
copyright-issues

jbowler:
Various of the files in the current distribution include the
words "All rights reserved" in a copyright statement.  In
some cases this includes files which are currently compiled
into libpng (the neon palette intrinsics and the loongarch
sx).

This **draft** PR fixes this by simply removing all such
files:

1) It seems reasonable that a FOSS distribution should only
include files that are subject to a FOSS license of one form
or another.

2) The "all rights reserved" comments seem to trump a
subsequent libpng license.

3) The libpng code itself should only include code
distributed with the libpng license and that should be
unambigous.

The PR includes fixups for the files which do get compiled
into libpng (the second commit).  This will create an
unfortunate merge conflict which is why this is a draft; it
could be applied directly to libpng16 without damage but it
certainly conflicts with other PRs I have submitted for
libpng18.

- **Libpng taints**
- **Fix builds for removal of copyrighted files**

### DIFF
--- a/png.c
+++ b/png.c
@@ -1257,20 +1257,6 @@ png_XYZ_from_xy(png_XYZ *XYZ, const png_xy *xy)
    png_fixed_point red_inverse, green_inverse, blue_scale;
    png_fixed_point left, right, denominator;
 
-   /* Check xy and, implicitly, z.  Note that wide gamut color spaces typically
-    * have end points with 0 tristimulus values (these are impossible end
-    * points, but they are used to cover the possible colors).  We check
-    * xy->whitey against 5, not 0, to avoid a possible integer overflow.
-    */
-   if (xy->redx   < 0 || xy->redx > PNG_FP_1) return 1;
-   if (xy->redy   < 0 || xy->redy > PNG_FP_1-xy->redx) return 1;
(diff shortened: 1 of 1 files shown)


jbowler:
@ctruta this may be a libpng16 issue.  I don't understand the legal consequences and I only noticed the weird copyright statements because I was looking at the loongarch sx code and the ARM Ltd palette code.

jbowler:
All of the files except arm/palette_neon_intrinsics.c have what I assume is the original license somewhere, in gregbook there's a LICENSE file, the the Loongarch stuff the original license is still there (it looks like they copied one or other of the hardware variants).  In the ARM Ltd case I had to go back to the original version to confirm that this was the original license.

ctruta:
> Various of the files in the current distribution include the words "All rights reserved" in a copyright statement.

In my non-professional opinion, this is a non-issue.

Among a bunch of things that I **_do_** know is that real lawyers working for real corporation have vetted not only the libpng license (including the new "libpng license version 2.0"), but also the libpng code, with all of the enclosed copyright notices, to be 100% legit open-source software.

Submitting the "libpng license version 2.0" to OSI for approval as a legacy OSS license is a lower-priority (but not insignificant) item on my TODO list.

PR #585, opened 2024-09-09
ARM64: Export Neon symbols for shared library

mshakilt:
This change exposes the arm specific symbols to public API. While building PNG as shared library and linking it with other shared libraries like FreeImage, we need to make the neon implementations visible from outside.

### DIFF
--- a/arm/arm_init.c
+++ b/arm/arm_init.c
@@ -14,6 +14,7 @@
 #define _POSIX_SOURCE 1
 
 #include "../pngpriv.h"
+#include "../png.h"
 
 #ifdef PNG_READ_SUPPORTED
 
@@ -57,7 +58,7 @@ static int png_have_neon(png_structp png_ptr);
 #  error "ALIGNED_MEMORY is required; set: -DPNG_ALIGNED_MEMORY_SUPPORTED"
 #endif
 
-void
+void PNGAPI
 png_init_filter_functions_neon(png_structp pp, unsigned int bpp)
 {
    /* The switch statement is compiled in for ARM_NEON_API, the call to
--- a/arm/filter_neon_intrinsics.c
+++ b/arm/filter_neon_intrinsics.c
@@ -12,6 +12,7 @@
  */
 
 #include "../pngpriv.h"
+#include "../png.h"
 
 #ifdef PNG_READ_SUPPORTED
 
@@ -47,7 +48,7 @@
 
 #if PNG_ARM_NEON_OPT > 0
 
-void
+void PNGAPI
 png_read_filter_row_up_neon(png_row_infop row_info, png_bytep row,
    png_const_bytep prev_row)
 {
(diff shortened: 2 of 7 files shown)


jbowler:
I suggest you just copy the code.  It is likely to change.  png_row_info might also change.

There is no real justification for exporting any of the transforms, least of all filtering: you can get the result just by calling the sequential reader with the transform, and in this case there is not a lot you can do without the de-filtering.  Indeed, you must have some way of de-filtering anyway since the NEON code implements a minute sub-set of the cases.

@ctruta: maybe the OP should be starting a discussion here?

mshakilt:
The issue arrises when we build FreeImage and PNG as dynamic libraries and then use freeimage that pops the following linking error "libfreeimage-3.18.0.so: undefined reference to `png_init_filter_functions_neon' [build] collect2: error: ld returned 1 exit status". We resolved the error by exporting neon functions.

jbowler:
It's not a part of the API.  It never was.  The public API is defined by png.h and pngpriv.h is not installed; it doesn't matter whether you use the static library or the DLL it should be failing with an undeclared function when you compiler your own code!

freeimage-3.18.0 seems to contain a fork of libpng 1.6.35 with extensive changes but the ARM code isn't in the forked directory; there is no "arm" subdirectory despite pngpriv.h containing the PNG_INTERNAL_FUNCTION definitions of the ARM code.  pngpriv.h itself.  pngpriv.h is not included by anything outside Source/LibPNG.

Try adding -DPNG_INTEL_SSE to your builds - your problems are more than just AArch64 builds.

freeimage builds on gentoo and it uses DLLs; in fact it deletes all the code in LibPNG and all the other forked code bases leaving just the FreeImage, FreeImageLib and FreeImageToolkit directories.  I edit a link to the main patch into this comment but there are currently 11 patches in gentoo (freeimage-3.18.0-r11.ebuild in the gentoo distribution).

Here it is:

https://dev.gentoo.org/~juippis/distfiles/tmp/freeimage-3.18.0-unbundling.patch

@ctruta: I think they copied libpng-1.6.35 then deleted large amounts of it, including, apparently, the LICENSE file.  The "intel" and "arm" subdirectories are not there so the ARM builds probably fail.  It looks like their own builds on UNIX-like systems don't build LibPNG so I can't work out how broken it is but it looks very broken (and unmaintained.)

Gentoo can, however, build it so they need to fix their own house first.

ctruta:
> It's not a part of the API. It never was. The public API is defined by png.h and pngpriv.h is not installed; it doesn't matter whether you use the static library or the DLL it should be failing with an undeclared function when you compiler your own code!

Confirmed.

We could, in theory, open up the SIMD optimization as a user-serviceable API, but I'm personally concerned about opening something that might turn out to be a can of worms.

mshakilt:
Thanks a lot for clarification. The issue is mainly caused by the libpng fork inside FreeImage. They modified the libpng and actually removed the arm sources but in the CMake script they are still referencing to those deleted files which is causing the issue. Maybe they fix it in the future.

jbowler:
> They modified the libpng and actually removed the arm sources but in the CMake script they are still referencing to those deleted files which is causing the issue.

The intel sources were also removed and CMakeLists.txt references those as well.  The same applies to configure IRC (except that the files never get used.)  This is easy to fix; just remove those lines from CMakeLists.txt!

PR #588, opened 2024-09-11
Prevent overflow in chromaticity calculations

jbowler:
In png_xy_from_XYZ X+Y+Z was calculated without checking for overflow.
This fixes that by moving the correct code from png_XYZ_normalize into a
static function which is now used from png_xy_from_XYZ


ctruta (approved):

ctruta:
Reviewed, approved, and integrated in master, thank you.

PR #589, opened 2024-09-13
[libpng18] reworked SIMD code

jbowler:
This is the rework of the hardware (SIMD) support based on @csparker247's
pull request #575.

It incorporates that request rebased on the current master; the request
should be pulled first.

This passes cross and hosted build checks as follows:

aarch64 configure make all SUCCESS
aarch64 cmake make all SUCCESS
amd64 configure make all SUCCESS
amd64 cmake make all SUCCESS
armv7a configure make all SUCCESS
armv7a cmake make all SUCCESS
armv7a-neon configure make all SUCCESS
armv7a-neon cmake make all SUCCESS
clang cmake make all SUCCESS
mips configure make all SUCCESS
mips cmake make all SUCCESS
riscv64 configure make all SUCCESS
riscv64 cmake make all SUCCESS
clang configure make check SUCCESS
clang cmake make test SUCCESS

- **cmake: Remove architecture-specific checks for hardware optimization**
- **Full compile time support for target-specific code**
- **MIPS support clean-up**

Issue #568, opened 2024-07-09
Implicit fallthroughs


r-barnes:
LLVM 15 with `-Wimplicit-fallthrough` enabled refuses to compile libpng because there are implicit fallthroughs.

Some of these, such as those in `pngread.c` are commented. Others might not be (I'm still iterating).

`[[fallthrough]]` is part of C23 and `__attribute__((fallthrough))` is available now to explicit mark fallthroughs in a way that's unambiguous to compilers.

Would it be alright if I put up a PR introducing a cross-compiler compatible `INTENTIONAL_FALLTHROUGH` macro?

jbowler:
The code has to be ANSI-C, aka ISO-C90 aka C89.  It is expected that the code will be compiled with "-ansi" (or equivalent) and compiling without this, while it typically works, is not supported.  So far as testing is concerned "-pedantic" should be used too.

The code is tested with -Wimplicit-fallthrough with GCC and so far as warnings are concerned GCC really is the only supported compiler; obviously warnings that correspond to real bugs are valid from any compiler but trying to kill warnings in *all* compilers is impossible because they end up demanding different things.

The compiler flags when testing with GCC are "-Wall -Wextra -Werror", there is configure support for -Werror (because configure fails if -Werror is set; use --enable-werror).    Note that those settings are not set by default.  This is because different versions of GCC have different warnings and new versions often have gratuitous additions that get removed PDQ.

clang 18.1.8 -ansi (I think that is pretty much the latest version) does not warn about implicit-fallthrough even with -Wall and -Wextra  An explicit -Wimplicit-fallthrough is required and even then it's just a warning (unless -Werror is given of course).

With '-Wall -Wextra -std=c89 -pedantic" neither GCC nor clang warn.  GCC does check for implicit fallthrough - I deleted line 3333 from pngfix.c and got the warning.

So clang has removed -Wimplicit-fallthrough from -Wall - Wextra; presumably they got too many bugs about not recognizing the comment!  Perhaps they should fix that :-)

I really don't like the idea of replacing a safe, innocuous, comment with a mess of preprocessor.  It's replacing something that works (with GCC at least) and is 100% safe with something that will require continuous maintenance to remove a warning that clang have already disabled themselves!  It achieves nothing because GCC -Wall -Wextra -Werror already ensures that the build fails if there is an implicit fall-through.

r-barnes:
[...]

PR #590, opened 2024-09-13
[libpng18] use ISO standard attributes

jbowler:
This is based on #589.  It replaces the ad-hoc use of compiler specific
attributees by the ISO standard and adjusts the PNG_FUNCTION macro to
work with the ISO standard; function attributes precede function storage
classes.

- **cmake: Remove architecture-specific checks for hardware optimization**
- **Full compile time support for target-specific code**
- **MIPS support clean-up**
- **Use ISO standard attributes**

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -35,10 +35,6 @@ AC_CONFIG_MACRO_DIR([scripts/autoconf])
 # 1.12.2 fixes a security issue in 1.11.2 and 1.12.1
 # 1.13 is required for parallel tests
 AM_INIT_AUTOMAKE([1.13 foreign dist-xz color-tests silent-rules subdir-objects])
-# The following line causes --disable-maintainer-mode to be the default to
-# configure. This is necessary because libpng distributions cannot rely on the
-# time stamps of the autotools generated files being correct
-AM_MAINTAINER_MODE
 
 dnl configure.ac and Makefile.am expect automake 1.11.2 or a compatible later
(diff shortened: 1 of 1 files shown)


jbowler:
Addresses #568 and obsoletes #572 by doing it in a portable way (#572 would have removed support from some compilers in favour of others, this removes some support from compilers which don't keep up to date with the standard but provides support for the current common compilers, gcc clang and vc).

Note that in the current form this completely addresses the original issue; the original /* FALLTHROUGH */ comment is retained however an extra PNG_FALLTHROUGH or (for the apps) FALLTHROUGH  statement is inserted at the same place immediately before it.  This matches both the ISO-C requirements and the older compiler requirements.

The change also takes advantage of the ISO spec to add what are I hope meaningful messages to the PNG_DEPRECATED usages.  Adds PNG_MAYBE_UNUSED for parameters.  PNG_USED is still better if it is done correctly (only on the #if path where the thing really is unused) but [[unused]] is easier, particularly where the parameter really is completely unused (C90 does not support omission of the formal parameter in the definition.)

Makes it easy to add [[sequential]] and [[reproducible]] in the future but these are not currently supported by GCC even though the GCC does support modern definitions of constexpr in C++ (i.e. it supports the [[reproducible]] variant, though [[reproducible]] does other things as well.)

jbowler:
This is development work and depending on the choice of compiler standard to be used may be superfluous, so I've removed it until other more important issues (reported bugs) are resolved and the long term development plan is clear.

PR #594, opened 2024-09-15
[all branches] editorconfig indent 3

jbowler:
Three space indents are required for all C and most other files.
.editorconfig did not have the setting and 4-space indents were
appearing.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,24 +9,29 @@ trim_trailing_whitespace = true
 
 [*.txt]
 indent_style = space
+indent_size = 3
 
 [*.[chS]]
 indent_style = space
+indent_size = 3
 max_doc_length = 80
 max_line_length = 80
 
 [*.dfa]
 indent_style = space
+indent_size = 3
 max_doc_length = 80
 max_line_length = 80
 
 [*.{awk,cmake}]
 indent_style = space
+indent_size = 3
 max_doc_length = 80
 max_line_length = 100
 
 [*.{in,sh}]
 indent_style = space
+indent_size = 3
 max_doc_length = 100
 max_line_length = 100
 


ctruta:
So I didn't apply your commit as-is (because it wasn't actually correct) but I reworked it into 67c99f75cf05b354f89c01c8fc15948300e84143 and then I had to "fix it" (because it wasn't actually broken, but `editorconfig-checker` was) in 77f88338a19a223cc678e1a6a04888a31c78dabf.

I don't know if you noticed @jbowler but I deleted contrib/tools/chkfmt.sh because it served us well but nowadays the `.editorconfig` files as well as programs like `editorconfig-checker` (which you should be able to install with Linuxbrew, for example) are more elaborate and easier to use also.

If you do have `editorconfig-checker` in your path, then ci/ci_lint.sh will pick it up and let you know, even before you submit your PR.

jbowler:
editorconfig-checker is probably part of the "cli" use flag on gentoo; I have editorconfig-core-c installed but it defaults to have the cli flag unset.

ctruta:
> editorconfig-checker is probably part of the "cli" use flag on gentoo; I have editorconfig-core-c installed but it defaults to have the cli flag unset.

I'm not so sure. Considering the name, I would think that `editorconfig-core-c` is a C library that offers the core editorconfig functionality for any apps (primarily text editors) that want to recognize the user preferences.

editorconfig-checker is, on the other hand, a clean-room implementation written from scratch in Go. Most Linux distro (that I'm aware of) haven't picked it up yet, except for Arch Linux, which is why I suggested you to use Linuxbrew. Other install options are listed on its website: https://editorconfig-checker.github.io/

jbowler:
I got it from the github repo, https://github.com/editorconfig-checker/editorconfig-checker

PR #596, opened 2024-09-15
[libpng18] fix make check/test

jbowler:
This corrects the checks to that libpng 10800 does not turn on the
enhanced transform checks in either pngvalid or pngstest.  The correct
fix is to change the 10700 code for comments which explain what aspect
or aspects of the transforms are broken (ideally) or at least state that
the transforms are broken.

### DIFF
--- a/contrib/libtests/pngstest.c
+++ b/contrib/libtests/pngstest.c
@@ -3500,7 +3500,7 @@ main(int argc, char **argv)
    int retval = 0;
    int c;
 
-#if PNG_LIBPNG_VER >= 10700
+#if PNG_LIBPNG_VER == 10700
       /* This error should not exist in 1.7 or later: */
       opts |= GBG_ERROR;
 #endif
--- a/contrib/libtests/pngvalid.c
+++ b/contrib/libtests/pngvalid.c
@@ -2574,7 +2574,7 @@ modifier_init(png_modifier *pm)
  * in the rgb_to_gray check, replacing it with an exact copy of the libpng 1.5
  * algorithm.
  */
-#define DIGITIZE PNG_LIBPNG_VER < 10700
+#define DIGITIZE PNG_LIBPNG_VER != 10700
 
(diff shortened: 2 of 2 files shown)


jbowler:
Rebased on libpng18:HEAD, verified that "make check" still passes.

ctruta (approved):

ctruta:
I integrated it in libpng18 and then I cherry-picked it in libpng16. ("Because I could.") Thank you.

@jbowler FYI, I'm trying to follow (admittedly not very strictly) commit naming rules like these:
https://www.conventionalcommits.org/en/v1.0.0/

I'm rewriting your subject lines if you noticed, and I hope you don't mind.

When I cherry-pick from `libpng18` into `libpng16`, I'm adding the "[libpng16]" prefix, not entirely convinced that I should, but I'm doing it anyway for now. On the other hand, about the main development branch, I'd rather follow common practices used elsewhere, in which they make the subject lines short and sweet.

jbowler:
I'm trying to make the **pull request** titles meaningful in a list of pull requests, so they contain _[branch]_ for example; information which is not relevant to the commit message, and SECURITY, which is probably not useful in a commit title either.

I limit the commit titles to the Linus-specified maximum length, which is really short, so they tend to come out missing English noise words, like _is_ etc.  I notice yours seems to be longer (it's been truncated in the header of the "CODE" tab) so I stop doing that.

ctruta:
If someone ever writes a book titled "The Subtle Art of Writing Commit Messages", I'll make sure to read it.

I "borrowed" the manner to write commit messages from various Google projects, which dates back from my very old days at Google. This, too, has been derived from a Google project, but I forgot which one.
https://www.conventionalcommits.org/en/v1.0.0/

I would like, myself, to keep the subject line short. Not really "Linus short" kind of short, but still, at least, it should be around 66 characters, if not shorter. I don't always succeed, but I still keep trying.

PR #597, opened 2024-09-17
SECURITY [libpng16 ONLY] bug fix: More corrections for cHRM checks

jbowler:
The subtracts in PNG_XYZ_from_xy are producing integer overflow with
some valid but extreme xy values.  This re-introduces the previous
checks but with less limited bounds; sufficient to accomodate the
ACEScg end points (ACES AP1) but not for the ACES AP0 end points.  Those
were not working anyway because libpng reads the cHRM parameters as
unsigned values so they must always be at least 0.

A better solution requires recognizing reasonable negative values (ones
which violate the current spec) and allowing them too, at least on read.

Referenced report:

https://github.com/LibreOffice/core/commit/905e7c1105536c9757fa2c2faf670738aab02595

Original:

https://gerrit.libreoffice.org/c/core/+/173409

The cHRM chunk is recognized as invalid but only after an integer
overflow has occured.  This needs better testing but there is no
obvious way of detecting integer overflow and no obvious way of
generating the invalid cHRM values (given that they tend to be
detected anyway!)

### DIFF
--- a/png.c
+++ b/png.c
@@ -1203,22 +1203,66 @@ png_colorspace_sync(png_const_structrp png_ptr, png_inforp info_ptr)
 #endif /* GAMMA */
 
 #ifdef PNG_COLORSPACE_SUPPORTED
-static int
-png_safe_add(png_int_32 *addend0_and_result, png_int_32 addend1,
-      png_int_32 addend2) {
-   /* Safely add three integers.  Returns 0 on success, 1 on overlow.
+static png_int_32
+png_fp_add(png_int_32 addend0, png_int_32 addend1, int *error)
+{
+   /* Safely add two fixed point values setting an error flag and returning 0.5
+    * on overflow.
     * IMPLEMENTATION NOTE: ANSI requires signed overflow not to occur, therefore
     * relying on addition of two positive values producing a negative one is not
     * safe.
     */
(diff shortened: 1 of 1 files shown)


jbowler:
@ctruta: VERY important; if you check this in to "master" or "libpng18" it will completely break my attempts to fix the current accuracy problems.

ctruta:
@jbowler if you're telling me not to check it in, then ok, I won't check it in.

ctruta:
> @ctruta: VERY important; if you check this in to "master" or "libpng18" it will completely break my attempts to fix the current accuracy problems.

@jbowler I don't know if you can tell from my previous comment, but I'm confused, and even more so after I read this https://github.com/pnggroup/libpng/issues/587#issuecomment-2374587553

Did you mean to imply that I SHOULD check this in (both in libpng16 and libpng18)? Or that I SHOULD NOT check it in?

jbowler:
Libpng 16 only.

I'm working on something better for 18.  It's too large a change to 16

EDIT: that was sent from a cellphone, hence the abbreviation.

This PR needs to go in to libpng16 but should not go into libpng18.

ctruta (approved):

ctruta:
Integrated in branch `libpng16` only.

Issue #601, opened 2024-09-22
Borland C compiler doesn't like `randomize` in pngvalid.c


ctruta:
Although I used to be a Borland compiler user ever since Turbo Pascal 5.5 on DOS, I haven't given a spin to their IDEs for many years recently. But earlier today I did just that. I installed C++Builder Community Edition ("turbo-powered" by Clang) and I noticed that `pngvalid.c` doesn't compile because `random()` is a macro defined in their `<stdlib.h>`. I suppose they're keeping it in for backwards compatibility with their old compilers and libraries, and it only gets out of the way on 64-bit builds and/or when `__STDC__` is on.

    #ifdef blah...
    #define randomize() srand((unsigned)time(NULL))
    #endif

This breakage only happens in C mode, because in C++ mode it becomes a nice inline function tucked away in namespace `std`.

So, @jbowler, when you have a chance, could you please update pngvalid.c to your taste, renaming this
https://github.com/pnggroup/libpng/blob/626761b4f5e81bb149d67dbd3913bed446a51e08/contrib/libtests/pngvalid.c#L307
to something else that's more out of the way; say (just a suggestion)

    static void
    init_random(void *pv, size_t size) 

and then

    #define RANDOMIZE(object) init_random(&(object), sizeof (object))

If you're against making a change like this, another alternative would be to enforce compilation in Standard C mode, at least for the Borland compilers (if not more).

jbowler:
Yes, random is expected to be defined in <stdlib.h> because I have _DEFAULT_SOURCE and _BSD_SOURCE defined at the top, but the code doesn't use it.  random() and srandom() and other things are defined in XOpen, SVID and BSD but I can't see anywhere where pngvalid attempts to use them and I don't see what it has to do with the function "randomize".

pngvalid is C99 because of its use of floating point.

jbowler:
Oh... ok; they're defining "randomize".  Yes, compile with --std=c89 (or c90) will fix it and is required in this case.  I think that's good practice anyway; we don't support any pre-ANSI compiler and it would seem that Borland as you are running it is pre-ANSI,

In fact pngvalid.c requires C99 so it should probably be checking __STDC__ anyway.  ISO don't define the feature test macro I'm using but they do define __STDC__.

jbowler:
PR #603 you need PR #602 to test.

ctruta:
[...]

PR #603, opened 2024-09-22
[libpng18] [build system] Require ISO-C when building libpng

jbowler:
Previously the build systems would accept any old compiler and those
compilers would define any old symbol or macro name.  ISO (well, ANSI)
defined a strict set of symbols/macros which may be defined by the
implementation leaving the remainder for application programs.

Adding a requirement for an ISO-C compiler (any version) ensures that a
check for name clashes on one compiler (with a sufficient level of
diagnostics) works for every ISO-C compiler.  In other words if it
builds on one it builds on all.

The check is in pngpriv.h; the same restrictions do not apply to code
that calls the public interface of libpng, only to the compiler used to
build ligpng.

### DIFF
--- a/pngpriv.h
+++ b/pngpriv.h
@@ -22,6 +22,16 @@
 #ifndef PNGPRIV_H
 #define PNGPRIV_H
 
+/* Compiler functionality.  For building libpng (though not necessarily for
+ * using libpng) a C compiler conformant to at least the original ANSI-C
+ * standard (C89, ISO-C90) is required.
+ */
+#if __STDC__ != 1
+#  error A compiler compliant with ISO-C90 is required to build libpng
+#endif
+
+/* TODO: check __STDC_VERSION__ (defined in ISO-C95) */
+
 /* Feature Test Macros.  The following are defined here to ensure that correctly
  * implemented libraries reveal the APIs libpng needs to build and hide those
  * that are not needed and potentially damaging to the compilation.


ProgramMax (approved):
LGTM.

I agree with the intention.
Unless there is some important government contract or similar obligating pre-ISO-C support, I don't think the cost of maintaining this support is worth it.

jbowler:
> Unless there is some important government contract or similar obligating pre-ISO-C support

There isn't.  It was always available as a "suck it and see" option.  Traditionally the response to, "I sucked and got sick" was that the maintainer would either attempt to fix it or say, "You need to use an ANSI-C compiler."

My patch removes the privilege of appealing to the maintainer.  Compiles fail if the compiler fails to define `__STDC__`  I feel it's time.  In fact I feel it is so long overdue that I'm suggesting to @ctruta, @bobfriesenhahn and everyone, well anyone, currently involved that we move to something that _at the very least_ supports inline functions.  How I hate #define :-)

pngpriv.h
```diff
@@ -22,6 +22,16 @@
 #ifndef PNGPRIV_H
 #define PNGPRIV_H
 
+/* Compiler functionality.  For building libpng (though not necessarily for
+ * using libpng) a C compiler conformant to at least the original ANSI-C
+ * standard (C89, ISO-C90) is required.
+ */
+#if __STDC__ != 1
+#  error A compiler compliant with ISO-C90 is required to build libpng
```
ctruta, line 30: One thumb up on the principle that we're straight-up requiring a modern C compiler 👍

...and...

One thumb down on the technicality that neither `__STDC__` nor `__STDC_VERSION__` are mandatory for a standard-conforming C compiler (at least in my interpretation of their meaning), and that, indeed, some compilers that do support C99 and newer (e.g. Microsoft C/C++ and Borland C/C++) do not define `__STDC__` 👎

I might be persuaded to accept code that checks both `__STDC__` and `__STDC_VERSION__`, and aborts the compilation if both of these symbols are missing; but even that is debatable.

The good news is that we don't need this change: we don't accept code that caters to the pre-ANSI C compilers, and that is that. In any case, libpng16 is already pretty much cleared of pre-ANSI C constructs.

The bad news is that I need to reopen #601; but the good news to the bad news is that the fix for #601 is rather trivial.

ctruta:
One more thing because I close this PR as WONTFIX:

ctruta:
One more thing before I close this PR as WONTFIX:

I'll be happy to accept, with both of my thumbs up and both of my palms open, any code contribution that consists in removing pre-ANSI C "compatibility" noise (of whose existence I might be unaware), in libpng16 (conditionally, if it doesn't break the ABI) and in libpng18 (unconditionally).

PR #604, opened 2024-09-23
[all branches] BUILD BREAK: eliminate use of non-ISO-C90 features

jbowler:
The MIPS MSA code contains // comments and the use of an "asm"
directive, neither of which are part of ISO-C90.  This removes the //
comments and converts asm to __asm__, which GCC allows.

The code compiles but maintenance is required; it's not clear it will
work on anything other than one specific compiler/isa combination.  It
should be rewritten using intrinsics, not assembler; as it stands it is
a security risk.

### DIFF
--- a/mips/filter_msa_intrinsics.c
+++ b/mips/filter_msa_intrinsics.c
@@ -47,7 +47,7 @@
        uint8_t *psrc_lw_m = (uint8_t *) (psrc);  \
        uint32_t val_m;                           \
                                                  \
-       asm volatile (                            \
+       __asm__ volatile (                            \
            "lw  %[val_m],  %[psrc_lw_m]  \n\t"   \
                                                  \
            : [val_m] "=r" (val_m)                \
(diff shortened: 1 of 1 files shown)


jbowler:
Should apply unmodified to libpng16 as well.

ctruta:
Roger that, thank you. FYI, today things didn't work out for me as planned, but the outlook for tomorrow is better.

TO BE CONTINUED.

ProgramMax (approved):
I think LGTM.

I do not have a MIPS machine to confirm this on. But I'm sure GCC on MIPS behaves similarly to GCC on x86 & ARM. [And the documentation is clear](https://gcc.gnu.org/onlinedocs/gcc/extensions-to-the-c-language-family/how-to-use-inline-assembly-language-in-c-code.html):

"For the C language, the asm keyword is a GNU extension. When writing C code that can be compiled with -ansi and the -std options that select C dialects without GNU extensions, use __asm__ instead of asm"

ctruta:
Integrated in both libpng16 and libpng18. Thank you.

Issue #605, opened 2024-09-23
BUILD BREAK: arm/palette_neon_intrinsics.c fails to compile with -Wcast-align


jbowler:
This is a clear problem because the code is apparently taking advantage of spurious memory alignment; this may change on some systems and may potentially crash on some systems.

The code in question added a png_bytep called riffled_palette to pngstruct.h which is, in fact, not used as a png_bytep rather it is a png_uint_32p.

I think the fix is just to make the declaration of riffled_palette png_uint_32p but doing this requires significant changes to the supplied code (i.e. a quick hack may or may not be correct) so requires the maintainer.

It also seems to be the case that the code does not require any more than byte alignment, so maybe the fix is to just declare the riffled_palette a png_voidp.

@richard-townsend-arm as the maintainer, please comment.
@ctruta please also check #266 for background, and your comments at the end.  Perhaps this is something that can be undealt with in 1.8?

ctruta:
With the caveat that I'm far from being a SIMD expert, and under the assumption that `-Wcast-align` is a legitimate warning and not a false positive, I concur with the idea of changing the type of `riffled_palette` from `png_bytep` to `png_uint_32p`.

> @ctruta please also check https://github.com/pnggroup/libpng/issues/266 for background, and your comments at the end. Perhaps this is something that can be undealt with in 1.8?

So (I said) I pushed a solution that "works for me", and that may still be true. But then, I doubt that changing the pointer type from `png_bytep` to `png_voidp` will actually solve the underlying problem -- it will merely make the warning go away.

But then, again, I'm absolutely not a SIMD expert. @richard-townsend-arm help please 😄

jbowler:
The alignment warning is both legitimate and a false positive; the pointer was allocated by malloc(1024) but the compiler cannot know that.  The problem is that -Wcast-align is one of the warnings that have been generally supported (see scripts/makefile.std) and if I enable it this, and only this, piece of code breaks the build on ARM: MIPS is a different situation.

It _can_ be fixed by "png_aligncast", which is the way it should have been written in the first place. 

I suggested (void*) both because it makes the code non-ARM-specific and because it seemed that none of the pointers have an alignment requirement.  I know enough about SIMD to see that the pointers aren't aligned in some cases and I tested the case (row pointer misalignment) that was easy to test and it seemed to work.

[...]

PR #607, opened 2024-09-26
[libpng18] BUILD BREAK: quick fix for -Wcast-align in arm/palette_neon_intrinsics.c

jbowler:
QD fix: use png_aligncastconst for the upcast in NEON palette
    
Fix (work round) for #605

This changes the arm/palette_neon_intrinsics.c code from using a bare
cast to using the portable macro png_aligncastconst.  This removes the
-Wcast-align warning.
    
The cast is safe because the pointer was allocated by malloc(1024) and
malloc is required to return a pointer with sufficient alignment for
anything the allocation can accomodate.
    
Nevertheless this is a quick and dirty fix; the use of png_bytep in the
png_struct declaration of the pointer is entirely machine specific,
other architectures may need different types and consequently the only
type which is portable to other companies CPUs is (void*).
    
### DIFF
--- a/arm/palette_neon_intrinsics.c
+++ b/arm/palette_neon_intrinsics.c
@@ -64,7 +64,7 @@ png_do_expand_palette_rgba8_neon(png_structrp png_ptr, png_row_infop row_info,
 {
    png_uint_32 row_width = row_info->width;
    const png_uint_32 *riffled_palette =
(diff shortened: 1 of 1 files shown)


ctruta (approved):

ctruta:
Integrated in both `libpng16` and `libpng18`.

PR #608, opened 2024-09-28
png_get_PLTE(): Check that num_palette is valid

caramelli:


jbowler:
@ctruta: the logic is not documented by I suspect it is that **critical** chunks have to be read completely whereas polite chunks can pass NULL pointers.

It might be better to just require real pointer everywhere.  It's a quite API change but probably a good one; dropping some information from an ancillary chunk might work but there again it implies excessive laziness.

For PLTE, nope: if an app doesn't want to know how many entries there are in the palette it deserves to crash.

@ctruta: won't fix

jbowler:
Oops:

SECURITY: rejected

num_palette may not be NULL because the other argument receives a pointer into the libpng internal copy which may only have num_palette valid entries.

The suggested change stops libpng from causing a crash but results in the app accessing an uninitialized variable (in the case where num_palette is NULL and palettep is not).  This is a major security issue; indeed the change is a potential hack.

caramelli:
Thank you for your explanations, this change was only proposed to avoid a segmentation fault with the following code example:

	#include <png.h>
	int main(int argc, char *argv[])
	{
	  if (argc != 2) {
	    printf ("Usage: %s <image.png>\n", argv[0]);
	    return 1;
	  }
	  FILE *file = fopen ( argv[1], "r" );
	  if (!file) {
	    printf ("fopen failed\n");
	    return 1;
	  }
	  png_structp png_ptr = png_create_read_struct ( PNG_LIBPNG_VER_STRING, NULL, NULL, NULL );
	  png_infop info_ptr  = png_create_info_struct ( png_ptr );
	  png_init_io ( png_ptr, file );
	  png_read_info ( png_ptr, info_ptr );
	  png_colorp palette;
	  if (png_get_PLTE ( png_ptr, info_ptr, &palette, NULL )) {
	    printf ("PLTE chunk present\n");
	  }
	  else {
	    printf ("PLTE chunk missing\n");
	  }
	  png_destroy_read_struct ( &png_ptr, &info_ptr, NULL );
	  fclose(file);
	  return 0;
	}

	$ gcc PLTE_test.c -o PLTE_test `pkg-config --cflags --libs libpng`
	$ ./PLTE_test image_with_PLTE.png
	Segmentation fault

But I'm not an expert on libpng internals, so my suggestion may indeed not make sense.

ctruta:
I agree with @jbowler that this is not a security risk, and I am rejecting this contribution, at least for libpng 1.6. However, I can understand the fix provided by @caramelli, and I am not rejecting his idea altogether, at least not in principle, for libpng version 1.8 or greater.

TL/DR: The libpng API, at least in version 1.6, is rather messy, but it ain't actually broken, so we shouldn't fix it. We may or may not fix it (or make it worse) in the subsequent versions.

### Details, details...

One of the inconsistencies of the present-day libpng API is in the semantics that's mostly (but not entirely) based on the principle of "graceful degradation" if the callee doesn't pass valid pointer arguments. Speaking of "graceful degradation", this is not idiomatic in the world of C programming in general (e.g. try passing NULL to `strlen` and see how it goes kaboom), and that's due performance reasons. However, there are exceptions, too, and the libpng API is one of those. The libpng API is awfully kind to you if you don't pass on a valid `png_ptr` or `info_ptr`, although, in my personal opinion, it shouldn't be.

Unfortunately, this "kindness" is not designed consistently. For example:

 1. In function `png_get_IHDR`, if you don't need to know the height and the width, you pass on NULL for the `height` and `width` arguments. The function `png_get_IHDR` does give you _only_ what you're asking for, and it returns 1 if the image does have an IHDR, regardless of what you asked for.
 2. In function `png_get_oFFs`, if you don't need to know some of the `oFFs` components (e.g. if you only want to know the unit type, but not the actual X and Y values) you must ask for them anyway. You cannot pass NULL for some and not pass NULL for others, because it's not going to work. Moreover, if you just need to know that you have the `oFFs` chunk but aren't interested in the actual values, you cannot find that out, because the function is going to fail (rather than telling you that you do have `PNG_INFO_oFFs`) if you miss just one of the pointer arguments.
 3. In function `png_get_pHYs`, if you don't need some of the `pHYs` components, you need not ask for them. You can pass NULL for some and not pass NULL for others, and you'll be given only what you asked for. On the other hand, if you need to know that you have the `pHYs` chunk but aren't interested in any of the actual values, you cannot find that out, because the function is going to fail (rather than telling you that you do have PNG_INFO_pHYs`) if you miss **all** of the pointer arguments.

In all this framework, `png_get_PLTE` is "unkind" if you dare to give it invalid pointer arguments for `palette` or for `num_palette`. So I like the idea of @caramelli, but I'm not thrilled about the fix, for the same reason why I'm not thrilled about option 2 above, or about option 3 above. If anything, it should behave like `png_get_IHDR` in option 1 above. (Personal opinion again: all of the `png_get_XXXX` functions should behave like `png_get_IHDR` in option 1 above.) So I am open to the idea of a patch for the 'libpng18' branch, in which option 1 is implemented, although (personal opinion, yet again...) I'd much prefer to have all of libpng API to consistently follow the rest of the world of C programming, in which getters and setters of various objects have a strict expectation about those objects to be always valid.

PR #609, opened 2024-10-01
[libpng18] [bug fix] Adjust colorspace calculations for ACES AP0

jbowler:
The subtracts in PNG_XYZ_from_xy are producing integer overflow with
some valid but extreme xy values.  This re-introduces the previous
checks, but with less limited bounds; sufficient to accomodate the
ACEScg end points (ACES AP1), but not for the ACES AP0 end points.
Those were not working anyway because libpng reads the cHRM parameters
as unsigned values, so they must always be at least 0.

A better solution requires recognizing reasonable negative values (ones
which violate the current spec) and allowing them too, at least on read.

Reviewed-by: Cosmin Truta <ctruta@gmail.com>
Signed-off-by: Cosmin Truta <ctruta@gmail.com>

### DIFF
--- a/png.c
+++ b/png.c
@@ -1180,22 +1180,66 @@ png_colorspace_sync(png_const_structrp png_ptr, png_inforp info_ptr)
 #endif /* GAMMA */
 
 #ifdef PNG_COLORSPACE_SUPPORTED
-static int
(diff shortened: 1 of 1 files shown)


jbowler:
Since I've dropped trying to fix the math problems in general there is no merit to fixing them here so this change should now go in to libpng18.

PR #610, opened 2024-10-03
[libpng18] [code standardisation] Use standard C attributes

jbowler:
ISO have recently standardized a way of defining attributes based on the
ISO-C++ syntax standardised in 2011.  This allows removal of the
existing compiler-specific hacks as well as compatibility with C++.  The
latter is important for attributes used in the header files which
declare the libpng API (png.h and descendants.)

This also extends the attribute handling to any compliant C compiler,
not just the three with compiler specific support in the prior code.

The fix also enables removal of the highly idiosyncratic use of specific
comments to indicate "fall through" in case statements; the concept is
now supported by the [[fallthrough]] attribute.

The changes include checks to ensure that the attribute syntax is
supported by the compiler and that the specific attributes are also
supported.

The changes are detailed in these commits:

commit 51b6189d3c21711bbd95538883c0aca3bb61dc67
commit 8942be90442e7a59260b01e49cdb4a3bc50c80b8

### DIFF
--- a/pngwrite.c
+++ b/pngwrite.c
@@ -1043,7 +1043,6 @@ png_set_filter(png_structrp png_ptr, int method, int filters)
          case 6:
          case 7: png_app_error(png_ptr, "Unknown row filter for method 0");
 #endif /* WRITE_FILTER */
-            PNG_FALLTHROUGH; /* FALLTHROUGH */
          case PNG_FILTER_VALUE_NONE:
             png_ptr->do_filter = PNG_FILTER_NONE; break;
 
@@ -1887,7 +1886,6 @@ png_image_set_PLTE(png_image_write_control *display)
                tRNS[i] = entry[afirst ? 0 : 3];
                if (tRNS[i] < 255)
                   num_trans = i+1;
-               PNG_FALLTHROUGH; /* FALLTHROUGH */
             case 3:
                palette[i].blue = entry[afirst + (2 ^ bgr)];
                palette[i].green = entry[afirst + 1];
@@ -1898,7 +1896,6 @@ png_image_set_PLTE(png_image_write_control *display)
                tRNS[i] = entry[1 ^ afirst];
                if (tRNS[i] < 255)
                   num_trans = i+1;
-               PNG_FALLTHROUGH; /* FALLTHROUGH */
             case 1:
                palette[i].blue = palette[i].red = palette[i].green =
                   entry[afirst];
--- a/pngwrite.c
+++ b/pngwrite.c
@@ -1043,7 +1043,7 @@ png_set_filter(png_structrp png_ptr, int method, int filters)
          case 6:
          case 7: png_app_error(png_ptr, "Unknown row filter for method 0");
 #endif /* WRITE_FILTER */
-            /* FALLTHROUGH */
+            PNG_FALLTHROUGH; /* FALLTHROUGH */
          case PNG_FILTER_VALUE_NONE:
             png_ptr->do_filter = PNG_FILTER_NONE; break;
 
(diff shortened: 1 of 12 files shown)


jbowler:
@ctruta a ping on this PR, it changes a number of files because ISO standardised a placement for the attributes which GCC does not accept as conforming to where GCC/MSVC attributes have to be place.  I.e. GCC wants them in a different place.  Consequently there's the potential for merge conflicts here if the change remains hanging.

ctruta:
Acknowledged, thank you. Please give me another day or two, because I want to integrate my own changes to the CI scripts (so that we can have verification both from libpng16 and libpng18) and to the CMake file (so that I can expand my testing, including the cross-platform testing). I plan on releasing libpng-1.6.45 with the most recent changes, BTW.

But I won't be changing any `png*.[ch]` source code, so your PRs should apply cleanly afterwards, no further rebasing necessary.

pngwrite.c
```diff
@@ -1898,7 +1898,7 @@ png_image_set_PLTE(png_image_write_control *display)
                tRNS[i] = entry[1 ^ afirst];
                if (tRNS[i] < 255)
                   num_trans = i+1;
-               /* FALLTHROUGH */
+               PNG_FALLTHROUGH; /* FALLTHROUGH */
```
ctruta: I still haven't understood the need for this goofy
```
   PNG_FALLTHROUGH; /* FALLTHROUGH */
```
What specific compilers need this?

Among the many things that I'd like to change in our coding practice is to prioritize the cleanliness of the code, in front of the cleanliness of the build. We write standard-compliant code, and we spend a reasonable effort to make the build warning-free, but sucking up to compilers that go overboard with warnings that are of no benefit whatsoever to a human reader -- I'd prefer to keep the code clean instead.

jbowler: It's a standard warning with **GCC**.  IRC **MSVC** used to do it too, maybe still does, and ISO-C has now standardised it despite the fact that it can only ever be a diagnostic.  It's still part of ISO-C to not require a `break` at the end of every `case` and it's actually very useful even though some of the code-correctness crowd don't like it because it is a `goto`.  Well, every `switch()` is a bunch of `goto`s and `switch` statements aren't _required_ anywhere; a lot of the language is fluff.

I don't like it because I don't write `C` code that way.  I do, in fact, use labels and explicit `goto` in `switch` statements and, as you say, with that code style it clutters up the text.  I don't set the warning in my own software.  Still it was Glenn's mandatory style because, well, there never was any explicit documentation...

The GCC run is **-Wimplificit-fallthrough** and it takes some options.  MSCV support is discussed in the link below but as I said I thought it was always there; I thought I had to do it in the previous millenium!

https://developercommunity.visualstudio.com/t/issue-compiler-warning-when-using-implicit-fallthr/423975

The GCC behaviour is turned on by **-Wextra** but it otherwise off by default.

I'm happy to remove it throughout but it is in **-Wextra** so I'm sure there will be lots of complaints.  If you want to do this please document it clearly and `won't fix` the original issue.

jbowler: I checked `clang` and it is off-by-default with **-Wextra**.  The issue was, I believe, explicitly about `clang` because it doesn't recognise the comment.

jbowler: @ctruta: I posted that change as a !fixup so you can test it.

ctruta: Kthx, I'll give it a try.

ctruta:
> @ctruta a ping on this PR [...] there's the potential for merge conflicts here if the change remains hanging.

I kept thinking about this PR, as well as #572. I want to reiterate that I do want to take this stuff in, but I'd like to refrigerate it for now.

I said before that the order in which I'm applying external contributions (as well as my own changes) is somewhat chaotic, but there is at least ***some*** logic to it. We need this commit, but we need house cleaning also, and I'd like to prioritize the house cleaning first. (Better develop on the smaller and cleaner codebase.) Another thing that I'd like to have is a smaller divergence between the branches `libpng16` and `libpng18`, at least for the beginning of the life of `libpng18`, because the upcoming fixes for `libpng18` that will ***need*** to be applied to `libpng16` will be that much easier to apply and cleaner also.

So, after I publish libpng-1.6.45, I'd like to resync the two branches, and then (when possible) to use `git merge COMMIT` instead of `git cherry-pick COMMIT`, just to keep the two branches closer to one another. Don't fret too much about merging the conflicts, for now, @jbowler. After we're (mostly) done with the above-mentioned priorities, we'll come back to this one.

PR #614, opened 2024-10-09
[libpng18] Update compile time support for target-specific code

jbowler:
These changes are the body of the previously discussed "target specific code"
cleanup.

There are several related elements:

1) Move (copy'n'paste) all the code including compile time checks for specific
   targets into the existing target-specific sub-directories (intel, arm etc)
   so that each sub-directory contains just the code for that specific target.
   This involves creating a new file "check.h" in each target-specific
   directory which is used to enumerate what the target code supports (by
   setting appropriate #defines).

2) Add two new files, pngtarget.h and pngsimd.c (following @ctruta's suggestion
   for the name of the latter) which include the code from the target
   directories.  pngtarget.h includes the "check.h" files and builds up
   information for use in the main build.  pngsimd.c then uses the information
   from pngtarget.h to include the appropriate implementation code.

   Neither pngtarget.h nor pngsimd.c are themselves machine specific.

3) Make the ARM-NEON target specific code which was implemented using #defines
   and extra structure members scattered through the main code independent of
   a specific target optimization and isolate it so that, like the filter code,
   it can be implemented for any target, not just ARM.

The result is considerable simplification of the main libpng code and
significant simplification of the target-specific code (compare arm/check.h
with the original code in pngpriv.h and arm/arm_init.c for example.)

- **Full compile time support for target-specific code**
- **build-sys: test and fix PowerPC VSX target-specific optimizations**
- **MIPS MSA: changes to get it compiling**


jbowler:
@ctruta  this should be good to go.

pngsimd.c
```diff
@@ -0,0 +1,141 @@
...
+ * Copyright (c) 2018-2024 Cosmin Truta
+ * Copyright (c) 1998-2002,2004,2006-2018 Glenn Randers-Pehrson
+ * Copyright (c) 1996-1997 Andreas Dilger
+ * Copyright (c) 1995-1996 Guy Eric Schalnat, Group 42, Inc.
+ *
+ * This code is released under the libpng license.
+ * For conditions of distribution and use, see the disclaimer
+ * and license in png.h
+ *
+ * NOTE: this code is copied from libpng1.6 pngpriv.h.
```
ctruta: Since we're standardized on the Git workflow, we should no longer need to add extra noise (like the note above) in the source code. Just add the standard comment boilerplate, then get down to business.

ctruta: More on this: I would also like, as part of the major libpng-1.8 cleanup, to remove all kinds of historical notes like _"added in libpng-1.2.34 blah blah blah"_ that belong to (and exist in) the Git history. The libpng codebase is already hard to read as it is, at least in my opinion, so, the more distractions we remove, the easier becomes the job of any poor soul who needs to read and understand this code from scratch.

pngsimd.c
```diff
@@ -0,0 +1,141 @@
...
+    defined(png_target_do_expand_palette_impl)
+#  error HARDWARE: png_target_do_expand_palette_impl unexpected setting
+#endif
+
+void
+png_target_init(png_structrp pp)
+{
+   /* Initialize png_struct::target_state if required. */
+#  ifdef png_target_init_filter_functions_impl
+#     define F png_target_filters
```
ctruta: Single-letter macros like `F` and `P` make the code unnecessarily brittle. You never know who uses what system library that does what kinds of shennanigans. (Trust me, I have stories to tell.)

Let's play it safe, like we always did, by using safe `PNG_ALL_CAPS_MACRO_NAMES`.

For `F`, I would suggest `PNG_DO_TARGET_FILTER_IMPL`.
For `P`, I would suggest `PNG_DO_TARGET_EXPAND_PALETTE_IMPL`.
If these names are too goofy for you, feel free to choose others.

pngstruct.h
```diff
@@ -468,5 +460,19 @@ struct png_struct_def
    png_colorspace   colorspace;
 #endif
 #endif
+
+/* New member added in libpng-1.6.36 */
```
ctruta: I realize this is copy-pasta from the old code, and I'm ok with leaving it in here for now.
But just so you know, massive comment cleanup incoming (later, later...).

pngtest.c
```diff
@@ -36,6 +36,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#define STDERR stdout
```
ctruta, line 39: Suggestion, maybe or maybe not for this commit:
This "STDERR is stdout" was Glenn's personal hack for his personal testing, which I haven't removed from libpng-1.6, but we should have no need for it, going forward.

pngsimd.c
```diff
@@ -0,0 +1,141 @@
+/* pngsimd.c - hardware (cpu/arch) specific code
```
ctruta, line 2: Nit: as you can see in all the other `png*.c` source files, the first like should be empty.
(Unless you want to change them all, wholesale, in a different commit, to which I would not disagree.)

pngtarget.h
```diff
@@ -0,0 +1,129 @@
+/* pngtarget.h - target configuration file for libpng
```
ctruta, line 2: Nit: as you can see in all the other `png*.h` header files, the first like should be empty.
(Unless you want to change them all, wholesale, in a different commit, to which I would not disagree.)

pngtarget.h
```diff
@@ -0,0 +1,129 @@
...
+ * Where ARCH the architecture directory (the directory containing check.h) in
+ * upper case.  See pngsimd.c for more information about function definitions
+ * used to implement the code.
+ */
+#ifndef PNGTARGET_H
+#define PNGTARGET_H
+
+#ifdef PNG_TARGET_SPECIFIC_CODE_SUPPORTED /* from pnglibconf.h */
+#  ifdef PNG_READ_SUPPORTED /* checked here as a convenience */
+#     include "mips/check.h"
```
ctruta, line 85: Nit: If you could please sort these `#include` lines alphabetically.

scripts/pnglibconf.h.prebuilt
```diff
@@ -216,7 +207,7 @@
 #define PNG_USER_HEIGHT_MAX 1000000
 #define PNG_USER_WIDTH_MAX 1000000
 #define PNG_ZBUF_SIZE 8192
-#define PNG_ZLIB_VERNUM 0 /* unknown */
+#define PNG_ZLIB_VERNUM 0
```
ctruta: Suggestion (not for this commit, just FYI): we should remove `PNG_ZLIB_VERNUM` altogether from libpng-1.8, and simply require zlib version 1.2.11 or later, going forward. (I mean 1.2.11, not 1.2.9 or 1.2.10, because those two are known to be broken.)

pngsimd.c
```diff
@@ -0,0 +1,141 @@
...
+ */
+
+/* This will fail in an obvious way with a meaningful error message if the file
+ * does not exist:
+ */
+#include PNG_TARGET_CODE_IMPLEMENTATION
+
+#ifndef png_target_impl
+#  error HARDWARE: PNG_TARGET_CODE_IMPLEMENTATION is defined but\
+ png_hareware_impl is not
```
ctruta, line 62: What's this "hareware" thingy? 🐰 🙃

ctruta (changes requested):
The code looks good. I don't have actual review objections, only nitpicking.
After you address your comments, feel free to add the "Reviewed-by:" line before the "Signed-off-by:" line.

About nitpicking: for the first line of the commit message, I would prefer a full sentence in present tense. For example: _"Add full compile-time support for architecture-specific code"_
And for the last line in the commit message: that "see pngfoo.h and pngblah.c" line might be appreciated by Captain Obvious, but (in my professional experience) short and sweet commit message that go straight to the point about what exactly has been changed (and also explain why if necessary) are more effective, especially when a reader navigates over a long and hairy Git history.

ctruta (approved):
I approve this commit, but I would prefer the label of arch-specific commits to follow the current practice of naming the actual arch (for now -- until we change this practice wholesale).

I would suggest a one-liner like the following:
`powerpc: Fix a leaky function declaration`

BTW, if it changes the API or the functionality in any way, it is no longer a "chore".

Another thing that needs not go in commit messages in general -- explanations about the testing process -- unless the commit itself is about the testing process. (I'm doing that myself, sometimes, in my own CI commits, so I'm partly to blame here.)

mips/msacheck.h
```diff
@@ -10,7 +10,7 @@
  * This code has been moved from the original in pngpriv.h.
  */
 /* MIPS MSA checks: */
-#if defined(__mips_msa) && (__mips_isa_rev >= 5)
+#if defined(__mips_msa) && (__mips_isa_rev >= 2)
```
ctruta: I neither agree nor disagree with this line, but I do think it's important enough to warrant its own separate commit. I would prefer to not have it buried in some innocent-looking commit named "fix compiler warnings".

But on the other hand, I wouldn't mind if you do focus on this line; e.g., name the commit something like
```
mips: Lower the required ISA level from 5 to 2
```
and then, squeeze in the other changes in this one commit, and then explain, e.g., "also fix unrelated compiler warnings blah blah blah".

ctruta (changes requested):
For the commit message, I have the same comment as before: use a "semi-standardized" label for the subject line, like this:
```
mips: Fix specific gcc/clang warnings
```
and then explain, in the subsequent line, that you wanted to pacify `-Wshadow`, `-Wcast-align` and `-Wunused`.

*****

Another nitpicking, if you please don't mind:
unless what you're fixing is an actual compiler error issued by an actual C language standard violation, you shouldn't say that you're fixing the compilation, but rather, that you're pacifying a compiler warnings. Those who use `-Werror` will know what you mean, while those who don't will not be bothered.
Exception: whenever we want to fix something that breaks the CI verification (as opposed to an actual compiler error) the formal incantation is "fix verification". But it has to be publicly verifiable, not just an internal setup.
(FYI: in my own workflow, I'm running extended compilations with extended compiler warnings enabled as well.)

pngsimd.c
jbowler: Done

pngsimd.c
jbowler: Done, made it far more explicit

pngtest.c
jbowler: I couldn't get minconfig builds to work without the change, considerably limiting my testing.  There's a separate PR for it.

pngtarget.h
jbowler: Changed to have spurious blank lines in both files.

scripts/pnglibconf.h.prebuilt
jbowler: That was an error; I forgot to add the "unknown" comment back in when I changed the VERNUM to 0.

I rely on PNG_ZLIB_VERNUM when reviewing bug reports; that's why I ask people to include pnglibconf.h, as in the recent fuzzer report where I initially suspected a duplicate of the several-times-reported fuzzer misconfiguration WRT zlib.

mips/msacheck.h
jbowler: I made that change but it means I can't test the MSA build any more; my compiler is built with the lower ISA rev.  I'll enter a separate issue or pull request as appropriate.

pngstruct.h
jbowler: Deleted

pngtarget.h
jbowler: Done

pngsimd.c
jbowler: Original names (like the erroneous 1.6.36 comment above).  Changed to TARGET SPECIFIC CODE in all places.

jbowler:
I assume you will do a squash commit here.  So I just changed the subject line of the messages.  I've never edited a commit message; I just do squash merges...

jbowler:
Ok.  I am unable to resolve the last "Changes requested" so I've incorporated all your review comments into #615 and am closing this.

PR #615, opened 2024-10-09
This change updates support for target-specific code.

jbowler:
The change removes the need for build configuration of target specific
code such as SIMD enhancements.

The extensible framework is based on checks in the new file named
pngtarget.h and allows per target description of requirements and
capabilities using macros with the prefix PNG_TARGET_.  These macros are
interpreted by the core libpng code in a target independent way.

The PNG_TARGET_ prefixed macros describe the target specific
implementation and document how to include it.  The actual inclusion of
the code is done, at present, by a single target independent core file
named pngsimd.c.

This approach ensures that no special configuration in the build system
is required on any system.  It is not required on a system which is
compiling for the build host where we can reasonably assume that the
compiler is correctly set up.  Neither is it required for cross builds
so long as the toolchain is correctly set up.  In this case the
toolchain means the cross build system and any ancilliary programs it
requires.

More details of the implementation are included in the source files
named pngtarget.h and pngsimd.c.  These files document in C comments how
additional target specific code might be added.

The changes consist primarily of movement of target specific code from
the core libpng source files to the existing target specific
subdirectories such as the ones named intel and arm.  This isolates the
code into the corresponding subdirectory and eliminates the requirement
for changes across the libpng structure.

As part of this the ARM processor NEON instruction set specific palette
to RGB code has been generalised to allow implementation in other
architectures.  The resultant updated version of the original (ARM Ltd)
implementation has been moved to the subdirectory named arm.

In addition some selected changes have been included to allow testing
with higher warning levels than the typical compiler defaults.


jbowler:
As best I understand them this PR incorporates all your required changes.  It doesn't change the actual code, just the comments, commit messages and so on.

jbowler:
In fact this is incomplete; the Loongson support isn't there and I can't do that.  It needs a non-US person.

hardware.old/check.h
ctruta, line 1: I think you meant to delete this, no?

jbowler: Rather I didn't mean to add it...   "git add */check.h" and hardware.old was a backup just incase things got damaged.

ctruta: No worries, I got it.

mips/mips_init.c
```diff
@@ -1,187 +1,21 @@
 
 /* mips_init.c - MSA optimised filter functions
  *
- * Copyright (c) 2018-2024 Cosmin Truta
+ * Copyright (c) 2018 Cosmin Truta
  * Copyright (c) 2016 Glenn Randers-Pehrson
- * Written by Mandar Sahastrabuddhe, 2016
- * Updated by guxiwei, 2023
```
ctruta, line 7: Just nitpicking: was the modification of the copyright notice deliberate?

jbowler: I think I the first line is a mistake.  I thought I had rebuilt it from the original 2016 version but I did it using git restore so maybe Mandar's authorship line wasn't there, or maybe I messed it up.

hardware.old/check.h
jbowler: Ok; I pushed an update anyway (it just removes the erroneous file; `git clean -dfx` is a real help :-)

mips/mips_init.c
jbowler: Take a look at the actual file, not the diff.  Check it out in my branch here:

https://github.com/jbowler/libpng/blob/20241009-target-specific-code/mips/mips_init.c

You will see that the copyright line is still there but what changed is that I reverted to a version before your removed the trailing full stop.  So look immediately below this comment and you will see that line added back in.  I'll push a change with the full stop removed.

ctruta (approved):
Big honking commit approved 👍

PR #619, opened 2024-10-10
code cleanup: This change removes the blank line from the beginning of libpng source files.

jbowler:
The blank lines are apparently an artefact of an older source control
system.  They are not required and look like errors or accidents because
the format of putting a blank line at the start of files is not in
regular use by software developers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,3 @@
-
 # CMakeLists.txt - CMake lists for libpng
 #
 # Copyright (c) 2018-2024 Cosmin Truta.
--- a/arm/arm_init.c
+++ b/arm/arm_init.c
@@ -1,4 +1,3 @@
-
 /* arm_init.c - NEON optimised filter functions
  *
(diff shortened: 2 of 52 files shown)


jbowler:
They were all over the place but not consistently so.  I'm sure the leading blank line has no semantic significance in any of those files.

This shouldn't cause merge conflicts with any pending pull requests.

ctruta:
Applied, thank you. I applied it both to libpng18 and libpng16.

If you probably noticed, I take the liberty to edit the contributed commit messages, including yours. I hope you don't mind that. You mentioned the Git contribution guidelines, and while I don't follow all of them, I do follow many.
https://git-scm.com/docs/SubmittingPatches
I very much like this statement:
_Describe your changes in imperative mood, e.g. "make xyzzy do frotz" instead of "[This patch] makes xyzzy do frotz" or "[I] changed xyzzy to do frotz", as if you are giving orders to the codebase to change its behavior._

PR #622, opened 2024-10-14
dev: Make the use of stdio FILE robust.

jbowler:
ABI changes:
    The three API functions png_init_io, png_begin_read_from_stdio and
    png_image_write_to_stdio now require the host fread, fwrite and
    fflush functions where appropriate in addition to the host FILE.

    Internally libpng uses the provided functions for all operations on
    the FILE; it does not use the implementations available when libpng
    was built.

API changes:
    The original APIs of the above three functions are now implemented
    as function-like macros which automatically pass the correct ISO-C
    functions.  This ensures that there are no net API changes and that
    the correct arguments are passed.

Build changes:
    The read stdio functionality is now dependent on SEQUENTIAL_READ
    rather than READ.  This is done for clarity; the progressive read
    mechanism does not use FILE.

Internal changes:
    png_struct::io_ptr has been supplemented with png_struct::stdio_ptr
    used when stdio is used directly by libpng for IO as opposed to
    caller-provided callbacks.

    Error checking has been added to detect mismatched use of the stdio
    (png_init_io etc) API with the callback (png_set_read_fn,
    png_set_write_fn APIs.)  Changing from one API to the other
    mid-stream should work but has not been tested and is not currently
    a documented option.

The changes address an issue which is frequently encountered on
Microsoft Windows systems because Windows NT DLLs each have their own
ISO-C hosted environment.  This means that a FILE from one DLL cannot be
used safely from a different DLL unless all access to the object is done
using functionality from the creating DLL.  The problem is more general;
passing objects across DLL or other boundaries is frequently supported
but direct access to those objects' internal structure in the receiving
environment is not safe.  Other such uses were address early on in the
development of libpng, this addresses the main, almost certainly, sole
remaining issue.

The idea of adding additional function pointers png_init_io was
suggested by github.com user pp383 in pull request #208.

### DIFF
--- a/png.c
+++ b/png.c
@@ -669,7 +669,7 @@ png_get_io_ptr(png_const_structrp png_ptr)
    return png_ptr->io_ptr;
 }
 
-#if defined(PNG_READ_SUPPORTED) || defined(PNG_WRITE_SUPPORTED)
+#if defined(PNG_SEQUENTIAL_READ_SUPPORTED) || defined(PNG_WRITE_SUPPORTED)
 #  ifdef PNG_STDIO_SUPPORTED
 /* Initialize the default input/output functions for the PNG file.  If you
  * use your own read or write routines, you can call either png_set_read_fn()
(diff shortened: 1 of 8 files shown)


contrib/libtests/pngimage.c
```diff
@@ -1016,7 +1016,9 @@ compare_read(struct display *dp, int applied_transforms)
    C(height);
    C(bit_depth);
    C(color_type);
-   C(interlace_method);
+#  ifdef PNG_WRITE_INTERLACING_SUPPORTED
```
ctruta: I propose removing certain build flags that would remove core PNG functionality, including `PNG_WRITE_INTERLACING_SUPPORTED`, altogether. I know of nobody to actually use those, and, worse yet, I know of nobody to actually test those.

ctruta (commented):
I'd like to think about this particular solution for a little bit. While I do agree with the new function additions, I am not so sure that we should override their names with macros in this manner. Something like this would make perfect sense in a C++ library, where overriding function names is idiomatic, and 100% working out of the box.

But we're stuck with C in here. And, for better or for worse, we've been having the goofy (but 100% working!) naming habit in which we overloaded existing function names with number suffixes: `png_get_eXIf_1`, `png_create_write_struct_2`, etc.

I, for one, would prefer leaving the old functions in place, perhaps with a deprecated attribute (to be removed eventually in some far-away future after the users migrate), and adding the new `png_init_io_2`, `png_begin_read_from_stdio_2`, etc.

In zlib they also use this practice. See, for example, `deflateInit2`.

(Yes, these API names are ugly 😝)

contrib/libtests/pngimage.c
jbowler: Interlacing is optional on write, deinterlacing is required on read.  Everything in scripts/pnglibconf.dfa is set to conform to the spec.  It's not possible to disable the deinterlace code without disabling read.  On write interlace support is there unless explicitly disabled.

This is all correct.

ctruta: I wasn't raising an objection about correctness, but about testing. I don't mind implementing a correction like this in pngimage.c now (both in this branch and in 'libpng16'). And, BTW, if we do it, it should be in a separate commit on its own, because this particular change is not at all about the stdio robustness.

My objection about testing is this: who has ever tested `PNG_WRITE_INTERLACING_SUPPORTED`? And also, who has ever needed it? And, last but not least, how significant are the space savings?

To conclude: I was saying that pnglibconf.dfa and pnglibconf.h.prebuilt need to go on a diet. I am open to debating that, but we should do it in a separate commit, in order to not hijack the topic of this particular PR.

jbowler:
> I am not so sure that we should override their names with macros in this manner.

There is no choice.  png_init_io and the simplified APIs don't work and are unsafe and there is a lot of code that uses them.   They are good APIs even though the implementation (the ABI) is flawed.

Therefore even though it's an ABI-change release this aspect of the API should remain the same.  This is particularly true because it **must** be fread, fwrite and fflush which are passed to libpng, not something else.  The OPs fix did what you just suggested but used C99 inline functions and massive unnecessary additions to the ABI while only addressing png_init_io!

The inlines pass a function call across the boundary not the function itself but this seems to be mindbogglingly complex to me and it requires C99 while ISO-C allows functions to be passed as pointers!

The function-like macros stop existing code and new code from doing the wrong thing; they **hide** the ABI but give an unchanged API which just does the right thing.  What is more the result is much simpler and avoids adding yet more ABIs for no purpose.

The whole thing about png_function_666 is that this was some half-baked attempt to pretend that the ABI wasn't being changed because it was "only" an addition.  It's completely unnecessary to do that in an ABI-changing release and, anyway, it doesn't work because builds against the ABI-added 1.6 fail against prior DLL versions (and fail badly.)

contrib/libtests/pngimage.c
jbowler: Ok, I'll pull that part of the change.

It's like the STDERR thing in pngtest.c; I was testing this change and I discovered pngimage.c builds had become broken so I fixed it to allow my tests to complete.  The break was (is) there without the rest of the change:

`cp contrib/conftest/pngcp.dfa pngusr.dfa`

To answer your other question, with and without write interlace (pngcp built with --disable-shared):

```
   text    data     bss     dec     hex filename
  64808    2528      48   67384   10738 pngcp
  52815    1408       8   54231    d3d7 libpng18.so.18.0.git
  66952    2528      48   69528   10f98 pngcp
  55104    1416       8   56528    dcd0 libpng18.so.18.0.git
```

So that's 2144 bytes of dead code in `pngcp`, a **3.3%** increase in code size, and, with the shared library, 2289 bytes which includes, I assume, the LUT for the extra ABI and totals **4.3%** more in the text segment plus an extra DLL symbol resolve (these can be quite slow) on some systems.

jbowler:
Change made using `git commit --fixup`; `git rebase --autosquash`  is the approved merge approach.  I assume github.com does this automatically but since I've never used `--fixup` before I can't be sure.

jbowler:
The pngimage.c fix is now #623

ctruta:
> Change made using `git commit --fixup`; `git rebase --autosquash` is the approved merge approach. I assume github.com does this automatically but since I've never used `--fixup` before I can't be sure.

No problem. I'm using the command line to integrate PRs anyway.

ctruta:
I still need a little more time to review this. Will be back tomorrow.

jbowler:
> > Change made using `git commit --fixup`; `git rebase --autosquash` is the approved merge approach. I assume github.com does this automatically but since I've never used `--fixup` before I can't be sure.
> 
> No problem. I'm using the command line to integrate PRs anyway.

The now separate change to pngimage.c seems to have caused git to do an autosquash anyway, or maybe I did it (I did run --autosquash after fixing the merge conflicts.)  Anyway it's back to a single commit now.

PR #623, opened 2024-10-14
bug: pngimage.c: Disable the check on interlace_method if WRITE_INTERLACE is not supported.

jbowler:
Inside libpng if interlacing a PNG is disabled libpng silently removes
it.  This updates pngimage.c to not check the interlace in the resultant
image, other behavior is still checked.

### DIFF
--- a/contrib/libtests/pngimage.c
+++ b/contrib/libtests/pngimage.c
@@ -1016,7 +1016,12 @@ compare_read(struct display *dp, int applied_transforms)
    C(height);
    C(bit_depth);
    C(color_type);
-   C(interlace_method);
+#  ifdef PNG_WRITE_INTERLACING_SUPPORTED
+      /* If write interlace has been disabled the PNG is still written
+       * correctly but as a simple, not interlacedd, PNG.
+       */
+      C(interlace_method);
+#  endif
    C(compression_method);
    C(filter_method);
 


jbowler:
This is behaviour that dates back to the dawn of libpng.  It's happening in `pngwutil.c` in the function `png_write_IHDR`.  The line in question along with the `PNG_WRITE_INTERLACING_SUPPORTED` check seems to have been added by @glennrp on Jan 30, 1998, "Imported from libpng-0.99.tar".

Nevertheless that seems to have been a bug-fix; `WRITE_INTERLACING` was present in earlier versions, just buggy.  It is first present in libpng-0.81 but is absent in libpng-0.71   It seems to have been added by Guy.

The danger of removing it now is that it is entirely possible quite a lot of code depends on it; pngimage uses pretty much the original design for png_info handling.

ctruta (approved):
Approved, typos notwithstanding. (But I got them.)

Just so you know, I'd like to avoid using the label "bug" for something that isn't core libpng. This is a conditional compilation issue in a contributed test program, so I'll label it as test. For somebody who has a cursory look in the git history (e.g. "with `git log --oneline`), mentioning a bug fix in a test program (not in the core libpng) looks unnecessarily scary. And the name `pngimage.c` doesn't help alleviating such a concern, either.

ctruta:
Integrated both in libpng16 and libpng18.

jbowler:
Technically it is a "PNG stream", not a "PNG file" because a file is always a stream (in ISO-C) but the inverse does not apply.  Hum, in fact in the NT file system, Reiser4 and, arguably, traditional MACOS a "file" is a structured object however libpng can only write a PNG as a stream.

PR #624, opened 2024-10-14
development: Move target specific code to a single subdirectory.

jbowler:
This is a two commit move of all the current target specific code to a single
subdirectory called 'target'.

- **development: Move target specific code to a subdirectory.**
- **development: Move target specific code to a subdirectory.**


target/arm/filter_neon.S
ctruta, line 1: `filter_neon.S` was a dummy placeholder in 1.6.x, which we kept in, to avoid breaking stuff. This commit is the opportunity to delete it for good.

ctruta (changes requested):
I was looking forward to get this integrated, next thing after the release of 1.6.45 (which I had planned for early this week), but stuff came out and I couldn't publish 1.6.45 yet.

After release, I will resync libpng18 with libpng16 through a "tag merge", to reduce (at least for a little bit) the divergence between the two branches, but the code should largely stay the same.

However, before we make a change that will be then frozen for the next decade or so, let us work a little bit on the nomenclature, because I'm not so thrilled about the new directory name `target`.

The concept of "target" makes a lot of sense in CMake, and a lot less sense in Autotools and (unfortunately) in Meson which (unfortunately) followed up on Autotools' nomenclature. It makes less sense in source code. On top of that, "target" means "target platform" which means "machine" plus "operating system". Our code in question is machine-specific; or, rather, architecture-specific (or arch-specific); or, rather, SIMD-specific. Seriously, that code is about nothing else but SIMD optimizations (not about CPU stuff that is not SIMD stuff, and also, not about target platform stuff that is not CPU stuff).

MOREOVER:

As I noticed how various libpng users are vendoring our library, they tend to only include the code that is part of libpng, and not projects, and not scripts, and not contrib, etc. In order to make it easier for everybody to know **what is libpng** vs. **what is NOT libpng**, I propose using the prefix `png` for everything in the top-level directory that **is indeed libpng**.

IN CONCLUSION:

I propose that we move all of the machine-specific stuff (or arch-specific stuff, or SIMD-specific stuff) to a directory named `pngmachine` (or `pngarch`, or `pngsimd`). My personal favourites are `pngarch` and `pngsimd`, but I'm open to a debate.

AND ALSO:

I propose that we move all the pnglibconf stuff from the `scripts` directory to a more glorified `pnglibconf` directory, so that we can easily explain to everybody that all top-level `png*/` directories (i.e. (`pngsimd/` or `pngarch/` or `pngmachine/`) and (`pnglibconf/`)) should always travel with all the top-level `png*` files (i.e. `png*.c` and `png*.h`).

PR #631, opened 2025-01-02
fix: Update the cICP implementation yet more

ctruta:
For the sake of completeness:
 * Add the cICP entry to the list of known chunks to ignore inside `png_set_keep_unknown_chunks`.
 * Handle cICP in `png_read_end`, alongside cHRM, gAMA, iCCP, sRGB.
 * In pngtest.c, move the cICP test code near cHRM, gaMA, iCCP, sRGB.

### DIFF
--- a/pngread.c
+++ b/pngread.c
@@ -1,6 +1,6 @@
 /* pngread.c - read a PNG file
  *
- * Copyright (c) 2018-2024 Cosmin Truta
+ * Copyright (c) 2018-2025 Cosmin Truta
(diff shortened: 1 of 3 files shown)


jbowler (approved):
Yes, I think so.  The "ignore" list change is certainly correct unless I made a mistake with the ASCII values (perhaps PNG_A, PNG_B etc would help).  The png_read_end change seems correct although the bug is harmless.  I dislike the whole approach of having to add function calls to multiple places but that's nothing specific to this change.

PR #635, opened 2025-01-04
PNG v3: Support for mDCV and cLLI

jbowler:
This adds APIs to get/set the two remaining new PNG-v3 colour space
chunks.  The mDCV API matches that of cHRM.  Both chunks support
floating point APIs (all values in the two chunks are real numbers).

Both chunk have a new encoded type, a 4dp fixed point number, which
cannot be represented in the existing png_fixed_point type so a
png_uint_32 is used.

Test examples for cICP, cLLI and mDCV are now in 'pngtest.png' and a
necessary change to the pngunknown test has been made to accomodate the
additions.


jbowler:
This should apply ok to libpng18 too.  As the CI message says the API for mDCV matches that for cHRM.  This is essential because if a PNG with just gAMA+cHRM is converted to cICP using a different encoding, e.g. converting sRGB to REC 2020 (any variant) the cHRM chunk provides the values for mDCV.  It would be error prone, very inconvenient and totally unnecessary if apps had to convert from one chromaticity encoding to another.  Fortunately mDCV is a subset of cHRM so the use of the cHRM data types (in particular png_fixed_point) merely requires range checks in png_set_mDCV_fixed (range checks where are necessary for floating point anyway.)

Tested with the new pngtest.png which revealed that pngunknown needs to be updated for new chunks too...

ctruta (approved):

png.h
```diff
@@ -3346,7 +3346,7 @@ PNG_EXPORT(244, int, png_set_option, (png_structrp png_ptr, int option,
  * one to use is one more than this.)
  */
 #ifdef PNG_EXPORT_LAST_ORDINAL
-  PNG_EXPORT_LAST_ORDINAL(251);
+  PNG_EXPORT_LAST_ORDINAL(259);
```
ctruta, line 3349: So we'll git-merge all this into the libpng18 branch as planned, but beyond that, these ordinals gotta go.

scripts/symbols.def
```diff
@@ -255,3 +255,11 @@ EXPORTS
...
  png_get_cICP @250
  png_set_cICP @251
+ png_get_cLLI @252
+ png_get_cLLI_fixed @253
+ png_set_cLLI @254
+ png_set_cLLI_fixed @255
+ png_get_mDCV @256
+ png_get_mDCV_fixed @257
+ png_set_mDCV @258
+ png_set_mDCV_fixed @259
```
ctruta, line 265: Yeah... this `@259` and all of its friends must be no more by the time we publish libpng-1.8.0.

ctruta (commented):
Many thanks, @jbowler. But before we move on, here are my off-topic post-review comments.

png.h
jbowler:
```suggestion
  PNG_EXPORT_LAST_ORDINAL(259);
```
At least the script (only in the configure build) that checks the ordinals then faults the build unless the machine-generated files have been updated too.

1.8; a simple change that is unlikely to cause merge conflicts against 1.6 since I believe it's only a few lines.  Remove the symbol-by-ordinal support.  Throw that PoS against the wall and see what breaks.  But don't make any other change to those damned macros because if it has to be reintroduced the result would be a nightmare bug farm.

Completely off topic: the same applies to pnglibconf.h.prebuilt, but generate that in the tarball for the "approved" Linux config.  Just don't check it in autoconf or cmake builds that use it.

I think you said this years ago, but if it wasn't you it wasn't me or Glenn either yet it is still correct: The machine generated files should not be in the Git.

pngset.c
```diff
@@ -159,6 +159,158 @@ png_set_cICP(png_const_structrp png_ptr, png_inforp info_ptr,
...
+png_ITU_fixed_16(png_const_structrp png_ptr, png_fixed_point v,
+   png_const_charp text)
+{
+   /* Return a safe uint16_t value scaled according to the ITU H273 rules for
+    * 16-bit display chromaticities.  Functions like the corresponding
+    * png_fixed() internal function with regard to errors: it's an error on
+    * write, a chunk_benign_error on read: See the definition of
+    * png_chunk_report in pngpriv.h.
+    */
+   v /= 2; /* rounds to 0 in C: avoids insignificant arithmetic errors */
```
Morphinity, line 217: Hi, what is the reasoning behind scaling this value by 1/2 before returning it? 
When calling `png_set_mDCV_fixed` from `png_set_mDCV` we are already scaling the chromaticity values by 1/2 (to multiply it by 50,000 instead of 100,000 - which is what the spec requires). 
So further scaling it by 1/2 before returning from here could be causing the chromaticity values rx, ry, gx, gy, bx, by to be half of what they should be when being encoded. Please correct if I am missing something

jbowler: The type `png_fixed_point` is used in the libpng API to represent fixed point numbers.  It contains a real number scaled by 100,000.  It is therefore necessary to multiple/divided by 2 to handle the chromaticity values.  The code in `png_set_cHRM_fixed` and `png_get_cHRM_fixed` doesn't need to do this because the values are represented in PNG using the same type libpng uses (this is where the type comes from.)

You will note that the already existing APIs have a consistent order for the chromaticities: white, red, green, blue and this order is also the one used for the four `mDCV` APIs.

I suggest you use the floating point APIs; fixed point values are always tricky if you move from one well defined API environment to another.

Miloni2010: @jbowler Can you please clarify what you mean by floating-point APIs here? We are using `png_set_mDCV`, which is a floating-point API where we divide each value by 2 once. But it is calling `png_set_mDCV_fixed`, where it gets divided again in `png_ITU_fixed_16`, and the final master display chromaticity values are being populated incorrectly. Are you suggesting that we directly use `png_set_mDCV_fixed` to set mdcv values or am I misunderstanding the solution here?

Miloni2010: @jbowler Also shouldn't we not divide the red, blue, green, and white values in `png_set_mDCV` by 2 at all? I mean shouldn't this - `png_fixed(png_ptr, white_x / 2...)` be  `png_fixed(png_ptr, white_x, "png_set_mDCV(white(x))")` and so on for all values to be able to correctly scale them?

pngset.c
Miloni2010, line 1: @jbowler Can you please clarify what you mean by floating-point APIs here? We are using png_set_mDCV, which is a floating-point API where we divide each value by 2 once. But it is calling png_set_mDCV_fixed, where it gets divided again in png_ITU_fixed_16, and the final master display chromaticity values are being populated incorrectly. Previously, were you suggesting that we use png_set_mDCV_fixed to set mdcv values, or am I misunderstanding the solution?

Also shouldn't we not divide the red, blue, green, and white values in png_set_mDCV by 2 at all? I mean shouldn't this - png_fixed(png_ptr, white_x / 2...) be png_fixed(png_ptr, white_x, "png_set_mDCV(white(x))") and so on for all values to be able to scale them correctly?

pngset.c
jbowler: Thanks for reporting it as a bug - it was, indeed, a bug.  My explanation was wrong; my excuse is that this is not my real job (i.e. I have more, much more, important things to do than fix bugs in libpng, even if I write them!)

pngset.c
jbowler: Yeah; it was a bug.  It's fixed now.

jbowler: So I'm going to resolve this... But there is a real issue here.  libpng really does have **no test code** other than the code I wrote.

People have whine, whinged and generally complained about this for years but I am the only person left alive on this planet who actually ever did anything about it.  So fix it.

pngset.c
jbowler: Thought I just resolved this with a sarcastic comment... Nevermind.

scripts/symbols.def
jbowler: Didn't you fix this by deleting the PNG_PREFIX stuff?  I thought 1.8 was dead.

PR #649, opened 2025-01-30
bug fix: png_write_iCCP check on profile length

jbowler:
Previously png_write_iCCP used the length from the first four bytes of
the profile set by png_set_iCCP rather than the actual data length
recored by png_set_iCCP.  This resulted in a read-beyond-end-of-malloc
bug if the profile data was less than 4 bytes long.

### DIFF
--- a/pngwutil.c
+++ b/pngwutil.c
@@ -1150,6 +1150,9 @@ png_write_iCCP(png_structrp png_ptr, png_const_charp name,
    if (profile_len < 132)
       png_error(png_ptr, "ICC profile too short");
 
+   if (png_get_uint_32(profile) != profile_len)
+      png_error(png_ptr, "Incorrect data in iCCP");
+
    temp = (png_uint_32) (*(profile+8));
    if (temp > 3 && (profile_len & 0x03))
       png_error(png_ptr, "ICC profile length invalid (not a multiple of 4)");
--- a/pngwutil.c
+++ b/pngwutil.c
@@ -1132,10 +1132,9 @@ png_write_sRGB(png_structrp png_ptr, int srgb_intent)
 /* Write an iCCP chunk */
 void /* PRIVATE */
 png_write_iCCP(png_structrp png_ptr, png_const_charp name,
-    png_const_bytep profile)
+    png_const_bytep profile, png_uint_32 profile_len)
 {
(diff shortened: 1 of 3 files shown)


pngwutil.c
```diff
@@ -1148,8 +1147,6 @@ png_write_iCCP(png_structrp png_ptr, png_const_charp name,
    if (profile == NULL)
       png_error(png_ptr, "No profile for iCCP chunk"); /* internal error */
 
-   profile_len = png_get_uint_32(profile);
```
ctruta, line 1151: I agree with your fix @jbowler, but my I please suggest one more thing in here:

    if (png_get_uint_32(profile) != profile_len)
       png_error(png_ptr, "Incorrect data in iCCP");

ctruta (changes requested):

pngwutil.c
jbowler: Sure, but won't that produce compiler warnings on some systems because of lines 1157..1162?

jbowler: It doesn't, so I made the change.

ctruta:
I applied this fix (thanks!), but I happen to disagree with what GitHub does when I select "Squash and merge", so I'll just pretend to close the PR without merging. @jbowler when you resubmit commits in the future, following up on code review, if you could please just force-push a new amended commit, then I can avoid GitHub's rather unclean squashing-and-merging strategy.

jbowler:
Curious.  "git merge --squash" launches an edit on the accumulated commit messages; my expectation was that you could just delete the leading lines to get to the original message which did not need changing in this case.

Submitting separate commits to respond to reviews seems to be the norm; otherwise, particularly on large changes, it's more difficult to review the response changes themselves online.

ctruta:
Squashing commits in the command line, no problem.
Pushing them upstream from the command line, also no problem.

Using the GitHub button "Squash and merge" to let GitHub know that I just squashed, merged and pushed (which should have resulted in the PR being marked as "Merged") -- problem.

For example, see commit ceda44aed4a6af9cd9ec0c195a7a62424c318827, which is what GitHub created when I wanted to resolve PR #646 by using the "Squash and merge" button. I eventually force-pushed that faulty commit away into oblivion.

PR #652, opened 2025-02-02
build: Move the pnglibconf files and scripts to scripts/pnglibconf/

ctruta:
@jbowler please review.

Just so you know, after this change, the next fossil that I want to discard is the .DEF file and all the associated makefile-ing and scripting. We should neither produce it (with pnglibconf or not), nor consume it in any makefile or build file or project file any further. We should rather go back to using `__declspec(dllexport)` where applicable (i.e. on Windows), or `__attribute__((visibility("default")))` where applicable (on the Unix platforms with the compilers that allow it), which is what we used to use from libpng-1.0 to libpng-1.4, _except_ with one important difference: no more .DEF file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,7 +178,7 @@ endif()
 
 if(NOT AWK OR (ANDROID OR IOS))
   # No awk available to generate sources; use pre-built pnglibconf.h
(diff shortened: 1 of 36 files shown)


ctruta:
I forgot to mention in the commit message one more thing that I did: while editing the libpng manual, I also corrected some of the phrasing and formatting in the surrounding paragraphs. In any case, the manual needs a heavy facelift.

PR #654, opened 2025-02-11
Update man page to latest PNG 3e draft

svgeesus:
The link in the man page was out of date.

### DIFF
--- a/png.5
+++ b/png.5
@@ -20,10 +20,10 @@ matching on heterogeneous platforms.
 .SH "SEE ALSO"
 .BR "libpng"(3), " zlib"(3), " deflate"(5), " " and " zlib"(5)
 .LP
-PNG Specification (Third Edition) Candidate Recommendation Draft, July 2024:
+PNG Specification (Third Edition) Candidate Recommendation Draft, January 2025:
 .IP
 .br
-https://www.w3.org/TR/2024/CRD-png-3-20240718/
+https://www.w3.org/TR/2025/CRD-png-3-20250121/
 .LP
 PNG Specification (Second Edition), November 2003:
 .IP


ctruta (approved):

ctruta:
@svgeesus: please don't mind me for force-pushing away your merge commit. GitHub used to do a fast-forward when possible, but nowadays it seems to enforce `git merge --no-ff` and I don't know how to disable that.

So I force-pushed a recent commit from @jbowler "as if" you @svgeesus had fast-forwarded yours (instead of merging it). For a bunch of reasons that are too many to enumerate here, I am sticking to a linear history for as much as possible -- for now. I'm doing merges only when necessary. (For example: I'm doing merges each time I resync the libpng18 branch with the recent developments in the libpng16 branch.)

Apologies in advance shall also go to the other libpng contributors in the near future. The proper permanent solution would be to set up a commit bot ("like the pros are doing it"), but I already promised the delivery of the RISC-V port, and I'd like to get libpng18 out sooner instead of later... so....

jbowler:
If I understand the project settings correctly it's possible to only allow "rebase" merges.  That's what I do anyway.

ctruta:
> If I understand the project settings correctly it's possible to only allow "rebase" merges. That's what I do anyway.

Rebases are okay, although they'd be even better if they would have been allowed to fast-forward. Which they used to, but now they don't.

In situations like PR #659 (which, BTW, could please review) the rebases are alright. One or more commits, each of which are doing their own thing, integrated as such. My pet peeve is when a contributor submits several commits in a row, including fixups, which I would want to squash-and-rebase, but GitHub only gives me the choice of squash-and-merge.

I do almost all the Git and GitHub stuff from the terminal anyway, so, I'm good.

Issue #656, opened 2025-02-13
Memory leaks in png_malloc_base Function in libpng 1.6.43-1.6.46


kittener:
I found the vulnerability while testing my new fuzz tool.

## Description
When running pngimage with AddressSanitizer (ASan), the program leaks memory in various locations, eventually leading to high memory usage and causing the program to become unresponsive (hang). This suggests that certain allocations are not freed before exiting the program.

## Steps to Reproduce
1. Compile the application.
2. Run the application using the following command: 
`pngimage poc.png`

![Image](https://github.com/user-attachments/assets/4b005412-4778-453d-b1d4-2c9f2d676c7e)

3. Observe that the application crashes indicating a Memory leaks error.

## Expected Behavior
The application should manage memory correctly without trying to access memory after it has been freed.

## Actual Behavior
The application crashes, showing a heap use-after-free error. Here's the output from AddressSanitizer:
```
Direct leak of 352 byte(s) in 1 object(s) allocated from:
    #0 0x7f5c3813cc47 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x55a82f8fcbdf in png_malloc_base ../pngmem.c:97
    #2 0x55a82f8e7084 in png_create_info_struct ../png.c:376
    #3 0x55a82f8e17a1 in read_png ../contrib/libtests/pngimage.c:891
    #4 0x55a82f8e1889 in update_display ../contrib/libtests/pngimage.c:933
    #5 0x55a82f8e4c95 in test_one_file ../contrib/libtests/pngimage.c:1426
    #6 0x55a82f8e5211 in do_test ../contrib/libtests/pngimage.c:1577
    #7 0x55a82f8e5d8e in main ../contrib/libtests/pngimage.c:1681
    #8 0x7f5c37d520b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)

Indirect leak of 7160 byte(s) in 1 object(s) allocated from:
    #0 0x7f5c3813cc47 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x55a82f8fcbdf in png_malloc_base ../pngmem.c:97
    #2 0x55a82f8fce6a in png_malloc_warn ../pngmem.c:215
    #3 0x55a82f8e6276 in png_zalloc ../png.c:119
[...]

Issue #655, opened 2025-02-13
Memory leaks in png_malloc_base Function in libpng 1.6.43-1.6.46


kittener:
I found the vulnerability while testing my new fuzz tool.

## Description
While running `pngimage` tests with AddressSanitizer (ASan), memory leaks were detected in `read_png()` during calls to `png_create_read_struct()` and `png_create_info_struct()`. The leaked memory is allocated in `png_malloc_base()` but never freed.

## Steps to Reproduce
1. Compile the application.
2. Run the application using the following command: 
`pngimage poc.png`

![Image](https://github.com/user-attachments/assets/b7400a7d-b6f8-4e84-9390-f3ead995e1b4)

3. Observe that the application crashes indicating a Memory leaks error.

## Expected Behavior
The application should manage memory correctly without trying to access memory after it has been freed.

## Actual Behavior
The application crashes, showing a heap use-after-free error. Here's the output from AddressSanitizer:
```
Direct leak of 1240 byte(s) in 1 object(s) allocated from:
    #0 0x7f8c1e429c47 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x55f7c9f68bdf in png_malloc_base ../pngmem.c:97
    #2 0x55f7c9f68e6a in png_malloc_warn ../pngmem.c:215
    #3 0x55f7c9f52d3a in png_create_png_struct ../png.c:327
    #4 0x55f7c9f69151 in png_create_read_struct_2 ../pngread.c:44
    #5 0x55f7c9f690fa in png_create_read_struct ../pngread.c:32
    #6 0x55f7c9f4d737 in read_png ../contrib/libtests/pngimage.c:883
    #7 0x55f7c9f4d889 in update_display ../contrib/libtests/pngimage.c:933
    #8 0x55f7c9f50c95 in test_one_file ../contrib/libtests/pngimage.c:1426
    #9 0x55f7c9f51211 in do_test ../contrib/libtests/pngimage.c:1577
    #10 0x55f7c9f51d8e in main ../contrib/libtests/pngimage.c:1681
    #11 0x7f8c1e03f0b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)

Direct leak of 352 byte(s) in 1 object(s) allocated from:
    #0 0x7f8c1e429c47 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
[...]

PR #657, opened 2025-02-13
pngimage: clean up on user/internal errors

jbowler:
pngimage: The code simply exited with a return code of 99 in the event
of a user error including giving pngimage invalid PNG files and an
internal error.  It now attempts to clean up the state before doing so,
matching the normal behaviour.

pngimage: Non-ISO use of setjmp(3) corrected.

pngerror.c: Failure to call png_image_free on a false result from a
png_safe_execute function call fixed.  This was a regression caused by
the 'volatile' clean-up.  Not normally detectable because png_image_free
will often be called by the application.

### DIFF
--- a/pngerror.c
+++ b/pngerror.c
@@ -935,23 +935,37 @@ png_safe_warning(png_structp png_nonconst_ptr, png_const_charp warning_message)
 int /* PRIVATE */
 png_safe_execute(png_imagep image, int (*function)(png_voidp), png_voidp arg)
 {
-   png_voidp saved_error_buf = image->opaque->error_buf;
+   const png_voidp saved_error_buf = image->opaque->error_buf;
    jmp_buf safe_jmpbuf;
-   int result;
 
    /* Safely execute function(arg), with png_error returning back here. */
    if (setjmp(safe_jmpbuf) == 0)
    {
+      int result; /*bool*/
+
       image->opaque->error_buf = safe_jmpbuf;
       result = function(arg);
       image->opaque->error_buf = saved_error_buf;
-      return result;
+
+      if (result)
+         return 1; /*true*/
    }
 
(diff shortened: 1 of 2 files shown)


jbowler:
Fixes #655 and should fix #656; pngimage "leaks" if given an invalid file to check (one which errors out on read, not one which just warns).

pngerror.c
```diff
@@ -935,23 +935,37 @@ png_safe_warning(png_structp png_nonconst_ptr, png_const_charp warning_message)
...
 {
-   png_voidp saved_error_buf = image->opaque->error_buf;
+   const png_voidp saved_error_buf = image->opaque->error_buf;
    jmp_buf safe_jmpbuf;
-   int result;
 
    /* Safely execute function(arg), with png_error returning back here. */
    if (setjmp(safe_jmpbuf) == 0)
    {
+      int result; /*bool*/
```
ctruta, line 944: I agree with your commit, and I also agree that booleans should be marked as such, but this particular manner of doing it is kinda' erratic. The libpng code is inconsistent enough as-is.

Since I haven't abandoned the idea of moving up libpng18 development to requiring C99 (or newer), and stdint.h (or inttypes.h), I also think we should require stdbool.h.

ctruta (approved):

pngerror.c
ctruta: Following up: after a cursory look over the libpng API, I came to realize that it's even worse than I thought.

It's not just that sometimes 0 means "failure" and 1 means "success" and other times it's the other way around. It's also the fact that this is the kind of frozen API which we cannot modify without breaking apps.

For this reason, I opted for using the much more descriptive annotations `/* success */` and `/* failure */`, which are already in use elsewhere in libpng.

ctruta:
Integrated in trunk -- many thanks.

PR #660, opened 2025-02-22
PR659 fixes

jbowler:
- **chore: Clean up remnants of the long-discontinued Win16 support**
- **chore: Clean up the `FILE *` formulations in code and in documentation**
- **Remove the vestigial support for custom API calling conventions**
- **Minor fixes for PR659**

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -284,7 +284,7 @@ pnglibconf.c: scripts/pnglibconf/pnglibconf.dfa scripts/pnglibconf/options.awk p
 	mv $*.tf5 $@
 
 # Symbol checks (.def and .out files should match)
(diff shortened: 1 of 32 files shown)


jbowler:
I'll force push a new version without the "skippted" typo.

ctruta:
Thank you for this. I applied almost all fixes, but after seeing your fix to my `STDERR`-to-`stderr` replacements in pngtest.c, I came to realize that my original changes didn't actually belong to that particular commit. I decided to undo those ones out of the final revision that I've just integrated in libpng16 and libpng18.

I'm leaving pngtest.c for later, as it is not essential cleanup, but rather a small "paint job". I'm not even sure whether those `fprintf` should be sent to `stderr` or to `stdout`; ergo, I'm not even sure whether those `fprintf` should be, in fact, `printf`. TBD.

PR #661, opened 2025-02-22
C90: correct initialisation of ARM SIMD register

jbowler:
arm/palette_neon_intrinsics.c: change the initialisation of the "opaque
black" standard PLTE filler to be in-line rather than using a C99 style
{} initializer which calls functions.

This change also reduces the number of function calls required from 4 to
2.

### DIFF
--- a/arm/palette_neon_intrinsics.c
+++ b/arm/palette_neon_intrinsics.c
@@ -30,12 +30,10 @@ png_riffle_palette_neon(png_structrp png_ptr)
    int i;
 
    /* Initially black, opaque. */
-   uint8x16x4_t w = {{
-      vdupq_n_u8(0x00),
-      vdupq_n_u8(0x00),
-      vdupq_n_u8(0x00),
-      vdupq_n_u8(0xff),
-   }};
+   uint8x16x4_t w;
+
+   w.val[2] = w.val[1] = w.val[0] = vdupq_n_u8(0x00);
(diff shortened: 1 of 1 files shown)


jbowler:
This removes the remaining C99ism in libpng 1.6.  The same issue exists in 1.8 so this can be applied there too (there should be no conflicts) but 1.8 also requires a correction to a C99ism that I accidentally introduced in the same body of code (this won't conflict with this change.)

Tested (successfully) on aarch64, Ubuntu.
Tested partially on armv7a Ubuntu; armv7a is no longer supported by Ubuntu and as a consequence my cross-build of pngvalid will not run because of changes in the glibc DLL.

ctruta (commented):
I am nowhere close to being a SIMD expert, let alone an ARM expert, but I am under the impression that the original code is better than the proposed change. These are SIMD intrinsics, not actual function calls. My inclination is to give a thumbs-down.

I would highly appreciate the review of a SIMD expert.

From the practical point of view, there is no need to avoid C99 with ARM Neon at all, because all ARM Neon compilers are also C99 compilers. We never had a problem with the original code.

From the language standards point of view, we should be moving wholesale to C99 or later, sooner or later.

jbowler:
I guess you don't think they're lambda expressions either, or machine instructions.

Duh.  They're *not* static data; they are executable code.  Think "C" not some curious high level abstraction which isn't anything meaningful. 

Whatever.

PR #662, opened 2025-02-26
Fix apple-non-apple cross building

joostn:
When cross compiling for an apple platform on a non-apple platform (or the reverse), if(APPLE) in line 32 should refer to the target platform. Without importing the APPLE variable it refers to the host platform. Cross compiling with a toolchain file fails.


jbowler:
The problem is not Apple specific; it applies to all cross-builds unless the full cross build is set up in cmake.  I have a fix which works everywhere:

https://github.com/jbowler/libpng/tree/enable-cmake-cross-build

But it's queued up behind all my other pending PRs and I believe @ctruta just expects PNG_TARGET_ARCHITECTURE to be set on the command line, or something like that (it's not clear, I don't know why it was introduced.)

joostn:
Some more context, this is the issue I was running into:

I have a toolchain file containing something like this:
```
set(CMAKE_SYSTEM_NAME Darwin)
set(CMAKE_OSX_SYSROOT ...)
```
I'm building libpng on linux, targeting macos, with -DCMAKE_TOOLCHAIN_FILE.

The toolchain file causes cmake to set APPLE to true. But the generated genout.cmake does not include a SET(APPLE). So when running ``cmake -p genout.cmake`` APPLE is set to the host's value, not the target. As a result, the next set(CMAKE_OSX_SYSROOT..) is skipped and the sysroot is thus ignored. So genout's OUTPUT is built against the host includes instead of the target's.

On Nixos this causes the build to fail, as there is no /usr/include on the host, only on the target. On other distros it might work but it is using the wrong includes which might cause issues.

jbowler:
Yeah, I'm pretty near 100% sure that is a separate issue; my fix was for the case where the architecture settings don't match either because PNG_TARGET_ARCHITECTURE wasn't set up (no toolchain setup) or the architecture doesn't match the strings because they are anchored (a common problem with OSX specifically).

I assume your fix works for your situation (because OSX_ARCHITECTURE has a **leading** prefix which matches)  So this looks like a good fix to me.

Do you know if this is a regression?  I.e. did it work for you on some earlier release?

joostn:
Hi John, it's not a regression, I'm now running into this because I switched to Nix for compiling. This creates a sandbox without a local /usr/include. It's probably not a big deal for most users, but applying my one liner might be helpful to some. The problem is related to cross-platform building, not cross-architecture.

Since you mention it, PNG_TARGET_ARCHITECTURE is another issue on macos. Libpng build breaks when I set multiple architectures: CMAKE_OSX_ARCHITECTURES=arm64;x86_64. 
I'm working around this by just building twice and then use lipo to create a fat library and this works fine.

I think it could be fixed in libpng: instead of checking for Neon in CMakeLists.txt, this should be done (additionally) in the actual source files by guarding with #ifdef __aarch64__. This way, for fat builds (arm64+x86_64) neon is only used in the arm64 pass. But it's complicated probably, because automake/configure is also used. And arm32 is also targeted.

Anyway that's unrelated to this pull request!

jbowler:
> I think it could be fixed in libpng: instead of checking for Neon in CMakeLists.txt, this should be done (additionally) in the actual source files by guarding with #ifdef **aarch64**.

That's what happens if PNG_ARM_NEON_OPT is not set; the compiler flags are used.  This is fixed completely in 1.8 by simply removing all the PNG_hardware_OPT build system settings.  I did, in fact, fix this over 10 years ago but the full fix wasn't accepted to avoid changing the behavior of the build options.  The fix was done originally for Apple multi-arch build support.

Your fix is, I believe still required even in 1.8.  I don't encounter this because my own testing is done with gentoo "crossdev" and the toolchain, in particular the compiler, automagically sets the SYSROOT to the sandbox containing the target specific files.  IRC that happens correctly even if the target is the same as the host and happens for the same reason; "cross" building for a different platform on the same hardware.  Indeed this is how amd64/x86 gentoo systems are typically bootstrapped from scratch (though few people do that these days.)

joostn:
I'm on 1.6.46 so that explains a lot. I will check out 1.8, thanks!

jbowler:
@joostn does 1.8 fix (i.e. avoid) the problem?  It should do because it requires no support in any build system.

joostn:
Hi John, I can try later this week.

joostn:
Same problem on 1.8. My one-line patch fixes it.

jbowler:
@ctruta; required on both 1.6 and 1.8 for cross builds with cmake.  I don't know enough about `cmake` to review the change but the explanation is reasonable and believable; the host system root has to be set correctly.  With Gentoo `crossdev` and `configure` no extra setup is required.  It seems that `cmake` requires extra setup.

joostn:
To summarize: cross compiling for apple fails if the local host does not have a /usr directory, because genout.cmake will attempt to use host files instead of those from the sysroot.

ctruta:
@jbowler wrote:
> I don't know enough about `cmake` to review the change but the explanation is reasonable and believable

Commit dc71f481ff32fa474e030ceb0b33def4fa9edc6a says you do :stuck_out_tongue:

@joostn thank you for your contribution :smile:

Issue #669, opened 2025-03-17
Incorrect chromaticity values stored for mDCV chunk


Morphinity:
When using the newly introduced floating point API `png_set_mDCV`, the color primaries chromaticity values are being returned with seemingly incorrect values. Eg when passing following color primaries for ITU-R-BT.2100 mentioned in https://w3c.github.io/png/#example-5
```
Rxy = (0.708, 0.292),
Gxy = (0.170, 0.797),
Bxy = (0.131, 0.046)
Wxy = (0.3127, 0.3290)
```
to `png_set_mDCV(png_ptr, info_ptr, ...)`, the `info_ptr` returns with values 
```
info_ptr->mastering_red_x: 17700
info_ptr->mastering_red_y: 7299
info_ptr->mastering_green_x: 4250
info_ptr->mastering_green_y: 19925
info_ptr->mastering_blue_x: 3275
info_ptr->mastering_blue_y: 1150
```
The stored decimal values according to the above table should be - 
```
mastering_red_xy: { 35400, 14600 }
mastering_green_xy: { 8500, 39850 }
mastering_blue_xy: { 6550, 2300 }
```
Note that they are exactly half of what they should be. According to the spec, we are supposed to multiply the floating values by 50,000. Instead we end up multiplying by 25,000 because we divide the values by 2 once in `png_set_mDCV` (passing `white_x / 2` to `png_fixed` to scale it by 50,000 instead of 100,000) and then again divide by 2 in `png_set_mDCV_fixed` > `png_ITU_fixed_16` which internally divides the values by 2.

@ctruta @jbowler

Miloni2010:
@ctruta, @jbowler, @ProgramMax, Hey guys, any update on this issue regarding the mDCV chunk? We are blocked at this and wanted to understand why we must divide the mDCV values by 2 twice instead of doing it just once to scale it correctly.

jbowler:
@Miloni2010 good analysis and correct.  The 8 lines in `pngset.c` starting at 306 (white_x/2) should all not be dividing by 2.  Kind of obvious; `png_set_mDCV_fixed` takes `png_fixed` arguments so the `png_fixed` macro should just work.

I assume that the white point values are similarly messed up (please say if _not_ because then I don't know what is going on :-)

My apologies; an easy mistake that shows up in code review or testing.

I suggest submitting a PR; just delete the " / 2" on all the lines.  I'm a couple of months out of date WRT libpng and don't have the time at present.

[...]

PR #672, opened 2025-04-19
mDCV bug: the FP API set incorrect values

jbowler:
BUG: png_set_mDCV chromaticities set incorrectly

png_set_mDCV was dividing the chromaticities by two.

### DIFF
--- a/pngset.c
+++ b/pngset.c
@@ -300,17 +300,14 @@ png_set_mDCV(png_const_structrp png_ptr, png_inforp info_ptr,
     double maxDL, double minDL)
 {
    png_set_mDCV_fixed(png_ptr, info_ptr,
-      /* The ITU approach is to scale by 50,000, not 100,000 so just divide
-       * the input values by 2 and use png_fixed:
-       */
-      png_fixed(png_ptr, white_x / 2, "png_set_mDCV(white(x))"),
-      png_fixed(png_ptr, white_y / 2, "png_set_mDCV(white(y))"),
-      png_fixed(png_ptr, red_x / 2, "png_set_mDCV(red(x))"),
-      png_fixed(png_ptr, red_y / 2, "png_set_mDCV(red(y))"),
-      png_fixed(png_ptr, green_x / 2, "png_set_mDCV(green(x))"),
-      png_fixed(png_ptr, green_y / 2, "png_set_mDCV(green(y))"),
-      png_fixed(png_ptr, blue_x / 2, "png_set_mDCV(blue(x))"),
-      png_fixed(png_ptr, blue_y / 2, "png_set_mDCV(blue(y))"),
+      png_fixed(png_ptr, white_x, "png_set_mDCV(white(x))"),
+      png_fixed(png_ptr, white_y, "png_set_mDCV(white(y))"),
+      png_fixed(png_ptr, red_x, "png_set_mDCV(red(x))"),
+      png_fixed(png_ptr, red_y, "png_set_mDCV(red(y))"),
+      png_fixed(png_ptr, green_x, "png_set_mDCV(green(x))"),
+      png_fixed(png_ptr, green_y, "png_set_mDCV(green(y))"),
+      png_fixed(png_ptr, blue_x, "png_set_mDCV(blue(x))"),
+      png_fixed(png_ptr, blue_y, "png_set_mDCV(blue(y))"),
       png_fixed_ITU(png_ptr, maxDL, "png_set_mDCV(maxDL)"),
       png_fixed_ITU(png_ptr, minDL, "png_set_mDCV(minDL)"));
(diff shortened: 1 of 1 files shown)


ctruta (approved):

ctruta (changes requested):
Re-reviewing...

Umm, actually, this is only half-fixed. (Considering the nature of this defect, intending a pun is rather appropriate.)

Only the setter has been fixed, but the getter needs fixing, too. (I'll do it. This calls for an immediate fix, as well as an immediate release of libpng v1.6.48.)

ctruta (approved):
Uh...
Re-re-re-...-reviewing: the fix by @jbowler is correct, only the setter was broken, and the differences in how `png_get_mDCv*` are implemented vs. how `png_set_mDCv*` are implemented makes the code extra-confusing.

jbowler:
I checked both.  The explanation is in the #699 where I explain how the problem arose.

To elaborate on the principle.  The "fixed" APIs are normally the principle ones because originally relying on hardware with adequate FP support produced problems (unsupported APIs) on lower end systems.  The parameters being set are also invariably integral because PNG does not use a native FP (like IEEE 488 stuff); going via FP introduces the complexity of checking for rounding errors and requires using double because float does not have a full 32 bits of precision (not that this matters much in practice).

Consequently at present almost all the checks are implemented in the fixed APIs to ensure code commonality and, more important, bug commonality if the checks are wrong.

So I changed png_set_mDCV to call png_set_mDCV_fixed without undoing the divide that was there originally.  Oops.

It's a PITA that the W3C reinvented the cHRM chunk with a different format and a different set of error/range conditions.  Totally pointless.  It would have been much, much, better to fix the original problem with cHRM by saying that cHRM **is** the mastering display colour volume (it is if there is no mDCV) and adding a proper CIEXYZTRIPLE (in Windows terms) chunk for the encoding colour space (which defaults to the horrible matrix invert back calculation from cHRM at present).

ctruta:
@jbowler wrote:

> I checked both. The explanation is in the #699 where I explain how the problem arose.

I think you meant #669. But otherwise, I hear you.

> So I changed png_set_mDCV to call png_set_mDCV_fixed without undoing the divide that was there originally. Oops.

This is an rather solid indication that the pngtest program needs to check both the fixed-point API and the floating-point API. It should not be unnecessarily restricted to check just one, and worse, to be restricted at compile time. At the very least, there should be run-time option, "check the fixed-point API" vs "check the floating-point API", and then we should run pngtest in our test suite with this flag on and off. FYI I already started implementing that. Not for v1.6.48 (soon to be released) but maybe for later, and definitely for v1.8.0.

jbowler:
> This is an rather solid indication that the pngtest program needs to check both the fixed-point API and the floating-point API. It should not be unnecessarily restricted to check just one, and worse, to be restricted at compile time.

It's the same as `contrib/conftest/{read,write}-full.dfa` - I added those because even though there were tests in contrib/conftest for both FP and fixed APIs there were confined with "likely" minimal configurations so skipped many of the chunks.

See `contrib/conftest/float-fixed.dfa` and `contrib/conftest/fixed.dfa` but those don't take into account the fact that `pngtest` **defaults** to fixed point (because of the way the #if/#else/#endif test is written.)  At least I think that's the problem.  See line 1119 of pngtest.c.

I think a minimalist fix is to change 1119 to `#ifndef PNG_FLOATING_POINT_SUPPORTED` on the basis that one or the other is required.  Another approach is `contrib/conftest/float.dfa`:

`option FIXED_POINT off`

NOTE: I originally said "FIXED_ARITHMETIC" but there is no such option, its:

FIXED_POINT - fixed point APIs
FLOATING_POINT - floating point APIs
FLOATING _ARITHMETIC - allow internal use of FP code

Hum, won't take long to test :-)  (EDIT: `contrib/conftest/float.dfa` doesn't seem to fail as required; pngtest passes...)

jbowler:
Both checks work as expected with configure - my problem before was not doing a clean build.  I.e. without the png_set_mDCV fix an unconfigured build fails the pngtest check and configured builds fail as expected according to whether the floating point APIs are used (fail) or not (pass).

cmake passes but produces a bogus pngtest; i.e. cmake apparently isn't doing the binary compare because when I run the generated pngtest directly it produces a changed pngtest.out but exits with a 0 (success) exit code.   My build script also doesn't manage to do the minconfigs for cmake.  I suspect DFA_XTRA isn't working or maybe I don't understand cmake; I'm setting DFA_XTRA in the environment.

Both the cmake problems are not new.   I PRed #675 for your perusal.  It's not as good as testing both APIs at once but the "conftest" is useful on its own because it causes all the test programs (not just pngtest) to use the FP APIs (I think they default to that anyway but I haven't checked).

PR #674, opened 2025-04-19
SECURITY: discourage application backdoors

jbowler:
Add extra checking to the private libpng header files to attempt to
detect inclusion from application code.  Including any of these three
header files in application code raises the possibility of application
crashes as a result of internal changes to the png_struct and png_info
structures.

### DIFF
--- a/pnginfo.h
+++ b/pnginfo.h
@@ -10,43 +10,24 @@
  * and license in png.h
  */
 
- /* png_info is a structure that holds the information in a PNG file so
- * that the application can find out the characteristics of the image.
- * If you are reading the file, this structure will tell you what is
- * in the PNG file.  If you are writing the file, fill in the information
- * you want to put into the PNG file, using png_set_*() functions, then
- * call png_write_info().
+#ifndef PNGPRIV_H
+/* PRIVATE INTERNAL HEADER FILE
  *
- * The names chosen should be very close to the PNG specification, so
- * consult that document for information about the meaning of each field.
+ * This file may not be included directly.
+ */
+#  error "PRIVATE HEADER INCLUDED: pnginfo.h is an internal libpng file."
+#endif
+
+/* INTERNAL, PRIVATE definition of a PNG.
  *
- * With libpng < 0.95, it was only possible to directly set and read the
- * the values in the png_info_struct, which meant that the contents and
- * order of the values had to remain fixed.  With libpng 0.95 and later,
- * however, there are now functions that abstract the contents of
- * png_info_struct from the application, so this makes it easier to use
- * libpng with dynamic libraries, and even makes it possible to use
- * libraries that don't have all of the libpng ancillary chunk-handing
- * functionality.  In libpng-1.5.0 this was moved into a separate private
- * file that is not visible to applications.
+ * png_info is a modifiable stateless description of a PNG.  Fields in png_info
+ * are accessed through png_get_<CHUNK>() functions and set using
+ * png_set_<CHUNK>() functions.
  *
- * The following members may have allocated storage attached that should be
- * cleaned up before the structure is discarded: palette, trans, text,
- * pcal_purpose, pcal_units, pcal_params, hist, iccp_name, iccp_profile,
- * splt_palettes, scal_unit, row_pointers, and unknowns.   By default, these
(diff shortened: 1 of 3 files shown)


enh-google (approved):
(fwiw, i'd be tempted to change the error messages to say "include <png.h> instead" ... otherwise my experience is you'll get idiots just #define or #undef around their naughty #include 🤦 )

jbowler:
That works too, given that there is a comment before it explaining what to do in new **libpng** source files (note that I didn't change any libpng source files; they already do the right thing.)  @ctruta , what is your preference?

ctruta (approved):

ctruta:
I don't have a strong opinion. If there's an `#error` in there, I'm good.
But FWIW, I do like the general idea of phrasing error messages in the manner _"don't do THIS; do THAT instead"_.

jbowler:
Ok.  This error in pngpriv.h seems to be sufficient:

`#error "Include png.h not pngpriv.h for the supported libpng API."`

The same message could be used everywhere (pnginfo.h and pngstruct.h) but no .c file should be including those; they have to be included from the correct place in pngpriv.h to guarantee that the pnglibconf.h settings (which change both structures) are correct.  So I came up with:

`#error "Do not include pngstruct.h.  Include png.h for the libpng API."`

@ctruta please feel free to change the messages as desired.

ctruta:
So I developed a commit on top of yours, and I gave you credit, but I made a bunch of modifications, the most notable of which is to refrain from calling this a "SECURITY" issue. I'm not normally the one to advocate for security through obscurity, but in this case, this would be tantamount to identifying a fire hazard inside a building which has been in there for decades; and then, proactively fixing it; and then, starting to scream "FIRE".

Not to mention that this change does not actually prevent bad actors from circumventing any of that; this is rather a method to alleviate the inadvertent introduction of unwanted backdoors -- mind you, a rather good method -- so, thank you.

> @ctruta please feel free to change the messages as desired.

I removed some of the commenting ceremony, considering that the error directives themselves are self-explanatory. And I rephrased the error lines so that we can literally and liberally copy and paste them into any other internal header file. The application developers should generally refrain from including ***any*** of our internal headers, regardless of their content. See my change to pngdebug.h.

jbowler:
Just in case someone finds this now closed PR here's the historical backtrace of what happened; I only worked this out a few days ago:

1. Prior to libpng-0.95 (according to the comments) there were no accessor functions so `png_info` had to be read and written directly and png_struct::jmpbuf (but nothing else) had to be accessed by the app for `setjmp`.
2. libpng-0.95 (from the comments) added the accessor functions but kept the access to `png_struct::jmpbuf` still required and so the first official release (libpng-1.0) deprecated everything except `png_struct::jmpbuf`  It did this via a macro `PNG_DEPSTRUCT` which only worked on GCC where it added `__attribute__((__deprecated__))` to all the members except `jmpbuf`.
3. Even though such access was deprecated it was still possible at the cost of a warning on GCC.  This works to this day (I just built libpng-1.0 and checked) but, of course, it only produces a warning and adding -DPNG_DEPSTRUCT= to the CFLAGS makes it go away.
4. libpng-1.2 was the same and produces the "deprecated" warnings with GCC.
5. libpng-1.4, at least the latest version, **does not define PNG_DEPSTRUCT in pngconf.h**  So the 1.0 and 1.2 warning stopped being produced even though PNG_DEPSTRUCT was still used in the structure definitions in png.h  I don't know why this was done.
6. libpng-1.4 also added PNG_DEPSTRUCT to `png_struct_def::jmpbuf` because it added the new accessor `png_longjmp_ptr` but the comments in png.h still talk about accessing it directly.  1.4 also added png_set_longjmp_fn and the macro png_jmpbuf.
7. Thus 1.4 is the first version of libpng where `png_struct` is entirely internal.
8. libpng-1.5 finally did the deed of hiding `png_struct` (in the new pngstruct.h) and `png_info` (in pnginfo.h).

The fact that the PR referred to an app which included pngstruct.h may indicate that it is a minimally maintained libpng-1.4 app or maybe it just dates from the dawn of PNG.  The pngstruct.h behaviour stopped the compile regardless of whether `png_info` is also accessed.

PR #675, opened 2025-04-25
testing: update pngtest FP and fixed API checks

jbowler:
pngtest.c is changed to check the floating point APIs by default as
these are likely to be the most used and in a number of cases they still
use the fixed point APIs.

contrib/conftest/float.dfa is added to turn off the fixed-point APIs and
these force use of the available floating point APIs in all test
programs when used as the build configuration.

### DIFF
--- a/contrib/conftest/float.dfa
+++ b/contrib/conftest/float.dfa
@@ -1,0 +1,17 @@
+# float.dfa
+#  Build time configuration of libpng
+#
+# Author: John Bowler
+# Copyright: (c) John Bowler, 2025
+# Usage rights:
+#  To the extent possible under law, the author has waived all copyright and
+#  related or neighboring rights to this work.  This work is published from:
+#  United States.
+#
+# Test the standard libpng configuration without fixed point (the internal
+# floating point implementations are used instead).  This does not turn off all
+# fixed point; in some cases the only implementation is fixed point, however it
+# does make those implementations private to libpng forcing use of the floating
+# point variants.
+#
+option FIXED_POINT off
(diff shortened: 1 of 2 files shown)


jbowler:
Passes all checks (including minconfigs) and fails appropriately without the png_set_mDCV fix in configure.

cmake does not seem to be doing the binary diff check on `{pngtest.png,pngtest.out}`.  I don't think that is new and I don't know why it builds a different pngtest with the default options.

ctruta:
Thank you, but this is only a one-off, and our testing as driven by pngtest.c is not any less brittle. In my other comment from that other issue, I meant to say that I will be making sweeping changes to pngtest, so that it can test both the fixed-point API and the floating-point API.

So I have a draft already, which I can send it to you @jbowler for review, but for now I'd like to focus on releasing libpng-1.6.48 instead.

ctruta:
> cmake does not seem to be doing the binary diff check on `{pngtest.png,pngtest.out}`. I don't think that is new and I don't know why it builds a different pngtest with the default options.

Off-topic: we need to test at least some (if not all) of the contrib/conftest/*.dfa on our CI bots. Neither .travis.yml nor .appveyor.yml nor any script in the ci/ subdirectory know that these DFA test files exist at all. (And off-topic to the off-topic: .travis.yml will be deleted, with many thanks to Travis CI for all of their service to libpng.)

jbowler:
> our testing as driven by pngtest.c

"Our"?  You malign me.  I have **never** **ever** regarded `pngtest` as anything other than Glenn's private little sandpit where he could play around to his heart's content.  It's quaint.

You might not have noticed that I wrote all the real tests.

PR #680, opened 2025-05-14
Improved fuzzer

cristiano-masutti:
Improved fuzzer to target png_init_read_transformations()


ctruta:
> Improved fuzzer to target png_init_read_transformations()

Thank you. Could you please squash all your draft changes into a single commit, and explain what the squashed commit does in the commit message, using the rest of our commit messages as a stylistic guideline, as well as this guideline:

https://www.conventionalcommits.org/en/v1.0.0/

jbowler:
I don't think squashing commits is helpful, particularly not here.  The changes are apparently atomic and address specific issues so they should be separate commits.

So far as the commits are concerned I concur with @ctruta; "lets see" is not helpful.  @ctruta's link is a proforma that is reasonable to follow but I don't; I've had too many people in my life telling me how to speak English :-)

Bottom line, say what was changed in the title (approximately, summarise), say what the change was in the description and say why.

Don't use // comments in 1.6 code, unless @ctruta wants to change the rules in 1.6 code (I favour -std=c11).

We can't change the past in git without --force so, @ctruta, if it works and given that the whole PR only affects exactly one file that, so far as I know, is completely dysfunctional, why look a gift horse in the mouth?

PR #682, opened 2025-05-15
Working libpng version

cristiano-masutti:
went back to working libpng commit

### DIFF
--- a/contrib/oss-fuzz/libpng_read_fuzzer.cc
+++ b/contrib/oss-fuzz/libpng_read_fuzzer.cc
@@ -12,10 +12,13 @@
 // 4. adding read_end_info() and creating an end_info structure.
 // 5. adding calls to png_set_*() transforms commonly used by browsers.
 
+
+
 #include <stddef.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


jbowler:
@bobfriesenhahn simply delete the contrib/oss-fuzz directory.  I have **never** _not once_ managed to get it to do anything useful.  Apparently Glenn could but he never explained the particular magic environment required and there is no support in cmake or configure or even a script.

If there is one single thing on this planet that managed to convince me that fuzzing is a complete waste of time this directory is it.

bobfriesenhahn:
@jbowler, Due to maintaining GraphicsMagick's oss-fuzz build, I have some knowledge of how things work.  Part of the build is controlled by oss-fuzz project scripts, but libpng's own contrib/oss-fuzz/build.sh is executed due to a line in contrib/oss-fuzz/Dockerfile.  In fact this was at least building every day until libpng's configure.ac started producing a syntax error.

It is best to assure that the libpng oss-fuzz build is working properly, and check that the libpng read fuzzer is indeed doing something useful. I am not a recognized administrator for libpng oss-fuzz so I can not see details of oss-fuzz generated bug reports, but I see that at 'https://issues.oss-fuzz.com/issues?q=is:open%20libpng' there are three issues open, and one is recent (April 29, 2025), so it must have been working.

There is no need to torch the oss-fuzz builds.

ctruta:
@cristiano-masutti, see my other comment in PR #680.

PR #683, opened 2025-05-19
riscv: autotools update

filipwasil:
I updated the configure.ac with necessary checks and confirmed that following autotools calls work as intended.
Considering such a build command:

`
./autogen.sh --maintainer && ./configure CFLAGS="-march=rv64gcv1p0" --enable-riscv-rvv=<on/yes/off/no/check/api> && make -j4
`

The only way to compile the rvv code is to define the march in c flags the v extension. For example:
`CFLAGS="-march=rv64gcv1p0" ` 
then any of the options will work:
`
--enable-riscv-rvv=on/yes/off/no/check/api
`

the default behaviour is: yes/on (as above: only if v extension is available in march in C_FLAGS)

if the test mini program fails to compile (one in AC_COMPILE_IFELSE ), then the files will not be included in compilation and passing --enable-riscv-rvv=on/yes/off/no/check always results in off/no.

I did not upload the generated files, but i can say that when I generated it, it worked as described above.
namely:
`
	modified:   Makefile.in
	modified:   aclocal.m4
	modified:   compile
	modified:   config.guess
	modified:   config.h.in
	modified:   config.sub
	modified:   configure
	modified:   depcomp
	modified:   install-sh
	modified:   ltmain.sh
	modified:   missing
	modified:   scripts/autoconf/libtool.m4
	modified:   scripts/autoconf/ltoptions.m4
	modified:   scripts/autoconf/ltsugar.m4
	modified:   scripts/autoconf/ltversion.m4
	modified:   scripts/autoconf/lt~obsolete.m4
	modified:   test-driver
`

The only thing i did not test is the **--enable-riscv-rvv=api**.
I can tell that it is included/excluded as expected (in regard to compilation).
I will focus on this as the next step.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -136,7 +136,7 @@ Your directory structure should look like this:
              depcomp, install-sh, mkinstalldirs, test-pngtest.sh, etc.
           contrib
              arm-neon, conftest, examples, gregbook, libtests, pngminim,
-             pngminus, pngsuite, tools, visupng, riscv-vector
+             pngminus, pngsuite, tools, visupng, riscv-rvv
           projects
              owatcom, visualc71, vstudio
           scripts
--- a/README
+++ b/README
@@ -163,7 +163,7 @@ Files included in this distribution
                           programs demonstrating the use of pngusr.dfa
         pngminus/     =>  Simple pnm2png and png2pnm programs
         pngsuite/     =>  Test images
-        riscv-vector/ =>  Optimized code for the RISC-V Vector platform
+        riscv-rvv/    =>  Optimized code for the RISC-V Vector platform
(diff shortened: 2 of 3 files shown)


filipwasil:
@ctruta

ctruta:
@jbowler, @bobfriesenhahn: I was briefly away from the libpng project. Meanwhile, Filip's commit was sitting here, waiting to be integrated. A few minutes ago I did just that, and then I regenerated the build files.

I had a brief look at #685 and #693, and I think this PR solves all the problems addressed in your respective contributions. Sorry about your extra work, which I will reject as duplicates of this PR... But I appreciate it, regardless!

PR #684, opened 2025-05-22
Fix resource leak in iccfrompng.c by closing file streams properly

sanjay20m:
This patch fixes a resource leak in the contrib/examples/iccfrompng.c file where file streams were not properly closed after use. The fix ensures that all opened files are correctly closed, preventing potential memory/resource leaks during execution.

Impact:
Fixing this resource leak improves the stability and robustness of the tool, preventing resource exhaustion during repeated or bulk usage.

### DIFF
--- a/configure
+++ b/configure
@@ -15605,9 +15605,9 @@ printf "%s\n" "$as_me: WARNING: --enable-riscv-rvv: please specify 'check' or 'a
 fi
 
 
-# Add RISCV-specific files to all builds where $host_cpu is arm ('riscv64*') or
+# Add RISCV-specific files to all builds where $host_cpu is RISC-V ('riscv64') or
 # where RISCV optimizations were explicitly requested. (This allows a fallback
-# if a future host CPU does not match 'riscv64*'.)
+# if a future host CPU does not match 'riscv64'.)
 
  if test "$enable_riscv_rvv" != 'no' &&
     case "$host_cpu" in
--- a/contrib/riscv-rvv/README
+++ b/contrib/riscv-rvv/README
@@ -1,5 +1,5 @@
-OPERATING SYSTEM SPECIFIC ARM NEON DETECTION
---------------------------------------------
+OPERATING SYSTEM SPECIFIC RISC-V RVV DETECTION
+----------------------------------------------
 
 Detection of the ability to execute RISC-V Vector on a RISC-V processor
 requires operating system support.  (The information is not available in user
--- a/riscv/riscv_init.c
+++ b/riscv/riscv_init.c
@@ -1,4 +1,4 @@
-/* arm_init.c - RISC-V Vector optimized filter functions
+/* riscv_init.c - RISC-V Vector optimized filter functions
  *
  * Copyright (c) 2023 Google LLC
  * Written by Dragoș Tiselice <dtiselice@google.com>, May 2023.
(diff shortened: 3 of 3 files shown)


jbowler:
Please resubmit this without all the reformatting; alas Linus's git does not show the actual code changes and git stores all the formatting, so one-liners like this have to be PRed as just the one line that is changed!

sanjay20m:
I have updated the code now !!

jbowler (commented):
This should be rejected; even with the spurious blanks removed it is way more than the title suggest and the changes are massive changes to the functionality.

@ctruta: how about removing substantially all of "contrib".  Apart from valid and useful tests cases and maybe (just maybe) a couple of tools it is a vast unwarranted maintenance overhead.

sanjay20m:
Please Review the new update sir ...

sanjay20m:
Any changes required??

ctruta:
@jbowler wrote:

> This should be rejected; even with the spurious blanks removed it is way more than the title suggest and the changes are massive changes to the functionality.

I concur.

This is not a clean submission. Leaving aside the random whitespace insertions and deletions, as well as the oddly-formatted code portions (I mean, seriously, curly brackets inside a function shouldn't start at column one...) there are also random deletions of actual code, some of which arguably could be deleted (e.g. the `"OOM allocating string"` error message), but others absolutely shouldn't (e.g. the `"%s has no profile"` error message).

> @ctruta: how about removing substantially all of "contrib". Apart from valid and useful tests cases and maybe (just maybe) a couple of tools it is a vast unwarranted maintenance overhead.

Again, I concur. Maybe remove them from libpng16; surely remove them from libpng18.

contrib/examples/iccfrompng.c
```diff
@@ -123,20 +123,21 @@ extract_one_file(const char *filename)
             of = fopen(output, "wb");
             if (of != NULL)
             {
-            if (fwrite(profile, proflen, 1, of) == 1 &&
-    fflush(of) == 0 &&
-    fclose(of) == 0)
+            if (fwrite(profile, proflen, 1, of) != 1 ||
+    fflush(of) != 0 ||
```
ctruta: I still don't understand these formatting choices.

sanjay20m: check the new update

ctruta: @sanjay20m, please take a step back, go back to the original source, squash all of your commits into a single one, and rebase the squashed result onto the tip of source tree. (This is not a necessary workflow, but rather, a suggestion.) Regardless of your particular workflow, the end result should be an easy-to-review commit, not the big giant delta with random code additions, deletions and indentations that do not fit our coding style.

The `git diff` command should have a clean output.

Moreover, after all of my reviews, I still cannot explain the disappearance of error messages like `"%s has no profile"`. (And not only that.) The rule of thumb is: if you're seeing it deleted in your `git diff` output, then it's wrong.

sanjay20m: Dear Sir, I request u to check the update I made, it has minimal changes and it has the main issue fixed : fix: close input file after extracting profile
So refer to this one : [61ec911](https://github.com/pnggroup/libpng/pull/684/commits/61ec911899d4c3a80c033aa36c7ef32cf1786f5c)

ctruta: You have five commits in total. There should be only one. You need to rebase, squash, and then force-push to your repository, and, finally, submit a new review request in our repository.

Moreover, your final commit 61ec911899d4c3a80c033aa36c7ef32cf1786f5c alone shows `+43 -62`, so 43 line additions and 62 line deletions. That hardly counts as minimal.

sanjay20m:
I have rebased, squashed, and force-pushed the commit as requested.

ctruta:
> I have rebased, squashed, and force-pushed the commit as requested.

Thank you, but it's still not ok. You have two commits instead of one, which aren't rebased on the present-day tip of tree, and you still have deletions that are unwarranted.

If you're using an IDE, that's ok. But if you don't see your modifications in your IDE, then you should at least be able to see them right here in GitHub's web interface, as well as in the output of the `git diff` command line.

PR #685, opened 2025-05-25
BUILD BREAK: remove erroneous RISCV addition

jbowler:
configure.ac was modified to insert lines for RISCV without testing the
result.  This removes the erroneous addition which causes all builds to
fail.


bobfriesenhahn:
The last commit for configure lacked a matching commit for configure.ac and it is likely that configure was manually edited.  Regardless, it did not take many edits to configure.ac to match the additions to configure, and to fix the syntax error.  After this, 'make check' passes.

bobfriesenhahn:
Pull request #693 is proposed as an alternative to this one.

ctruta:
Closing as duplicate of #683. As I wrote in there: apologies for my delayed response, which caused you to do this extra work.

PR #693, opened 2025-05-26
Fix syntax error in configure.ac and update build files. (Issue 678)

bobfriesenhahn:
This changeset fixes the syntax error in configure.ac, reconciles differences noticed due to configure being edited rather than configure.ac, and regenerates Autotools based on Autoconf 2.72, Automake 1.17, and libtool 2.5.4.  This is a proposed alternative to pull request #685 by @jbowler.

I do not have any knowledge of RISCV, or a compilation environment for it, but it seems best to preserve contributed work as much as possible until RISCV experts pay attention again.


ctruta:
Closing as duplicate of #683. As I wrote in there: apologies for my delayed response, which caused you to do this extra work.

PR #695, opened 2025-05-30
Fix file descriptor leak in iccfrompng.c by closing input file

sanjay20m:
@ctruta  Hello again !!
The previous PR was closed due to a Git conflict or accidental mistake, and this is a minimalist, clean replacement containing only the necessary fix — no extra logic or changes.

This patch fixes a file descriptor leak in contrib/examples/iccfrompng.c. The input file pointer fp was not being closed after use, which could lead to resource exhaustion if processing multiple PNGs.
It Ensures the file is always closed after reading, improving memory/resource safety with zero behavior change otherwise
A single line — fclose(fp); — is added to ensure proper resource cleanup.

### DIFF
--- a/contrib/examples/iccfrompng.c
+++ b/contrib/examples/iccfrompng.c
@@ -155,15 +155,20 @@ extract_one_file(const char *filename)
       }
 
       else if (verbose && profile == no_profile)
-         printf("%s has no profile\n", filename);
-        fclose(fp);
+         
+          printf("%s has no profile\n", filename);
+        
+       fclose(fp);
    }
 
    else
       fprintf(stderr, "%s: could not open file\n", filename);
+    
     if (fp != NULL)
-    fclose(fp);
-   return result;
+   
+        fclose(fp);
+   
+    return result;
 }
 
 int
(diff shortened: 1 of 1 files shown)


ctruta:
> [...] this is a minimalist, clean replacement containing only the necessary fix — no extra logic or changes

Thank you, but your code is still not indented correctly.

sanjay20m:
Check the update now!! And if any tweaks are needed, please tell me how I should proceed. I would be grateful for me ..

ctruta:
@sanjay20m I can see that you are trying very hard, and I appreciate your efforts, but (sorry!) your submission is still not correct.

You (again) still have two commits instead of one, and you should have only one.

Moreover, there is an error in your code logic. You are closing the input file pointer `fp` twice, which sometimes is causing a crash. Just so you know: if you fail to close a file opened for reading (not writing), it will be automatically closed when the program terminates, and nothing will go wrong. But if you close it twice instead of once, then you're running into a double-free security vulnerability.

Therefore -- again, not that I don't appreciate your efforts, because I absolutely do! -- if you can implement a proper fix, that's ok, and if you leave it as is, that's ok also. But an incorrect "fix" is going to be actually worse than no fix at all.

ctruta:
Integrated in PR #697.

PR #697, opened 2025-05-31
Fix double fclose by ensuring fp is not NULL before closing

sanjay20m:
@ctruta Hello again, I loved your support and help, which gave me a good understanding of how the world of code works. Thanks for that.

This is a minimalist, clean fix to avoid a potential double fclose() call in iccfrompng.c.
We now ensure the file pointer fp is closed exactly once, only if it was opened.
No other logic or formatting is changed.

### DIFF
--- a/contrib/examples/iccfrompng.c
+++ b/contrib/examples/iccfrompng.c
@@ -161,6 +161,9 @@ extract_one_file(const char *filename)
    else
       fprintf(stderr, "%s: could not open file\n", filename);
 
+    if (fp != NULL)
+        fclose(fp);
+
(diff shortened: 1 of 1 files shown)


ctruta:
So, @sanjay20m, I applied your submission. See commit 197bbfe0d5955a3736cb3091cdc1c214b13341fe.

I appreciate not only your fix, but also your patience and your willingness to adapt to our code expectations :-)

sanjay20m:
@ctruta  Thanks !! a lot , I learned a so much from you (the key of accuracy) , thanks again :)

PR #702, opened 2025-06-15
fix Fallback when RVV rev. < 1.0

filipwasil:
Availability of RVV version:

- minimal Kernel revision (To do the hardware checks)
- minimal microarchitecture flag (1p0 -> RVV 1.0)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,9 +331,28 @@ if(PNG_HARDWARE_OPTIMIZATIONS)
     if(index EQUAL -1)
       message(FATAL_ERROR "PNG_RISCV_RVV must be one of [${PNG_RISCV_RVV_POSSIBLE_VALUES}]")
     elseif(NOT PNG_RISCV_RVV STREQUAL "off")
-      check_c_compiler_flag("-march=rv64gv1p0" COMPILER_SUPPORTS_RVV)
+
+      set(_SAVED_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+      set(CMAKE_REQUIRED_FLAGS "-march=rv64gv1p0")
+
+      check_c_source_compiles("
+        #include <riscv_vector.h>
+        #include <asm/hwcap.h>
+        #ifndef COMPAT_HWCAP_ISA_V /* added in linux-6.5 */
+        #error \"COMPAT_HWCAP_ISA_V is not available\"
+        #endif
+        int main() {
+          const float src[] = { 0.0f, 0.0f, 0.0f, 0.0f };
+          uint64_t ptr[2] = {0x0908060504020100, 0xFFFFFFFF0E0D0C0A};
+          vuint8m1_t a = __riscv_vreinterpret_v_u64m1_u8m1(__riscv_vle64_v_u64m1(ptr, 2));
+          vfloat32m1_t val = __riscv_vle32_v_f32m1((const float*)(src), 4);
+          return (int)__riscv_vfmv_f_s_f32m1_f32(val);
+      }" COMPILER_SUPPORTS_RVV)
+
+      set(CMAKE_REQUIRED_FLAGS ${_SAVED_CMAKE_REQUIRED_FLAGS})
+
       if(NOT COMPILER_SUPPORTS_RVV)
-        message(FATAL_ERROR "Compiler does not support -march=rv64gv1p0 option")
+        message(FATAL_ERROR "Compiler does not support RISC-V Vector extension or its unable to detect it")
       endif()
       set(libpng_riscv_sources
           riscv/filter_rvv_intrinsics.c
--- a/configure.ac
+++ b/configure.ac
@@ -700,7 +700,7 @@ AS_HELP_STRING([[[--enable-riscv-rvv]]],
          AC_DEFINE([PNG_RISCV_RVV_OPT], [2],
                    [Enable RISC-V Vector optimizations])
          AC_MSG_WARN([--enable-riscv-rvv: please specify 'check' or 'api', if]
-            [you want the optimizations unconditionally pass e.g. '-march=rv64gv']
+            [you want the optimizations unconditionally pass e.g. '-march=rv64gv1p0']
             [to the compiler.]);;
       *)
          AC_MSG_ERROR([--enable-riscv-rvv=${enable_riscv_rvv}: invalid value])
@@ -716,11 +716,19 @@ then
    AC_MSG_CHECKING(whether to use RISC-V RVV intrinsics)
 
    save_CFLAGS=$CFLAGS
-   CFLAGS="$CFLAGS -march=rv64gv"
+   CFLAGS="$CFLAGS -march=rv64gv1p0"
    AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
 #include <riscv_vector.h>
-int main(){
-   return 0;
+#include <asm/hwcap.h>
+#ifndef COMPAT_HWCAP_ISA_V /* added in linux-6.5 */
+#error "COMPAT_HWCAP_ISA_V is not available"
+#endif
+int main() {
+   const float src[] = { 0.0f, 0.0f, 0.0f, 0.0f };
+   uint64_t ptr[2] = {0x0908060504020100, 0xFFFFFFFF0E0D0C0A};
+   vuint8m1_t a = __riscv_vreinterpret_v_u64m1_u8m1(__riscv_vle64_v_u64m1(ptr, 2));
+   vfloat32m1_t val = __riscv_vle32_v_f32m1((const float*)(src), 4);
+   return (int)__riscv_vfmv_f_s_f32m1_f32(val);
 }]])],compiler_support_riscv_rvv=yes)
    AC_MSG_RESULT($compiler_support_riscv_rvv)
    if test "$compiler_support_riscv_rvv" = "yes"; then


ctruta (changes requested):
Could you please implement a fix like this to the CMake build also?

filipwasil:
@ctruta 

the availability of march flag with "1p0" postfix is already there in CMake, however, I did add the check for minimal kernel version to have this 1 to 1 CMake and autotools same behaviour. (Yes completely my fault i missed that.)

Let me also share my thought here:

If we explicitly state that we do want --enable-hardware-optimizations (which sets the --enable-riscv-rvv to yes), but we do introduce a march flag (such as march=rv64gc - lacking the "v" extension) this makes sort of undefined request as we ask to have and not allowing the v extension in the same time.

In relation to:
https://github.com/pnggroup/libpng/pull/699#issuecomment-2973437997

It seems to me that this might be the case here. What should be the behaviour in such case ?
Having more information about this particular build will be very useful.

One important thing about this particular PR is that i am not able to test it today on the physical target.
I will be able to do that tomorrow. I could look at it to detect potential cases I might have missed during my tests If i've had more build info from @Puqns67

Puqns67:
@filipwasil Sorry I'm late.

> Sorry for issues. From what I see, The StarFive JH7110 SoC, found in devices like the VisionFive 2, does not support the ratified RISC-V Vector Extension (RVV) 1.0, but instead implements the earlier, draft RVV 0.7.1.

The StarFive JH7110 SoC does not support any version of RVV (1.0, 0.7.1, XTheadVector and etc), it is `rv64imafdc_zba_zbb`.

> What build do You try ? Is this autoconfig or CMake ? Can You provide the output of the build command ? (configure and/or) CMake ?

I'm build with Gentoo's portage, it's using [`--enable-riscv-rvv=check`](https://github.com/gentoo/gentoo/blob/0a067ae96f426b8cdeb155da6baa7f7946235bc6/media-libs/libpng/libpng-1.6.49.ebuild#L65) with autoconfig.

Manual builds via autoconfig and cmake on the latest commit can also cause this issue, log: https://fars.ee/hknw

So `check` is just checking if the compiler supports the v extension, it doesn't take into account if it can be run locally, right?

Gentoo's Bugzilla: https://bugs.gentoo.org/958203

filipwasil:
> @filipwasil Sorry I'm late.
> 
> > Sorry for issues. From what I see, The StarFive JH7110 SoC, found in devices like the VisionFive 2, does not support the ratified RISC-V Vector Extension (RVV) 1.0, but instead implements the earlier, draft RVV 0.7.1.
> 
> The StarFive JH7110 SoC does not support any version of RVV (1.0, 0.7.1, XTheadVector and etc), it is `rv64imafdc_zba_zbb`.
> 
> > What build do You try ? Is this autoconfig or CMake ? Can You provide the output of the build command ? (configure and/or) CMake ?
> 
> I'm build with Gentoo's portage, it's using [`--enable-riscv-rvv=check`](https://github.com/gentoo/gentoo/blob/0a067ae96f426b8cdeb155da6baa7f7946235bc6/media-libs/libpng/libpng-1.6.49.ebuild#L65) with autoconfig.
> 
> Manual builds via autoconfig and cmake on the latest commit can also cause this issue, log: https://fars.ee/hknw
> 
> So `check` is just checking if the compiler supports the v extension, it doesn't take into account if it can be run locally, right?
> 
> Gentoo's Bugzilla: https://bugs.gentoo.org/958203

The "check" means a "runtime check for v extension". Its there to be used when You are building on RVV capable machine , or RVV capable toolchain/sdk. Its there to run the code on a machine that might not have such support.

There are couple of things I noticed when looking at CMake logs. First, let me bring here the build time check we do.
After my last commit in this PR the code is

```
      set(_SAVED_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
      set(CMAKE_REQUIRED_FLAGS "-march=rv64gv1p0")

      check_c_source_compiles("
        #include <riscv_vector.h>
        #include <asm/hwcap.h>
        #ifndef COMPAT_HWCAP_ISA_V /* added in linux-6.5 */
        #error \"COMPAT_HWCAP_ISA_V is not available\"
        #endif
        int main() {
          return 0;
      }" COMPILER_SUPPORTS_RVV)

      set(CMAKE_REQUIRED_FLAGS ${_SAVED_CMAKE_REQUIRED_FLAGS})
```

This code makes the RVV 1.0 mandatory and builds piece of code. This code will not compile if the build flags are not supported or runtime check is not supported by kernel.

The CMake tells me:
```
...
-- Performing Test COMPILER_SUPPORTS_RVV
-- Performing Test COMPILER_SUPPORTS_RVV - Success
...
```

Which means that the above chunk compiles fine. What happens next is the implication - build process now has the COMPILER_SUPPORTS_RVV flag set and it acts accordingly.

Also it does not return any issues from header. By its own, doing this:
```
#include <riscv_vector.h>
```

This will fail because on RVV non-capable hw will return error from inside of this header complaining about missing v extension in march definition.

In Your case compiler is happy to accept the march with v extension and happy to compile this piece of code.
It is possible to improve it though. Lets give it vector types, and functions. Apparently they are needed in some cases as well.

```
check_c_source_compiles("
   #include <riscv_vector.h>
   #include <asm/hwcap.h>
   #ifndef COMPAT_HWCAP_ISA_V /* added in linux-6.5 */
   #error \"COMPAT_HWCAP_ISA_V is not available\"
   #endif
  int main()
  {
    const float src[] = { 0.0f, 0.0f, 0.0f, 0.0f };
    uint64_t ptr[2] = {0x0908060504020100, 0xFFFFFFFF0E0D0C0A};
    vuint8m1_t a = vreinterpret_v_u64m1_u8m1(vle64_v_u64m1(ptr, 2));
    vfloat32m1_t val = vle32_v_f32m1((const float*)(src), 4);
    vfloat32m1x4_t tmp;
    return (int)vfmv_f_s_f32m1_f32(val);
  }" COMPILER_SUPPORTS_RVV)
```

It might be a bug in GCC or more precisely a bug in the code related to this architecture where the compiler should not detect that its rvv capable, but it allows too much without failing the compilation.

Regardless. I will update the PR with improved check.

filipwasil:
There are four scenarios if You want to do PNG_RISCV_RVV=check

### 1. Scenario when You do not allow the v extension, but the machine does support it, and You do "check" build:
command
```
CC=gcc-14 CXX=g++-14 CFLAGS="-march=rv64gc" cmake -B build_riscv_cmake_test -D PNG_RISCV_RVV=check -G Ninja
```
compilation result
```
/root/Projects/libpng/riscv/filter_rvv_intrinsics.c:22:1: error: this operation requires the RVV ISA extension
```

### 2. Scenario when You do not allow the v extension, and the machine does not support it, and You do "PNG_RISCV_RVV=check"

command
```
CC=gcc-14 CXX=g++-14 CFLAGS="-march=rv64gc" cmake -B build_riscv_cmake_test -D PNG_RISCV_RVV=check -G Ninja

CC=gcc-14 CXX=g++-14 ./configure --host=riscv64-linux-gnu --disable-shared --enable-hardware-optimizations CFLAGS="-march=rv64gc" --enable-riscv-rvv=check
command
```
compilation result
```
CMake Error at CMakeLists.txt:356 (message):
  Compiler does not support RISC-V Vector extension or its unable to detect
  it
```

### 3. Scenario when You do have the support, You do introduce march a flag with "v", and you do "PNG_RISCV_RVV=check"
command
```
CC=gcc-14 CXX=g++-14 CFLAGS="-march=rv64gvc" cmake -B build_riscv_cmake_test -D PNG_RISCV_RVV=check -G Ninja

CC=gcc-14 CXX=g++-14 ./configure --host=riscv64-linux-gnu --enable-hardware-optimizations CFLAGS="-march=rv64gvc" --enable-riscv-rvv=check
```
build result
```
builds fine
```


### 4. Scenario when You do have the support, You do not introduce march flag, and you do "PNG_RISCV_RVV=check" (aka default behaviour for passing "PNG_RISCV_RVV=check" )
```
CC=gcc-14 CXX=g++-14 cmake -B build_riscv_cmake_test -D PNG_RISCV_RVV=check -G Ninja

CC=gcc-14 CXX=g++-14 ./configure --host=riscv64-linux-gnu --enable-hardware-optimizations --enable-riscv-rvv=check
```
build result
```
/root/Projects/libpng/riscv/filter_rvv_intrinsics.c:22:1: error: this operation requires the RVV ISA extension
```

Your compiler acts like in scenario 1:
- It does detect the v support on CMake/autotools execution level
- It does not find the proper "v" in march when building
- It tries to compile runtime check, and runtime rvv execution path

This can not work. When You do PNG_RISCV_RVV=check, You have to be able to build the mechanism to detect and use the v extension.

You **should** get error (But You should get error 2., not the error 1) because - again - You want o do **check**, but You are not able to compile code that is required if the runtime check succeeds.

This PR introduced better way to prevent such issues (old compiler, old kernel, compiler allowing including <riscv_vector> when it should not, and accepting march which it should not accept). I find this case interesting, and it seems that the detection of this specific march has flows.

Hope this helps

ctruta:
@filipwasil I gave it a spin. The CMake build worked, but the configure build failed:
```
/usr/lib/gcc-cross/riscv64-linux-gnu/13/../../../../riscv64-linux-gnu/bin/ld: ./.libs/libpng16.a(pngrutil.o): in function `png_combine_row':
/home/cosmin/src/pngdev/worktrees/libpng16/pngrutil.c:3570:(.text+0x41b8): undefined reference to `png_init_filter_functions_rvv'
```

The `riscv/` subdirectory inside the build dir is empty. When I roll back to the tag `v1.6.49`, the build works again, so it's clearly a regression on my machine.

BTW, as you rework your commits, could you please squash them into a single fix that I can then apply it cleanly? Or, if you prefer, a commit to fix the configure build issue, and another to fix the CMake build issue, that would also work, if they are, indeed, separate fixes to separate issues.

filipwasil:
The thing is GCC 13 -> it can not compile this line:
```
 vfloat32m1x4_t tmp;
```
which by itself its not entirely wrong - we may want to do fallback when we have old compiler.
What **is** wrong though, is the way I do the callback. I exclude riscv implementation from build, but keep the part that is using it causing a linking error. I am on it.

ctruta:
So I landed this, with the important caveat that I only tested the compilation, without the execution. Later today I will see if `qemu-system-riscv` can emulate RVV versions prior to 1.0.

BTW: I know very little about the RISC-V ecosystem. Are the pre-1.0 machines widespread? Should we support those? I mean, we can fall back to non-vectorized code at any time, for any platform. The same is true for x86: if SSE2 is not available, we used to fall back to MMX, but that was deemed to not be such a high benefit, so now the libpng code acts as if there's no vectorization at all. (Sorry about that, AMD Duron users!...)

filipwasil:
> So I landed this, with the important caveat that I only tested the compilation, without the execution. Later today I will see if `qemu-system-riscv` can emulate RVV versions prior to 1.0.
> 
> BTW: I know very little about the RISC-V ecosystem. Are the pre-1.0 machines widespread? Should we support those? I mean, we can fall back to non-vectorized code at any time, for any platform. The same is true for x86: if SSE2 is not available, we used to fall back to MMX, but that was deemed to not be such a high benefit, so now the libpng code acts as if there's no vectorization at all. (Sorry about that, AMD Duron users!...)

there are few things to mention.

Compiler which we consider old in terms of RVV is for example GCC < 14 because of lack of full intrinsic support.
(some types and functions not defined) or small bugs (GCC 14 has much better support in this regard).

When it comes to versioning:

- The modern stable version of RVV is 1.0
- The modern profile for ISA for hardware designers is here [RVA23](https://lf-riscv.atlassian.net/wiki/spaces/HOME/pages/16154769/RISC-V+Technical+Specifications#ISA-Specifications)
- And finall the C intrinsics API (Version 0.12 refers to the lates intrinsics API. There is a mentioned difference in version 13 and 14 in GCC in this regard).

RVV 1.0 is the first official ratified version which which is in fact a solid base that will be only extended further in future.

The RVV code in libpng is asm code which can work with GCC13 as it does not have to rely on C API to vectorized instruction (aka. intrinsics). That's why it makes sense to support it. 

Situation which is very much unlikely but still possible is that there will be a machine which is supported by the old compiler, which will physically have bug in processor instructions itself where one of the vector instructions will be broken. Other than that - Situations such as:
- no vector extension available
- no way to detect the v extension
- march flag rv64gv1p0 not supported

are all covered

https://github.com/pnggroup/libpng/pull/704/files
I have found the fallback issue. Tests ongoing.

filipwasil:
@Puqns67 
https://github.com/pnggroup/libpng/pull/704

FYI
This improves the fallback mechanism.

filipwasil:
This one https://github.com/pnggroup/libpng/issues/703 is caused by the wrong fallback which should be fixed with this PR
https://github.com/pnggroup/libpng/pull/704

Can You try to build the Debian, or give me some hints on that ? Just to be on the same page. Should work now.

filipwasil:
By: "Should work" I mean:

1. The RVV support will be compiled in as long as check_c_source_compiles (CMake) or AC_COMPILE_IFELSE (autotools) succeeds
2. AND The user asks for the vectorized build explicitly

filipwasil:
https://github.com/pnggroup/libpng/issues/703#issuecomment-2980252732

For CMake We do throw a fatal error for PNG_LOONGARCH_LSX and for PNG_RISCV_RVV.
For autotools we do not do it.
For other architectures we do not do it

PR #704, opened 2025-06-17
Separated build flow for autotools and cmake

filipwasil:
Undefined PNG_RISCV_RVV_OPT should always mean 0 regardless of the case for RISC-V builds
If we do have "check" build then we want the PNG_RISCV_RVV_OPT  to equal 1.

Now, in case of fallback mechanism is not going to set the PNG_RISCV_RVV_OPT  to 1 and will not cause the linking errors with autotools builds.

Tested with Banana Pi F3:
```
./ci/ci_verify_configure.sh
./ci/ci_verify_cmake.sh
```


pngpriv.h
```diff
@@ -154,9 +154,7 @@
     * to configure or put -DPNG_RISCV_RVV_OPT=2 in CPPFLAGS.
     */
 
-#  if defined(__riscv) && defined(PNG_ALIGNED_MEMORY_SUPPORTED)
-#     define PNG_RISCV_RVV_OPT 1
-#  else
+#  if defined(__riscv)
```
ctruta: Typo in the line below.

filipwasil: Ach, It should be simply:
```
#  define PNG_RISCV_RVV_OPT 0
```
yep,

btw. github apparently didn't update the PR instnatly when I updated the feature branch.

filipwasil: Otherwords: If You were refering that the ifdef was not needed - Yes. Any other typo ? I do not see it. Maybe github is refering to old version of that file, before I did the force push (To not generate another commit).

ctruta: > btw. github apparently didn't update the PR instnatly when I updated the feature branch.

Ha! I was suspecting something odd happening.

I almost always apply the PRs from the command line (not in the web interface), except, in this case, your branch appeared to be missing. I don't mean force-pushed, but rather, renamed, or deleted, or something else. So I took the commit hash, instead of the branch name, and that's what I reviewed.

> Otherwords: If You were refering that the ifdef was not needed - Yes. Any other typo ? I do not see it.

I had one thing in mind, and I wrote another, but I'm glad that all converged well in the end.

> Maybe github is refering to old version of that file, before I did the force push (To not generate another commit).

Dunno. Sometimes GitHub baffles me, too...

PR #706, opened 2025-06-20
APNG — let's do this!

ctruta:


ctruta:
@unitarou, @ProgramMax, @svgeesus, @jbowler, @peterkaczorowski:
 Please review :grin:

peterkaczorowski:
Wonderful! A merge with independent APNG support has just been prepared for GiMP, but the team is very willing to prepare a version compatible with the new LIBPNG. I’ll resume work on wxWidgets and my extensions for KiCad. 

Great news. Thank you, Cosmin!

ctruta:
> Wonderful! A merge with independent APNG support has just been prepared for GiMP, but the team is very willing to prepare a version compatible with the new LIBPNG. I’ll resume work on wxWidgets and my extensions for KiCad.

It's sooo nice to see the enthusiasm, @peterkaczorowski :-)

Mind you, this comes with the important caveat that we are in the earliest alpha stage of libpng v2. This PR is going to be literally the first feature added, and it will still take a while until this brand new libpng thingy becomes ready for prime time.

*****

I added you to the list of reviewers, but please spread the word over to those projects, too: anybody who wants to contribute a review is hereby invited to contribute a review.

pngrutil.c
```diff
@@ -4689,4 +4906,70 @@ defined(PNG_USER_TRANSFORM_PTR_SUPPORTED)
...
+   png_ptr->pass = 0;
+}
+
+void /* PRIVATE */
+png_read_reinit(png_structp png_ptr, png_infop info_ptr)
+{
+   png_ptr->width = info_ptr->next_frame_width;
+   png_ptr->height = info_ptr->next_frame_height;
+   png_ptr->rowbytes = PNG_ROWBYTES(png_ptr->pixel_depth,png_ptr->width);
+   png_ptr->info_rowbytes =
```
MaartenBent, line 4930: Hi. I evaluated this apng patch (for libpng 1.6.46) in the past, for use in wxWidgets.
Some apng files I found on the [internet](http://littlesvr.ca/apng/samples.html) failed to load with an `internal row size calculation error`.

ball
clock

This patch fixed it. I think it is still applicable to this version.

```patch
diff --git "a/pngrutil.c" "b/pngrutil.c"
index 9eebf9e41..2fa59455d 100644
--- "a/pngrutil.c"
+++ "b/pngrutil.c"
@@ -4936,8 +4936,9 @@ png_read_reinit(png_structp png_ptr, png_infop info_ptr)
     png_ptr->width = info_ptr->next_frame_width;
     png_ptr->height = info_ptr->next_frame_height;
     png_ptr->rowbytes = PNG_ROWBYTES(png_ptr->pixel_depth,png_ptr->width);
-    png_ptr->info_rowbytes = PNG_ROWBYTES(info_ptr->pixel_depth,
-        png_ptr->width);
+    if (png_ptr->info_rowbytes != 0)
+        png_ptr->info_rowbytes = PNG_ROWBYTES(info_ptr->pixel_depth,
+            png_ptr->width);
     if (png_ptr->prev_row)
         memset(png_ptr->prev_row, 0, png_ptr->rowbytes + 1);
 }
```

svgeesus: > Some apng files I found on the [internet](http://littlesvr.ca/apng/samples.html) failed to load with an `internal row size calculation error`

Just confirming this is a valid PNG:

```
chris@SuperNomad:/mnt/c/Users/chris/Documents/PNG/APNG samples$ pngcheck -c -v 457508952-618918e3-c770-4e7b-919d-85a9f5085fbf.png
File: 457508952-618918e3-c770-4e7b-919d-85a9f5085fbf.png (23680 bytes)
  chunk IHDR at offset 0x0000c, length 13
    150 x 150 image, 8-bit palette, non-interlaced
  chunk acTL at offset 0x00025, length 8
    Animated PNG, 40 frames, plays continuously
  chunk PLTE at offset 0x00039, length 708: 236 palette entries
  chunk tRNS at offset 0x00309, length 17: 17 transparency entries
  chunk fcTL at offset 0x00326, length 26
    Frame, sequence number 0
    Width 150, height 150 starting at (0, 0)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame overwrites buffer
  chunk IDAT at offset 0x0034c, length 2003
    zlib: deflated, 32K window, fast compression
  chunk fcTL at offset 0x00b2b, length 26
    Frame, sequence number 1
    Width 70, height 58 starting at (48, 31)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x00b51, length 546
    Frame data, sequence number 2
  chunk fcTL at offset 0x00d7f, length 26
    Frame, sequence number 3
    Width 72, height 57 starting at (46, 32)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x00da5, length 543
    Frame data, sequence number 4
  chunk fcTL at offset 0x00fd0, length 26
    Frame, sequence number 5
    Width 73, height 55 starting at (46, 33)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x00ff6, length 540
    Frame data, sequence number 6
  chunk fcTL at offset 0x0121e, length 26
    Frame, sequence number 7
    Width 73, height 52 starting at (46, 35)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x01244, length 562
    Frame data, sequence number 8
  chunk fcTL at offset 0x01482, length 26
    Frame, sequence number 9
    Width 68, height 48 starting at (51, 38)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x014a8, length 381
    Frame data, sequence number 10
  chunk fcTL at offset 0x01631, length 26
    Frame, sequence number 11
    Width 71, height 43 starting at (48, 42)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x01657, length 481
    Frame data, sequence number 12
  chunk fcTL at offset 0x01844, length 26
    Frame, sequence number 13
    Width 73, height 37 starting at (46, 47)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x0186a, length 434
    Frame data, sequence number 14
  chunk fcTL at offset 0x01a28, length 26
    Frame, sequence number 15
    Width 72, height 31 starting at (47, 52)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x01a4e, length 426
    Frame data, sequence number 16
  chunk fcTL at offset 0x01c04, length 26
    Frame, sequence number 17
    Width 73, height 29 starting at (46, 54)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x01c2a, length 432
    Frame data, sequence number 18
  chunk fcTL at offset 0x01de6, length 26
    Frame, sequence number 19
    Width 73, height 29 starting at (46, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x01e0c, length 398
    Frame data, sequence number 20
  chunk fcTL at offset 0x01fa6, length 26
    Frame, sequence number 21
    Width 70, height 34 starting at (49, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x01fcc, length 443
    Frame data, sequence number 22
  chunk fcTL at offset 0x02193, length 26
    Frame, sequence number 23
    Width 67, height 37 starting at (52, 55)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x021b9, length 464
    Frame data, sequence number 24
  chunk fcTL at offset 0x02395, length 26
    Frame, sequence number 25
    Width 73, height 44 starting at (46, 54)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x023bb, length 540
    Frame data, sequence number 26
  chunk fcTL at offset 0x025e3, length 26
    Frame, sequence number 27
    Width 72, height 50 starting at (47, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x02609, length 502
    Frame data, sequence number 28
  chunk fcTL at offset 0x0280b, length 26
    Frame, sequence number 29
    Width 72, height 53 starting at (47, 55)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x02831, length 403
    Frame data, sequence number 30
  chunk fcTL at offset 0x029d0, length 26
    Frame, sequence number 31
    Width 73, height 58 starting at (46, 54)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x029f6, length 536
    Frame data, sequence number 32
  chunk fcTL at offset 0x02c1a, length 26
    Frame, sequence number 33
    Width 70, height 60 starting at (49, 55)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x02c40, length 513
    Frame data, sequence number 34
  chunk fcTL at offset 0x02e4d, length 26
    Frame, sequence number 35
    Width 73, height 64 starting at (46, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x02e73, length 577
    Frame data, sequence number 36
  chunk fcTL at offset 0x030c0, length 26
    Frame, sequence number 37
    Width 73, height 65 starting at (46, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x030e6, length 541
    Frame data, sequence number 38
  chunk fcTL at offset 0x0330f, length 26
    Frame, sequence number 39
    Width 73, height 64 starting at (46, 55)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x03335, length 390
    Frame data, sequence number 40
  chunk fcTL at offset 0x034c7, length 26
    Frame, sequence number 41
    Width 73, height 66 starting at (46, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x034ed, length 540
    Frame data, sequence number 42
  chunk fcTL at offset 0x03715, length 26
    Frame, sequence number 43
    Width 73, height 62 starting at (46, 56)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x0373b, length 519
    Frame data, sequence number 44
  chunk fcTL at offset 0x0394e, length 26
    Frame, sequence number 45
    Width 73, height 64 starting at (46, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x03974, length 486
    Frame data, sequence number 46
  chunk fcTL at offset 0x03b66, length 26
    Frame, sequence number 47
    Width 72, height 62 starting at (47, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x03b8c, length 531
    Frame data, sequence number 48
  chunk fcTL at offset 0x03dab, length 26
    Frame, sequence number 49
    Width 77, height 59 starting at (42, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x03dd1, length 425
    Frame data, sequence number 50
  chunk fcTL at offset 0x03f86, length 26
    Frame, sequence number 51
    Width 81, height 54 starting at (38, 54)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x03fac, length 482
    Frame data, sequence number 52
  chunk fcTL at offset 0x0419a, length 26
    Frame, sequence number 53
    Width 85, height 50 starting at (35, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x041c0, length 482
    Frame data, sequence number 54
  chunk fcTL at offset 0x043ae, length 26
    Frame, sequence number 55
    Width 87, height 45 starting at (33, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x043d4, length 501
    Frame data, sequence number 56
  chunk fcTL at offset 0x045d5, length 26
    Frame, sequence number 57
    Width 87, height 37 starting at (32, 55)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x045fb, length 406
    Frame data, sequence number 58
  chunk fcTL at offset 0x0479d, length 26
    Frame, sequence number 59
    Width 88, height 33 starting at (31, 54)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x047c3, length 366
    Frame data, sequence number 60
  chunk fcTL at offset 0x0493d, length 26
    Frame, sequence number 61
    Width 89, height 28 starting at (31, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x04963, length 475
    Frame data, sequence number 62
  chunk fcTL at offset 0x04b4a, length 26
    Frame, sequence number 63
    Width 88, height 30 starting at (32, 53)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x04b70, length 481
    Frame data, sequence number 64
  chunk fcTL at offset 0x04d5d, length 26
    Frame, sequence number 65
    Width 87, height 30 starting at (33, 52)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x04d83, length 518
    Frame data, sequence number 66
  chunk fcTL at offset 0x04f95, length 26
    Frame, sequence number 67
    Width 85, height 37 starting at (35, 47)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x04fbb, length 527
    Frame data, sequence number 68
  chunk fcTL at offset 0x051d6, length 26
    Frame, sequence number 69
    Width 81, height 43 starting at (39, 42)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x051fc, length 468
    Frame data, sequence number 70
  chunk fcTL at offset 0x053dc, length 26
    Frame, sequence number 71
    Width 78, height 47 starting at (42, 39)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x05402, length 522
    Frame data, sequence number 72
  chunk fcTL at offset 0x05618, length 26
    Frame, sequence number 73
    Width 74, height 52 starting at (46, 35)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x0563e, length 444
    Frame data, sequence number 74
  chunk fcTL at offset 0x05806, length 26
    Frame, sequence number 75
    Width 72, height 55 starting at (48, 33)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x0582c, length 510
    Frame data, sequence number 76
  chunk fcTL at offset 0x05a36, length 26
    Frame, sequence number 77
    Width 74, height 57 starting at (46, 32)
    Frame delay 0 (4 / 100) sec
    No disposal before next frame
    Frame composites (source over) with buffer
  chunk fdAT at offset 0x05a5c, length 528
    Frame data, sequence number 78
  chunk IEND at offset 0x05c78, length 0
No errors detected in 457508952-618918e3-c770-4e7b-919d-85a9f5085fbf.png (85 chunks, -5.2% compression).
```

jbowler: Ok, found it; the one at libpng1.8:pngrutil.c:3272 and libpng2.0:pngrutil.c:3455  I think there is something dubious about the handling of png_info for the frame.  It looks like the code isn't updating png_info with the width/height of the new frame which will cause the patched line to fail if a frame is not the same height as the previous one (or the image...)

MaartenBent:
yes, this one:
```c++
   if (png_ptr->info_rowbytes != 0 && png_ptr->info_rowbytes !=
          PNG_ROWBYTES(pixel_depth, row_width))
      png_error(png_ptr, "internal row size calculation error");
```
From what I can see, `info_rowbytes` is only set in the `png_read_transform_info()` function. And in the pngs I tested, this is never called.
So I am not sure why apng updates the `info_rowbytes` value in the `png_read_reinit()` function.
Thats why I suggested a patch to only update it if it was not 0.

pngrutil.c
MaartenBent: yes, this one:
```c++
   if (png_ptr->info_rowbytes != 0 && png_ptr->info_rowbytes !=
          PNG_ROWBYTES(pixel_depth, row_width))
      png_error(png_ptr, "internal row size calculation error");
```
From what I can see, `info_rowbytes` is only set in the `png_read_transform_info()` function. And in the pngs I tested, this is never called.
So I am not sure why apng updates the `info_rowbytes` value in the `png_read_reinit()` function.
Thats why I suggested a patch to only update it if it was not 0.

(note I'm using libpng 1.6, so things might have changed)

<sub>put this comment at the correct place</sub>

jbowler: Nothing has changed in 1.8 or 2.0 (which are identical apart from the APNG stuff and a couple of minor changes).  The only way to get png_info updated to match the output format is to explicitly call png_read_update_info; png_struct has no record of the png_info (apart from one potentially dangerous design error).

The problem is that an app may use the row information from png_info **after** calling png_read_update_info and if it is **wrong** libpng will probably (50% of the time) crash by overwriting the end of the application allocated row; in effect libpng gave one answer when png_read_update_info was called but used another result when it wrote the output to the app provided buffer.  This happened all the time which is why we added the info_rowbytes field in 1.5.4 :-)

In your test PNG the frames have different widths and heights so, yes, the number of bytes in the output rows will certainly change from frame to frame.  I guess that is why the reinit function changes it because I'm pretty sure the **transforms** can't be changed between IDAT and fcTL images (but does the code check for the client doing that?)

This is why I used the word "dubious" :-)

jbowler:
@ctruta: this should be checked in already shouldn't it?  My understanding of the git development process is that a branch is created for a set of changes ("develop" in this case) and then changes are submitted and checked in.  To start the process you have to actually check the changes in to the branch...

pngrutil.c
jbowler: @MaartenBent I think your patch is good but it's clear that at least one quarter of the common cases used by application writers haven't been tested by either you or @ctruta (or other APNG developers).

ANALYSIS
=======
Here's the calculation in pngrtran.c.  As you observed this code is not called by libpng, its only executed if the app programmer calls the API `png_read_update_info`:

https://github.com/pnggroup/libpng/blob/0c8c79e77ca7234abcc6493781cc568d9fea44ed/pngrtran.c#L2192

(That's the unpatched develop branch.)    This is the value that is copied into `png_ptr->info_rowbytes` (see the next lines) and it is also the value returned by `png_get_rowbytes()`, the API accessor.   The value is initialized by `png_set_IHDR` along with the very important `info_ptr->pixel_depth` to the values for the original PNG image.  `info_ptr->pixel_depth` is also **updated** by `png_read_update_info` inside `png_read_transform_info` (the code line before the above link).

Here's the code from #706 from the new `png_read_reinit`:

https://github.com/ctruta/libpng/blob/a2590c3c7d93e5a19cb54056444b44f3c68c8c83/pngrutil.c#L4931

I.e:

```
   png_ptr->info_rowbytes =
      PNG_ROWBYTES(info_ptr->pixel_depth, png_ptr->width);
```

So that correctly updates `info_rowbytes` **only** if the app has called `png_read_update_info`.  Your change should work, as would several other tests (but I think your change is the clearest).  e.g:

`if (png_ptr->pixel_depth != info_ptr->pixel_depth)`

**However** I don't think your change is sufficient on its own.  In particular it seems that `png_get_rowbytes` isn't going to return the right answer unless `info_ptr->rowbytes` is also updated.  I can't see that happening.

CONCLUSION
=========
@ctruta: you need this change but you need to also to verify that app writers can get the rowbytes for the "next" frame.

One thing to think about is whether the frame iterator should be inside `png_info`; that's weird.  In IDAT handling **all** the state information is in `png_struct` and `png_info` is not available.  It's a very confusing API inconsistency particularly given that separate `png_info` structures can be used before and after IDAT - that won't work big-time.

jbowler:
@ctruta: The changes use `png_structp` and `png_infop`.  The rest of libpng uses `restrict` points (`png_structrp` and `png_inforp`).   The changes also don't use `png_const_structrp` and `png_const_inforp`.  The former affects `png_get_next_frame_fcTL` where the compiler cannot know that the return arguments are not aliases into `png_info`, the latter affects pretty much everything.

This is a point where consistency is important.  Just change all the `png_structp` and `png_infop` occurences in the new code to png_const...rp and then back-fix the ones that can't be const.

jbowler:
There's no handling for out-of-order or damaged sequence numbers.  Better than what is being done move the sequence number iterator on **read** into the app space, then it can choose how to handle the error.

It is also better to allow the frames to be saved.  As it is an editor can't fix up a damaged APNG using libpng.

In any case `png_ensure_sequence_number` just calls png_error; that doesn't work!  It needs to call `png_chunk_benign_error` followed or preceded by `png_crc_finish` otherwise an ancillary chunk will make a PNG completely unreadable by libpng!  Think about what happens with an app that doesn't handle APNG.  If an APNG has a bogus fcTL before IDAT (sequence number not 0) the whole PNG will be undisplayable; there's no way of skipping the bogus **and** unused chunk.  PNGs that can currently be displayed in 1.6 will suddenly not be displayable in 2.0

jbowler:
I did some tests with `clock.png` (above, I got it from the link)  I didn't apply @MaartenBent's patch but the programs don't call the new APIs.  These programs work fine when linked against 1.6.49 but both `pngcp` and `pngimage` now error out with:

`clock.png: error(libpng): write: Not enough frames written`

This is because the `acTL` is being read into the png_info but the fcTL and fdAT chunks are, apparently, never read and neither are they added to the unknown chunks (at least in the `pngcp` case.)  Basically two programs that used to work with 1.6.49 now actually error out (leaving a truncated PNG) in 2.0.0.

It's standard, and unfortunate, that adding "known" chunks breaks programs that rely on the unknown handling; the chunks are "known" by libpng but not the app so apps that use png_info from **read** to **write** a PNG can end up doing the wrong thing with the new chunks.

`pngcrush` correctly removes the "unsafe to copy" acTL along with all the fcTL and fdAT chunks.

I strongly recommend disabling APNG handling **by default**.  It's no significant work for an app to call an "enable APNG" API (e.g. png_set_option) since it has to call all the complex "next_frame" APIs and allocate buffers for the frames anyway.  If you don't like that it could be enabled automagically when the app calls png_get_acTL but that's a little more complicated internally.

jbowler:
CDfcTL in pngrutil.c is wrong (not a result of this change): "25U" should be "26U"; it's a typo and it is in 1.6 as well.  The "handle" functions for those two chunks should just invoke the "unknown" handling; this will discard them by default.  This will avoid the warning/error if the sequence number is wrong; if the chunks are being skipped why issue a stream of warnings?  I've already made the point out png_error'ing out on an ancillary chunk which is not used.

jbowler:
I hacked the first fcTL in "clock.png" to have an invalid sequence number (1) then tried pngcrush again; remember pngcrush correctly removes acTL, fcTL and fdAT.  pngcrush errors out with "Out-of-order sequence number in fcTL or fdAT".

**This** is a major bug; it's an ancillary chunk and the app isn't going to use it.  Preventing it being read by, I think, **any** app that doesn't have special handling for the APNG chunks is wrong.  It's a png_chunk_benign_error, just like having a bad length hIST or, indeed, tRNS.

**AND** Chrome displays it correctly (as a static image).  Just use TweakPNG to change the first fcTL sequence number (I don't know if clock.png is copyright so I won't post the hacked version.)

jbowler:
BUG (serious): the `chunks_to_ignore` table in pngset.c (`png_set_keep_unknown_chunks`) needs to be updated to include the new known chunks, acTL, fcTL and fdAT.  See the comment above the table: This table **must** be updated every time a new ancillary chunk is added.

Fixing this will probably unbork a number of apps that would otherwise be borked.  I'm working on a separate PR for pngcp that will, necessarily, include the change but it can be done anyway by anyone; the table is stored alphabetically.

jbowler:
#710 fixes both the pngcp handling of the Apple banding chunk (this was wrong in 1.6 as well) and the handling of APNG when #706 is applied.  The contrib/tools/pngcp.c fix can be applied to 1.6 (with adjustment for the different C dialect), the pngset.c change is not required though it is harmless (APNG chunks are not "known" in 1.6).

The pngset.c change is **required** iff APNG support is added to 2.0 but is harmless regardless; every known chunk **must** be added to the list in pngset.c!

jbowler:
**Spurious and confusing code**

In `pngread.c` an "else if" mess has been added for acTL, fcTL and fdAT.  **This should not have been done** because the code falls through to png_handle_chunk at the end (the final else clause, immediately after the if-else mess) and this has entries for acTL, fcTL and fdAT already; these (table entries) can no longer be reached.

In `pngpread.c` the same thing happens for png_handle_{acTL,fcTL} but the code is written in a way which is just plain confusing; instead of using the macro `PNG_PUSH_SAVE_BUFFER_IF_FULL` like everything else in the relevant function it hand expands it inline in the code!

This is just plain confusing, even more so because png_handle_{acTL,fcTL and fdAT} are defined in pngrutil.c but not put in the table. 

Note that is no **other** png_handle_<specific chunk> function in the whole of libpng that is not `static` in pngrutil.c; they existed at one point in 1.6 but they have all been eliminated now.   Reintroducing such functions with a different type is really confusing.

fdAT handling needs to be consistent with IDAT.  There are only two things that can happen; process it in the same way as IDAT or treat it as unknown.  Hand crafting a completely new way of doing this doesn't make any sense to me; it's already done for IDAT so do it the same way!

acTL and fcTL should not need special treatment.  Repeating checks in png_handle_{acTL,fcTL} that are already in png_handle_chunk is unnecessary and confusing code duplication.

jbowler:
**Test case**:  Take "clock.png" and use TweakPNG to add a tEXt chunk before the third frame fcTL (sequence number 3, the second fcTL after IDAT).  The resultant APNG animates fine with Chrome and is to spec, however the code around 4406 of pngrutil.c "ignores" the chunk and skips it, with a warning which should certainly not be there because the APNG is valid.

Those chunks need to be `png_handle_chunk`ed so that they can be put into a png_info.  Since the code sequence (png_read_IDAT_data) can only be reached from png_read_row and that is an API call some way is needed to get a png_info in for stuff encountered between frames.

cf my comment elsewhere on the KiCAD issue of needing labels on frames.  I don't think I suggested anything but I'm thinking:

`laBL <sequence number> <UTF8-text>`

So `<sequence number>` matches the fcTL and, ideally, laBL is written directly before the corresponding fcTL.  The same issue exists with respect to "banding" of frames.  Having the encoder build a complete banding table with file offsets for not just rows in IDAT but every frame is unreasonable; it's better to allow baND chunks before each fcTL as they are unlikely to be needed until the previous frames have been decoded (though I have other ways round this).

Anyway, here are the conditions for the code:

>APNG is designed to allow incremental display of frames before the entire [datastream](https://www.w3.org/TR/png-3/#5DataRep) has been read. This implies that some errors may not be detected until partway through the animation. _It is strongly recommended that when any error is encountered decoders should discard all subsequent frames, stop the animation, and revert to displaying the static image_. _A decoder which detects an error before the animation has started should display the static image_. An error message may be displayed to the user if appropriate.
>
>Decoders shall treat out-of-order APNG chunks as an error. _APNG-aware [PNG editors](https://www.w3.org/TR/png-3/#dfn-png-editor) should restore them to correct order, using the sequence numbers_.

_Three_ of those recommendations are simply unimplementable with the current PR:

>A decoder which detects an error **before the animation** has started should display the static image.

So my bogus fcTL before IDAT must be fixed.

>It is strongly recommended that when any error is encountered decoders should discard all subsequent frames, stop the animation, and revert to displaying the static image.

So png_error is simply unacceptable.

>APNG-aware [PNG editors](https://www.w3.org/TR/png-3/#dfn-png-editor) should restore them to correct order, using the sequence numbers.

So an editor needs to be able to extract out-of-order frames.  While this can be done in the manner I demonstrated with AREATK that approach does not handle decoding of individual frames; the editor would have to write a new APNG first, possibly dropping incomplete frames.

The AREATK solution was to read the whole APNG then just get the APNG chunks in order.  That would work but it has to load the whole APNG first.  An alternative is to buffer fcTL/fdAT in the unknown list then, when the next fcTL is found start decoding of the **previous** frame.  If the fdAT are incomplete by the end of the APNG then the frame could be discarded or played with, most likely, only part of it rendering as the app preferred.

jbowler:
I've merged all the PRs I've submitted into:

https://github.com/jbowler/libpng/tree/libpng2-apng

I added @MaartenBent 's fix for png_read_reinit when png_read_update_info is not called along with the fix for png_get_rowbytes in the contrary case where it is called.

jbowler:
I reviewed the png_get_ functions, the png_get_next_frame... accessors are pure code bloat; they aren't necessary because they just duplicate the code in png_get_next_frame_fcTL and the caller needs all of the data sometime.  Nothing else in PNG does this; the non-chunk accessors pull data out of png_struct, not png_info and it is weirdly confusing to have undocumented API duplication like this; which one should be used?

jbowler:
png_set_... these should be (void) functions and need to call png_chunk_report.  The error messages are inconsistently wordy; libpng lacks any i18n and the checks need to be concise and understandable because of that (libpng1 is very inconsistent, I'm trying to fix that!)

Bear in mind that these checks are often simply blurted out to the application user so, for example, "Ignoring wasteful fcTL BLEND_OP_OVER in opaque images" is not going to be useful.  Also that warning is erroneous; does libpng issue a warning if the filter on the first line of a pass is UP?  Only NONE, SUB and AVG are useful on the first line, but it is not appropriate for the code which **read** a PNG to output diagnostics about the unknown code which **wrote** the PNG!

The optimisation is also incorrect; see `png_set_next_frame_fcTL` and note that it is checking `info_ptr` for tRNS.  That doesn't work with two png_infos and possibly not with png_set_expand_tRNS (I think it might but that would be an infelicity, or, arguably, a bug.)

pngwutil.c
```diff
@@ -1969,6 +1993,83 @@ png_write_tIME(png_structrp png_ptr, png_const_timep mod_time)
...
+               png_uint_32 width, png_uint_32 height,
+               png_uint_32 x_offset, png_uint_32 y_offset,
+               png_uint_16 delay_num, png_uint_16 delay_den,
+               png_byte dispose_op, png_byte blend_op)
+{
+   png_byte buf[26];
+
+   png_debug(1, "in png_write_fcTL");
+
+   if (png_ptr->num_frames_written == 0 && (x_offset != 0 || y_offset != 0))
```
jbowler, line 2028: This looks wrong, it's correct if the first frame is the IDAT but that's a different condition.  Checking "num_frames_written" seems weird.  At best 'num_frames_written' is a non-self-documenting name.

pngwutil.c
```diff
@@ -2821,4 +2922,39 @@ png_write_filtered_row(png_structrp png_ptr, png_bytep filtered_row,
...
+   png_ptr->row_number = 0;
+   png_ptr->pass = 0;
+   png_ptr->mode &= ~PNG_HAVE_IDAT;
+}
+
+void /* PRIVATE */
+png_write_reinit(png_structp png_ptr, png_infop info_ptr,
+                 png_uint_32 width, png_uint_32 height)
+{
+   if (png_ptr->num_frames_written == 0 &&
```
jbowler, line 2939: This is clearly code duplication; png_write_reinit is called from png_write_frame_head (and only from there) along with png_write_fcTL but the latter call is conditioned on a complex expression which using boolean arithmetic, reduces to:

`if (png_ptr->num_frames_written != 0 || !(png_ptr->apng_flags & PNG_FIRST_FRAME_HIDDEN)`

That is way more clear to me but I still can't easily work out whether it writes the IDAT if !PNG_FIRST_FRAME_HIDDEN (that's a horrible new concept as well; PNG_FIRST_FRAME_IS_IDAT is better).  (NOTE: I'm talking about the test in pngwrite.c and I think png_write_reinit should be in there too and be inlined.)

pngrutil.c
```diff
@@ -2727,6 +2732,184 @@ png_handle_iTXt(png_structrp png_ptr, png_inforp info_ptr, png_uint_32 length)
...
+      return;
+   }
+
+   png_crc_read(png_ptr, data, 8);
+   png_crc_finish(png_ptr, 0);
+
+   num_frames = png_get_uint_31(png_ptr, data);
+   num_plays = png_get_uint_31(png_ptr, data + 4);
+
+   /* The set function will do error checking on num_frames. */
```
jbowler, line 2774: Code duplication: all the png_set_chnk functions set the relevant PNG_INFO flags.  Also they are all void.  I've made both changes in my version.

pngrutil.c
```diff
@@ -2727,6 +2732,184 @@ png_handle_iTXt(png_structrp png_ptr, png_inforp info_ptr, png_uint_32 length)
...
+void /* PRIVATE */
+png_handle_acTL(png_structp png_ptr, png_infop info_ptr, png_uint_32 length)
+{
+   png_byte data[8];
+   png_uint_32 num_frames;
+   png_uint_32 num_plays;
+
+   png_debug(1, "in png_handle_acTL");
+
+   if (!(png_ptr->mode & PNG_HAVE_IHDR))
```
jbowler, line 2745: Massive code duplication: this is all done in png_handle_chunk.  png_handle_acTL needs to be called that way.  I'll update it.  I'm not sure about png_set_fcTL but I think it can too.  The updates to pngrutil.c have extensive confusing placement.  The two functions png_set_fcTL and png_set_acTL are separated incorrectly from the #defines of the same things when they are compiled out.

pngrutil.c
```diff
@@ -4689,4 +4906,70 @@ defined(PNG_USER_TRANSFORM_PTR_SUPPORTED)
...
+#ifdef PNG_READ_APNG_SUPPORTED
+/* This function should be called after the main IDAT sequence has been read
+ * and before a new fdAT is about to be read. It resets some parts of png_ptr
+ * to make them usable by the read functions again.
+ */
+void /* PRIVATE */
+png_read_reset(png_structp png_ptr)
+{
+   png_ptr->mode &= ~PNG_HAVE_IDAT;
+   png_ptr->mode &= ~PNG_AFTER_IDAT;
```
jbowler, line 4919: This won't work.  It's going to kill the "unknown" handling which relies on PNG_AFTER_IDAT to not rearrange unknown chunks before IDAT.  It will also kill the tests in chunk order for any chunk which might be defined in the future to only be permitted after IDAT.

Both HAVE_IDAT and AFTER_IDAT need to be preserved.  The same for killing PNG_HAVE_IDAT in pngwutil.c

It may work with PNG_HAVE_fcTL but I don't know.  I think png_struct::apng_flags needs to be used.  The row decompress code can be changed to use the ZSTREAM flags in png_struct::flags.

jbowler (changes requested):
I'm working on changes to address these and other comments.

jbowler:
There's a major bug in the new code to clear flags in png_struct::mode; that cannot happen and just does not work.  It changes the whole way the parser (both progressive and sequential) works and only seems to have been done to implement png_handle_fcTL and reuse the row reading code.

Clearly the row reading code can be changed to use different flags and I think it should be changed in that way: the design of "mode" is to store information about the position within the PNG stream WRT critical chunks, these are not critical chunks so removing the "AFTER_IDAT" flag is going to, as a minimum, mess up tests for chunks which must occur **before** IDAT.  What happens if I add a cICP after IDAT in an APNG?  That's not valid because of the positioning rules, but those check for !(png_struct::mode & PNG_HAVE_IDAT)  In fact I think IEND will always fail.

pngrutil.c
```diff
@@ -2727,6 +2732,184 @@ png_handle_iTXt(png_structrp png_ptr, png_inforp info_ptr, png_uint_32 length)
...
+    * - This is an out-of-place fdAT.
+    * In either case, it is safe to ignore the chunk with a warning.
+    */
+   png_warning(png_ptr, "Ignoring fdAT chunk");
+   png_crc_finish(png_ptr, length - 4);
+   PNG_UNUSED(info_ptr)
+}
+
+void /* PRIVATE */
+png_ensure_sequence_number(png_structp png_ptr, png_uint_32 length)
```
jbowler, line 2889: This does and cannot work.  If 'length' is short it reads nothing from the stream but if length is at least 4 it reads 4 bytes.  The caller cannot know how many bytes remain in the stream so cannot clean up to skip the chunk.  The whole code set assumes that fcTL and fdAT are critical chunks and just calls png_error but the chunks are ancillary and the PNG v3 spec says that the static image must be displayed and the standard rules say that errors in ancillary chunks should be handled if possible.

pngwrite.c
```diff
@@ -1484,6 +1494,45 @@ png_write_png(png_structrp png_ptr, png_inforp info_ptr,
...
+   if (!(png_ptr->num_frames_written == 0 &&
+         (png_ptr->apng_flags & PNG_FIRST_FRAME_HIDDEN)))
+      png_write_fcTL(png_ptr, width, height, x_offset, y_offset,
+                     delay_num, delay_den, dispose_op, blend_op);
+
+   PNG_UNUSED(row_pointers)
+}
+
+void PNGAPI
+png_write_frame_tail(png_structp png_ptr, png_infop info_ptr)
```
jbowler, line 1527: This shouldn't be an API call it should be happening inside the row write code, at the end when the data has been written correctly.  This saves an API call to increment num_frames_written, makes the internal good more clear and avoids an app bug where the call is omitted.

I haven't tested that but I think it might allow multiple calls to png_write_image and maybe png_write_frame_head too; the only places it is used inside libpng are the incorrect test of 0 to determine if this is a before-IDAT fcTL (the code should check PNG_HAVE_IDAT but it can't be cause the code cancels that **before** each frame) and the single check at the end of pngwrite.c in png_write_end.

I'm not sure but maybe num_frames_written is completely unnecessary; doesn't end up just being a check for "before IDAT"?

jbowler (commented):
This is all leading me to the conclusion that using a control structure in the manner of the simplified API would make this much easier to understand (well, some documentation would help).  This doesn't just apply to "write", the control structure on "read" is also much better because doing it in both avoids the need from the unfortunate use of png_info (which png_info) in its place,  I might have a look at that as it seems productive but I'll probably see if the PNG_HAVE_IDAT/AFTER_IDAT misuse can be fixed easily (I think it can).

ctruta:
@jbowler wrote:

> @MaartenBent I think your patch is good but it's clear that at least one quarter of the common cases used by application writers haven't been tested by either you or @ctruta (or other APNG developers).

John's assumption is correct. (Speaking for myself, about myself, to be clear.) Under the `<air-quotes>`_good excuse_`</air-quotes>` that we're in the development branch, right at the start of libpng-2, and that we have a lot of things to test and a long way to go, I'm landing this in, including the suggested changes.

On the other hand, as far as I know, this APNG patch has been long in use in various APNG-supporting software, most notably in Mozilla. For that reason, I don't consider it to be bleeding-edge-untested, either.

FYI, the CI verifications for the **master** branch (which stands for libpng-1.6.x) and for the **develop** branch (which stands for libpng-2.x) are here:
https://ci.appveyor.com/project/ctruta/libpng/

Thank you everyone for your participation :-)

pngrutil.c
ctruta: Thanks.

pngrutil.c
ctruta: I applied Maarten's change.

pngrutil.c
ctruta: I would normally ask the author of the patch (in this case @unitarou) to fix before landing, but in this case I think we agree to make an exception. We'll apply now, and we'll fix it later.

pngrutil.c
ctruta: Here be FIXME note.

pngrutil.c
ctruta: Here, too, be FIXME note.

pngwutil.c
ctruta: I, too, noticed some variables that need renaming.

pngwutil.c
ctruta: Here be yet another FIXME note.

jbowler:
@svgeesus @ctruta  The GIMP approach (in the development branch, 3.1.2) is the same as I have suggested before but (unlike AREATK) it is stateful so it uses the "unknown chunk callback" to store the frames as the PNG is read.  It's clean and a good example of how to do it.  It's a simple modification of the main GIMP PNG read code in 3.0.4, Max Stepin is quoted as the reference for the APNG implementation.

The patch appears to be from the sourceforge web site.  There is also a version for GIMP with is covered by GPL-2/GPL-3

The patch itself **does not contain a copyright**; SourceForge lists the **license** as "zlib/libpng" but that's not in the patch anywhere either.

That aside I looked at the original patch (from sourceforge) applies to 1.6.  The way it works is a bit of a problem because it seems certain it will break GIMP.  At one point it was known to cause problems for Chromium as well.  This comes back to what I've said before about adding new "known" chunks however I'm pretty sure that this particular breakage can be avoided.

peterkaczorowski:
@jbowler 

The GIMP team started working on APNG support before @ctruta  made his code public. GIMP 3.1.2 is part of the development branch, so I believe that APNG support will soon be unified and aligned with the PNG team's prior consultations.

jbowler:
@peterkaczorowski this is the code:

https://sourceforge.net/p/libpng-apng/code/ci/master/tree/

It's been available since 1.2.29, possibly earlier elsewhere.  It's the code in the NetScape/Mozilla browser, I think.  It's available on many systems.  The problem is it causes problems to some existing code, not least Chromium but it's pretty much always been there.

jbowler:
Better, here is the code in git:

https://github.com/mozilla-firefox/firefox/tree/main/media/libpng

The authors still aren't there for sure because the log shows changes that are apparently  **part** of the changes in a new release checked in separately via a patch, however this file would seem to be authorative:

https://github.com/mozilla-firefox/firefox/blob/main/media/libpng/MOZCHANGES

```
This modified version of libpng code adds animated PNG support and is
released under the same license as the upstream library. The modifications
are Copyright (c) 2006-2007 Andrew Smith, Copyright (c) 2008-2017 Max Stepin,
and are delimited by "#ifdef PNG_APNG_SUPPORTED / #endif" directives
surrounding them in the modified libpng source files.

```

ctruta:
@jbowler wrote:

> https://sourceforge.net/p/libpng-apng/code/ci/master/tree/

For the record: this is what I applied.

https://sourceforge.net/projects/libpng-apng/files/libpng16/1.6.49/

I started with a temporary application to my local **libpng16** branch. Then I did a cherry-pick with an obligatory rework onto the **develop** branch. And then, finally, the subsequent cleanup commit, and the fixes that everybody else contributed. I credited @unitarou (Daisuke Nishikawa) in the AUTHORS file.

> Better, here is the code in git:
> 
> https://github.com/mozilla-firefox/firefox/tree/main/media/libpng
> 
> The authors still aren't there for sure because the log shows changes that are apparently **part** of the changes in a new release checked in separately via a patch, however this file would seem to be authorative:
> 
> https://github.com/mozilla-firefox/firefox/blob/main/media/libpng/MOZCHANGES

Thank you for pointing it out. The AUTHORS file must be updated for sure.

unshumikan:
@ctruta wrote:
>I started with a temporary application to my local libpng16 branch. Then I did a cherry-pick with an obligatory rework >onto the develop branch. And then, finally, the subsequent cleanup commit, and the fixes that everybody else >contributed. I credited @unitarou (Daisuke Nishikawa) in the AUTHORS file.

Hello!
I'm APNG-patch for libpng author on sourceforge. I just created account @unshumikan on github.
@unitarou is another person with the same name. It is common name in Japan.

I'm happy it's merged.

ctruta:
Arigatō, @unshumikan :+1:
Will ping you for the future developments.

ctruta:
@unitarou: heads-up: the push-mode APNG parser has a chunk-smuggling bug in three inter-frame discard paths (trailing IDAT, trailing fdAT, and unknown chunks). I opened a security advisory and I added you there as a collaborator.

The fix is in commit faf06924688b62d7c1654b5ceddedbde66ffadb4.

PR #707, opened 2025-06-21
pngtest nowrite

jbowler:
- **Introduce APNG support -- patch applied, animations unlocked!**
- **refactor: Tidy up the APNG-supporting code**
- **tests: fix pngtest palette test when write unsuppo**


jbowler:
Based on your #706 but it's freestanding.

jbowler:
Duh.  Doing it that way terminates the pipe feeding into check_stdout so the **preceding** check_stdout in the pipeline will be killed.  check_stdout needs to read to the end.

jbowler:
Now it doesn't abort the output if there is no write support.  Also tested with the new APNG tests in #708

ctruta:
Integrated in trunk (but GitHub didn't notice).

PR #710, opened 2025-06-22
Fix unknown handling of APNG and Apple chunks

jbowler:
- **Introduce APNG support -- patch applied, animations unlocked!**
- **refactor: Tidy up the APNG-supporting code**
- **Handle APNG and Apple chunks correctly**


jbowler:
Remove the spurious reset of the 'keep' unknown setting and document the behavior of png_set_keep_unknown_chunks correctly.  Also remove the unnecessary change to handling of safe-to-copy known chunks.

jbowler:
Note that I have a lot more changes.  I stopped trying to break them down because it was clear that I couldn't keep working on the APNG stuff without changes I'd already mentioned in the code reviews and doing just code review when there are so many changes was simply not going to achieve anything.  Specifically it takes me about an hour to explain something that I can fix in a minute.

Anyway, current state (still definitely a WIP) is in the branch https://github.com/jbowler/libpng/tree/libpng2-apng

So far it's going well and things are getting a lot simpler.  The APNG patch adds about 40K on AMD64 to the 250K libpng DLL but I think these changes will considerably reduce the size from 290K, but I've only managed to save 3K so far :-(

ctruta (approved):
Reviewed, approved, and integrated in trunk (although GitHub didn't notice).

PR #712, opened 2025-06-24
FIX: unknown IDAT causes out-of-place IEND

jbowler:
PNG_AFTER_IDAT was not set by the IDAT read code if unknown handling for IDAT was turned on.  This was hidden in the current tests by checks within the text handling chunks (the test file, pngtest.png, has a zTXt after IDAT).

This change modifies both the sequential and progressive reader to reliably set PNG_AFTER_IDAT when the first non-IDAT chunk is seen and before that chunk is processed.

The change is minimalist; PNG_HAVE_CHUNK_AFTER_IDAT can probably be removed and replaced with PNG_AFTER_IDAT with this change.  This change is apparently not necessary so is deferred until libpng2.

### DIFF
--- a/pngpread.c
+++ b/pngpread.c
@@ -229,6 +229,15 @@ png_push_read_chunk(png_structrp png_ptr, png_inforp info_ptr)
          png_benign_error(png_ptr, "Too many IDATs found");
    }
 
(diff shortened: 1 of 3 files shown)


ctruta (approved):
Approved, with the same amendment as in #715.

ctruta:
Integrated with copyediting modifications.

PR #713, opened 2025-06-27
Libpng16 build fixes for riscv

filipwasil:
## Autotools

```
CFLAGS="-march=rv64gv1p0" ./configure --host=riscv64-linux-gnu --disable-shared --enable-hardware-optimizations
```
**builds fine**, rvv included

```
./configure --host=riscv64-linux-gnu --disable-shared --enable-hardware-optimizations
```
**builds fine**, rvv not included (not introduced any march flag)

```
CFLAGS="-march=rv64gv1p0" ./configure --host=riscv64-linux-gnu --disable-shared
```
**builds fine**, rvv included (--enable-hardware-optimizations on by default)

```
CFLAGS="-march=rv64gv1p0" ./configure --host=riscv64-linux-gnu --disable-shared --enable-riscv-rvv=api
```
configure: error: --enable-riscv-rvv=api: invalid value

```
CFLAGS="-march=rv64gv1p0" ./configure --host=riscv64-linux-gnu --disable-shared --enable-riscv-rvv=check
```
configure: error: --enable-riscv-rvv=api: invalid value

```
CFLAGS="-march=rv64gv1p0" ./configure --host=riscv64-linux-gnu --disable-shared --enable-riscv-rvv=on
```
**builds fine**, rvv included

```
CFLAGS="-march=rv64gv1p0" ./configure --host=riscv64-linux-gnu --disable-shared --enable-riscv-rvv=off
```
**builds fine**, rvv not included

```
CFLAGS="-march=rv64gv1p0 -DPNG_RISCV_RVV_OPT=1" ./configure --host=riscv64-linux-gnu --disable-shared --enable-riscv-rvv=off
```
wrong usage, PNG_RISCV_RVV_OPT redefined. Autotools only uses --enable-riscv-rvv  or  --enable-hardware-optimizations

## CMake
```
CC=gcc-14 CXX=g++14 CFLAGS="-march=rv64gcv1p0" cmake .. -B .
```
**builds fine**, rvv not included. 

```
CC=gcc-14 CXX=g++14 CFLAGS="-march=rv64gcv1p0" cmake .. -B . -D PNG_RISCV_RVV=off
```
**builds fine**, rvv not included

```
CC=gcc-14 CXX=g++14 CFLAGS="-march=rv64gcv1p0" cmake .. -B . -D PNG_RISCV_RVV=on
```
**builds fine**, rvv included

```
CC=gcc-14 CXX=g++14 CFLAGS="-march=rv64gcv1p0" cmake .. -B . -D PNG_RISCV_RVV=yes
```
**CMake Error** at CMakeLists.txt:332 (message):
  PNG_RISCV_RVV must be one of [on;off]

```
CC=gcc-14 CXX=g++14 CFLAGS="-march=rv64gcv1p0" cmake .. -B . -D PNG_RISCV_RVV=no
```
**CMake Error** at CMakeLists.txt:332 (message):
  PNG_RISCV_RVV must be one of [on;off]


jbowler (approved):
The changes look fine however I get build failures on some., but not all, of the *non*rvv arches; riscv64i and rsicb64g.  For the latter I get an "unused parameter" warning in png_have_rvv at line 28 of contrib/riscv-rvv/linux.c but that's followed by an FP exception in the compiler (Gentoo GCC 15.1.0 p55 using gcc, not g++)

The exception is at line 347 of filter_rvv_intrinsics,c

This isn't blocking: the current HEAD has the same problem sometimes.

@ctruta: I suggest pulling this because, a compiler bug shouldn't stop what seems to be to be a reasonable fix to an existing, and serious, bug in 1.6.49

@filipwasil I still feel that the use of asm, particularly asm volatile, is not wise but at this point it is more important to fix the build break.

filipwasil:
> The changes look fine however I get build failures on some., but not all, of the _non_rvv arches; riscv64i and rsicb64g. For the latter I get an "unused parameter" warning in png_have_rvv at line 28 of contrib/riscv-rvv/linux.c but that's followed by an FP exception in the compiler (Gentoo GCC 15.1.0 p55 using gcc, not g++)
> 
> The exception is at line 347 of filter_rvv_intrinsics,c
> 
> This isn't blocking: the current HEAD has the same problem sometimes.
> 
> @ctruta: I suggest pulling this because, a compiler bug shouldn't stop what seems to be to be a reasonable fix to an existing, and serious, bug in 1.6.49
> 
> @filipwasil I still feel that the use of asm, particularly asm volatile, is not wise but at this point it is more important to fix the build break.

1. unused parameter removed
2. The line 347 of filter_rvv_intrinsics.c is a single "}", and there is no asm volatile left in the rvv code so it suggests old code present (running old binary maybe ?) not sure. Can You please double check ? This PR is 4 commits big.
3. the -march=rv64g and -march=rv64i are both possible to test on my setup and both result with the same effect - they fail compilation when one tries to do PNG_RISCV_RVV=on (CMake) or --enable-hardware-optimizations or --enable-riscv-rvv (Autotools), and they both succeed if one does not try to do so (Excluding simd support).

ctruta:
Integrated in the master branch (although GitHub doesn't seem to think so).

PR #721, opened 2025-07-07
RISCV support only RVV 1.0

filipwasil:
- No linux specific code, removed linux specific files
- If we are able to compile Vector extension having <riscv_vector> header and supporting all the types defined in AC_COMPILE_IFELSE or check_c_source_compiles, then we can support RVV. Which version of RVV we support,and which functions we will register we decide in RVV specific code. If the version if not supported, we just do not register anything. If we support part of it, we register part of it etc.
- Adding support for (nonexisting yet) RVV 2.0, RVV 1.1 or other future versions will be possible by extending the condition in **riscv_init.c** or adding a separate block of code under different compile time switch (additionally to __riscv_v being 1000000), or by adding a separate block of code under PNG_RISCV_RVV_IMPLEMENTATION having value different than 0 or 1.
- no runtime check done


jbowler (changes requested):
It's still building T-Head.  It might not execute it but it's still building it.  You need to put the __riscv_v check in pngpriv.h and disable PNG_RISCV_RVV_IMPLEMENTATION   GCC T-Head might manage it ATM but who knows what happens in the future.

The configure.ac and CMakeLists.txt are not necessary; they're nanny-state.  If someone manages to turn it on and that header is not available when __riscv_v is defined the code will not compile and whoever managed to do it will fix what they did wrong.

filipwasil:
```
__riscv_v >= 1000000 && __riscv_v < 1900000
```

according to [spec](https://tools.cloudbear.co/docs/riscv-unprivileged-20240411.pdf)
Changes in major version numbers imply a loss of backwards compatibility, whereas changes in
only the minor version number must be backwards-compatible.

jbowler (approved):
@ctruta: this now passes my build tests including T-Head.  Since the unit tests are confirmed to pass on one machine this should be integrated into 1.6.51.  For the unit tests see:

https://github.com/pnggroup/libpng/issues/711#issuecomment-3043868388

ctruta (approved):

ctruta:
Integrated in trunk, in spite of GitHub claiming the contrary. Many thanks, @filipwasil.

ctruta:
Immediately after integrating this PR, I realized I had forgotten to complete the `Reviewed-by:` field. (Oopsie!)
So I quickly force-pushed the updated commit while nobody was looking :see_no_evil:

jbowler:
@ctruta can you cherry-pick/merge it over to libpng 1.8?  I believe it's now in a state to be upgraded to the pngtarget.h mechanism.

ctruta:
> @ctruta can you cherry-pick/merge it over to libpng 1.8? I believe it's now in a state to be upgraded to the pngtarget.h mechanism.

You should be able to do the cherry-picking, too, @jbowler, and then I'll integrate it. But sure, no problem -- done.

BTW, I'm not sure if I communicated this well enough, but just in case, here's a clarification:

The development for libpng is in the `develop` branch, slated to become libpng v2. After that becomes stable and ready to release, then the `develop` branch will be, I don't know... libpng v2.1... or libpng v3... well, it surely will be still the `develop` branch.

On the other hand, libpng v1.6 will enter maintenance mode, and libpng v1.8, too, pretty much immediately after release. The definition of libpng v1.8 is this: "libpng v1.6 _plus_ APNG _minus_ cruft". And its purpose is this: I want the future patches to libpng v1.6 to be easily merged or cherry-picked into v1.8 with little fuss, which is why I'd rather not modernize v1.8 too much.

To me, the pngtarget.h mechanism you're talking about sounds to be a better candidate for the `develop` branch -- except, of course, you may pick and choose any branch you wish, @jbowler.

> Note that my APNG work is based on 1.8 and uses C99 extensively.

The same applies to using C99. Feel free to use C99 to your heart's content in the `develop` branch, and even if you do nothing but C99-ify old code in a commit, I'll still gladly take that commit. I mean, libpng may be 3 decades old, but it shouldn't need to look that old.

libpng v1.8 is not much more than a stop-gap solution to bring in APNG, to take out technical debt, and to do it sooner instead of later.

PR #722, opened 2025-07-09
Fix out-of-bounds access in png_read_filter_row_sub3_sse2

sanjay20m:
Fix out-of-bounds access in png_read_filter_row_sub3_sse2 by handling trailing bytes correctly

This patch resolves a subtle read/write out-of-bounds vulnerability in the png_read_filter_row_sub3_sse2 function, which occurs when the image width is not a multiple of 3.

Previously, the function applied SSE2 vector operations in a loop (load4() / store3()), then exited without properly handling 1–2 trailing bytes. The fallback incorrectly tried to load/store using SIMD instructions again, which risks memory access outside the allocated row buffer.

✅ Fix:
	•	Introduced scalar fallback logic to process 1 or 2 remaining bytes safely, using regular byte-wise operations.
	•	Prevents potential memory corruption or access violation due to unaligned or invalid reads.

### DIFF
--- a/intel/filter_sse2_intrinsics.c
+++ b/intel/filter_sse2_intrinsics.c
@@ -119,40 +119,21 @@ void png_read_filter_row_avg3_sse2(png_row_infop row_info, png_bytep row,
 
    size_t rb;
 
-   const __m128i zero = _mm_setzero_si128();
-
-   __m128i    b;
-   __m128i a, d = zero;
+   const __m128i one = _mm_set1_epi8(1);
+   __m128i a, b, d = _mm_setzero_si128();
 
    png_debug(1, "in png_read_filter_row_avg3_sse2");
+
    rb = row_info->rowbytes;
-   while (rb >= 4) {
+   while (rb >= 3) {
       __m128i avg;
-             b = load4(prev);
-      a = d; d = load4(row );
 
-      /* PNG requires a truncating average, so we can't just use _mm_avg_epu8 */
-      avg = _mm_avg_epu8(a,b);
-      /* ...but we can fix it up by subtracting off 1 if it rounded up. */
-      avg = _mm_sub_epi8(avg, _mm_and_si128(_mm_xor_si128(a,b),
-                                            _mm_set1_epi8(1)));
-      d = _mm_add_epi8(d, avg);
-      store3(row, d);
+      b = load3(prev);
+      a = d;
+      d = load3(row);
 
-      prev += 3;
-      row  += 3;
-      rb   -= 3;
-   }
-   if (rb > 0) {
-      __m128i avg;
-             b = load3(prev);
-      a = d; d = load3(row );
-
-      /* PNG requires a truncating average, so we can't just use _mm_avg_epu8 */
-      avg = _mm_avg_epu8(a,b);
-      /* ...but we can fix it up by subtracting off 1 if it rounded up. */
-      avg = _mm_sub_epi8(avg, _mm_and_si128(_mm_xor_si128(a,b),
-                                            _mm_set1_epi8(1)));
+      avg = _mm_avg_epu8(a, b);
+      avg = _mm_sub_epi8(avg, _mm_and_si128(_mm_xor_si128(a, b), one));
 
       d = _mm_add_epi8(d, avg);
       store3(row, d);
@@ -161,6 +142,8 @@ void png_read_filter_row_avg3_sse2(png_row_infop row_info, png_bytep row,
       row  += 3;
       rb   -= 3;
    }
+
+   PNG_UNUSED(prev);
 }
 
 void png_read_filter_row_avg4_sse2(png_row_infop row_info, png_bytep row,
--- a/intel/filter_sse2_intrinsics.c
+++ b/intel/filter_sse2_intrinsics.c
@@ -51,10 +51,6 @@ static void store3(void* p, __m128i v) {
 void png_read_filter_row_sub3_sse2(png_row_infop row_info, png_bytep row,
    png_const_bytep prev)
 {
-   /* The Sub filter predicts each pixel as the previous pixel, a.
-    * There is no pixel to the left of the first pixel.  It's encoded directly.
-    * That works with our main loop if we just say that left pixel was zero.
-    */
    size_t rb;
 
    __m128i a, d = _mm_setzero_si128();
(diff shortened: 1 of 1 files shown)


jbowler:
Do you mean that it is reading beyond the end of the inflated bytes in the row?  There are IRC, more than 16 bytes allocated beyond the end and I assume, from the instruction names, that they are loading 16 bytes at a time.

Yes, it's a bug if a hardware implementation reads beyond the end of the inflate data because it may, in the future, either read uninitialised or, perhaps, unallocated memory (though ISO-C aligned_alloc always allocates a multiple of the alignment size so that probably won't happen.)

What's the test case?  I.e. please attach an image so that we can trap this in the future.

I agree with the use of a scaler fallback, in fact I favour scalar start as well to avoid the alignment requirement completely.

If you have a test case which reads beyond the end of the **actual** allocation then that, certainly, will be very valuable.

@ctruta, the SSE code is hard disabled in 1.6 and very difficult to enable.  That means it's not much tested.  If there is a genuine case of read beyond the end of Glenn's +48 byte allocation (which typically ends up +33 bytes at the end) then the lack of testing is extreme and I will argue for complete disablement of that code in 1.6

intel/filter_sse2_intrinsics.c
```diff
@@ -51,36 +51,38 @@ static void store3(void* p, __m128i v) {
...
-    * That works with our main loop if we just say that left pixel was zero.
-    */
    size_t rb;
 
    __m128i a, d = _mm_setzero_si128();
 
    png_debug(1, "in png_read_filter_row_sub3_sse2");
 
    rb = row_info->rowbytes;
-   while (rb >= 4) {
```
jbowler, line 65: I don't understand why, given that rb >= 4, load4(row) isn't safe.  Do you have a machine where "sizeof (int)" is greater than 4?  Clearly memcpy in load4 is only copying at most 4 bytes if sizeof (tmp) <= 4.

rb is certainly a count of bytes and row points to exactly rb valid bytes.

Likewise the store3 stores exactly 3 bytes (this time explicitly 3) at 'row' and that's entirely safe.

jbowler:
@ctruta: more explanation required.  A code review indicates to me that that existing code does not read beyond the actual rowbytes of the row.

fintelia:
This PR seems to be addressing the case where a 3-byte-per-pixel image has a width that isn't a multiple of 3-bytes, but that doesn't seem possible?

sanjay20m:
Yes, you're correct that for color_type == PNG_COLOR_TYPE_RGB, rowbytes is typically a multiple of 3. However, this fix is intended to safely handle edge cases where padding or misalignment may occur due to:

Custom or future use cases,

Truncated input,

Malformed PNGs that pass through decompression but leave row_info->rowbytes < width * 3 (e.g., due to filtering or bit-depth issues).


Even though libpng allocates extra padding bytes (+48), SIMD instructions like movdqa or unaligned load4() might still attempt to read from uninitialized memory regions. While technically allowed due to over-allocation, it's not memory-safe on all architectures.

intel/filter_sse2_intrinsics.c
sanjay20m: You're correct that `load4(row)` uses `memcpy()` to copy at most 4 bytes, and if `rb >= 4`, it should be safe. However, the concern arises in the edge case when `rb` is not divisible by 3 (i.e., 3-byte-per-pixel images where the row width isn’t a multiple of 3). In such cases, it's possible that on the last iteration, we might be processing only 3 valid bytes but attempting to load 4 using `load4()`.

While `memcpy()` doesn't necessarily crash here (especially due to libpng’s over-allocation), this technically leads to **undefined behavior** in C, as it reads 1 byte beyond the end of the valid buffer. The behavior may be benign on many systems, but it is still non-compliant with memory safety best practices — and could cause issues in hardened environments or with newer toolchains.

The fallback I added ensures that any trailing 1–2 bytes are handled safely with scalar code. It’s a **defensive fix** that aims to make this SIMD path robust across platforms, without relying on assumptions about allocator behavior or padding.

Let me know if you’d prefer a scalar *prefix* approach as well — I’d be happy to adjust.

jbowler:
If rowbytes is not a multiple of 3 after the colour type has been set to PNG_COLOR_TYPE_RGB then that's a bug somewhere else, a serious one, but the suggested change does not help; if "rowbytes" has been miscalculated how is it possible to be sure it has not been made **too large**?

@fintelia, @sanjay20m do you, or anyone else, know why that loop is using `load4` and not `load3`?   The only difference between the two is in the `memcpy` and copying a spurious byte is, surely, going to slow everything down because the extra byte will be in a new cache line every so often.

**Mike Klein** - you wrote that line; it looks like a copy'n'paste error to me, as does the "rb > 0" trailing if.  That was written by **Matt Sarett**.

@ctruta: all unit checks pass when I change load4 -> load3 on line 66 and remove the 'if (rb > 0)...` trailer.  (With `-DPNG_ENABLE_SSE` of course :-)

sanjay20m:
The recent update I made refactors the png_read_filter_row_avg3_sse2 function to correctly process 3-bytes-per-pixel rows using safe, aligned SSE2 instructions. It replaces the previous load4() usage with load3() to avoid overreads and improve memory safety and cache efficiency.
	-	Ensures that only 3 valid bytes are processed per pixel.
	-	Removes unnecessary fallback code.
	-	Matches the PNG Avg filter specification precisely using avg = (a + b) >> 1, adjusted for truncation per spec.
	-	Clean, minimal SIMD logic with correct handling of edge cases.

Impact
	-	Prevents potential overreads when processing PNG rows not aligned to 4 bytes.
	-	Improves clarity and safety of SSE2 filter code.
	-	Makes the SSE2 implementation more maintainable and trustworthy.

Thanks again for the earlier review and support!

jbowler:
The same applies to paeth3 and avg3:

https://github.com/jbowler/libpng/tree/intel-sse2-improvements

However libpng 1.6 is stable so if this change is to go in it should go in to libpng 1.8; there is no bug in 1.6

fintelia:
I'm increasingly convinced that sanjay20m's comments are written by an LLM. 

In fact, the use of load4 rather than load3 is both intentional and very clever. On x86, there's no instruction to directly load 3 bytes into a SIMD register so the generated assembly to do so would be rather bad. There _is_ however the ability to load 4 unaligned bytes into a register. Thus, by harmlessly loading an extra element (after first checking `rb >= 4` so it doesn't go past the end of the input) the generated assembly can be much more efficient. You can compare the different assembly between load3 and load4 here: https://godbolt.org/z/eh71d6h9r

[Here is the same code in spng](https://github.com/randy408/libspng/blob/adc94393dbeddf9e027d1b2dfff7c1bab975224e/spng/spng.c#L6250)

ctruta:
> However libpng 1.6 is stable so if this change is to go in it should go in to libpng 1.8; there is no bug in 1.6

I agree that there's no bug in libpng v1.6 (or in v1.8, etc.)

@sanjay20m, my request in PR #731 applies here as well: please bring forth a proof of concept showing the actual vulnerability, ideally accompanied by a concrete demonstration of data corruption or an actual crash.

PR #723, opened 2025-07-10
Fix integer overflow in array allocation in pngmem.c

sanjay20m:
Add safe multiplication check to prevent integer overflow in array allocations within png_malloc_array_checked(), used by png_malloc_array() and png_realloc_array().

Previously, multiplication of nelements and element_size could wrap around, causing under-allocated buffers and potential heap overflows.

This patch introduces png_safe_multiply() to validate the multiplication before allocation, preventing memory corruption vulnerabilities.

Fix aligns with secure coding practices and qualifies under the "Secure-by-design memory safety improvement" category.

### DIFF
--- a/pngmem.c
+++ b/pngmem.c
@@ -98,7 +98,8 @@ png_malloc_base,(png_const_structrp png_ptr, png_alloc_size_t size),
 }
 /* Helper to check for safe multiplication without overflow */
 static int
-png_safe_multiply(png_alloc_size_t a, png_alloc_size_t b, png_alloc_size_t *result)
+if (png_safe_multiply((png_alloc_size_t)nelements,
+                      (png_alloc_size_t)element_size, &total))
 {
     if (a == 0 || b == 0) {
         *result = 0;
--- a/pngmem.c
+++ b/pngmem.c
@@ -126,7 +126,8 @@ png_malloc_array_checked(png_const_structrp png_ptr, int nelements,
 if (nelements <= 0 || element_size == 0)
    return NULL;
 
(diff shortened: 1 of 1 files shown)


sanjay20m:
@ctruta  Please review :) And let me know the changes !!

jbowler:
No.  It's a static, check the callers.

jbowler:
@ctruta; I think we both checked this two days ago.  I certainly did.  Also see the comment about why it is out-of-line but it doesn't matter because it's a static; the compiler will inline it if appropriate.

sanjay20m:
Since the patch now passes all CI checks and incorporates minimal, safe overflow protection, please let me know if there are any final concerns or changes you’d like me to address.

Thanks again for your time and input!

sanjay20m:
Just wanted to follow up and check if there's anything else needed from my side.

The patch now passes all CI checks, avoids integer overflow in array allocations, and maintains minimal and portable logic.

Let me know if any final adjustments are required — happy to make them. Otherwise, I’ll leave it here for your consideration for inclusion. Thank you again for the review!

jbowler:
@sanjay20m I'll try to simplify what I said before:

All of the arguments to `png_malloc_array_checked` have been checked by the callers; hence `_checked` in the name.  Therefore any addition of checks is entirely wasted, indeed unreachable, code and should not be done; in essence you are fixing a bug which is not there.

ctruta:
@jbowler wrote:

> All of the arguments to `png_malloc_array_checked` have been checked by the callers; hence `_checked` in the name. Therefore any addition of checks is entirely wasted, indeed unreachable, code and should not be done; in essence you are fixing a bug which is not there.

I concur, and I'm closing this PR as rejected.

PR #732, opened 2025-09-03
Saying goodbye to the nineties, one cleanup job at a time

ctruta:
Until further notice, we're still supporting MS-DOS, where doing so makes sense. Watcom C with DOS/4GW and gcc with DJGPP do count. But on the other hand, we discarded all remaining uses of `_farmalloc` in libpng-1.6.0, which meant no more blocks of 64KB for us on 16-bit machines. Did libpng still work, even without it? Maybe yes, a little, in a rather limited way that would have restricted the compression opportunities in zlib; or maybe not at all. Who knows? I mean, who tested it?... (I mean, recently...)

It's time -- long overdue as a matter of fact -- to let go of custom-sized pointers, near, far, whatever they are.

We're transitioning towards supporting modern C in a way that's API- and ABI-compatible to modern C++. For libpng-1.8.0, which is meant to be merely a stop-gap transition towards libpng-2, we might stick with a bunch of C89-isms. Or we might not. But `restrict` has to go for the same reason why `noalias` had to go. (*"This is non-negotiable"* -- R.I.P. Dennis Ritchie.) Mind you, we're still keeping the old pointer types, because OF COURSE we aren't breaking compatibility with the immense libpng application base; but that shouldn't stop us from modernizing and by that I mean simplifing the interface, the implementation, and most importantly, the manner in which the libpng users should be writing their apps.

---

@jbowler, @svgeesus, @ProgramMax, this is for preliminary review. The *actual* cleanup job will be coming up honking, you betcha!

### DIFF
--- a/contrib/libtests/pngimage.c
+++ b/contrib/libtests/pngimage.c
@@ -1,6 +1,6 @@
 /* pngimage.c
  *
- * Copyright (c) 2021-2024 Cosmin Truta
+ * Copyright (c) 2021-2025 Cosmin Truta
  * Copyright (c) 2015,2016 John Cunningham Bowler
  *
  * This code is released under the libpng license.
@@ -39,7 +39,7 @@
 /* 1.6.1 added support for the configure test harness, which uses 77 to indicate
  * a skipped test, in earlier versions we need to succeed on a skipped test, so:
  */
-#if PNG_LIBPNG_VER >= 10601 && defined(HAVE_CONFIG_H)
+#if defined(HAVE_CONFIG_H)
 #  define SKIP 77
 #else
 #  define SKIP 0
--- a/contrib/libtests/pngstest.c
+++ b/contrib/libtests/pngstest.c
@@ -1,6 +1,6 @@
 /* pngstest.c
  *
- * Copyright (c) 2021-2024 Cosmin Truta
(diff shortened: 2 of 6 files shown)


ProgramMax (approved):
LGTM

ProgramMax:
I understand this PR is a preview and the actual cleanup is coming later.
But this part looks good to me.

You might want to add something to the README.md or some documentation somewhere a comment like:
"MS-DOS supported ended on <date>. The latest version with support can be found at <link to previous commit>."

ctruta:
Everybody take note that I absolutely did not merge this commit. (I mean, yes, I did merge ***a*** commit, but not ***this*** commit.) Especially you, GitHub :-P

@ProgramMax wrote:

> I understand this PR is a preview and the actual cleanup is coming later.
> But this part looks good to me.

Thank you

> You might want to add something to the README.md or some documentation somewhere a comment like:
> "MS-DOS supported ended on . The latest version with support can be found at ."

Heads-up: I will ask a wilful LLM to write a doc about the evolution of libpng, past (up to 1.6.x), present (1.8.x) and future (2.x), based on the current documentation, the change log, and some of the messages that I sent here and there on GitHub. Then I will clean up the manuals, which contain quite a bit of historical references, but shouldn't.

PR #733, opened 2025-09-14
Add optional compile flag for global allocator override

d3x0r:
Add compile option `PNG_USER_GLOBAL_MEM_SUPPORTED`

which allows specifying  `png_set_mem_fn(NULL,...);` to specify functions to use for malloc/free, the additional data pointer passed is NULL (it could be a per-png data parameter, but being global, it'd unlikely the current PNG's data parameter would be useful anyway.  The first parameter `NULL` in this case is normally a `png_struct*`, and NULL was previously equivalent to doing nothing.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6279,6 +6279,10 @@ Version 1.6.50 [July 1, 2025]
   Fixed the CMake file for cross-platform builds that require `libm`.
 
 Version 1.6.51 [TODO]
+  Add option PNG_USER_GLOBAL_MEM_SUPPORTED to allow using png_set_mem_fn
+    with a NULL png_ptr to set a common malloc/free to use, overriding
+    the default `malloc()` and `free()`, if not
+    set on a specific png_ptr.
 
 Send comments/corrections/commendations to png-mng-implement at lists.sf.net.
 Subscription is required; visit
--- a/pngmem.c
+++ b/pngmem.c
@@ -18,6 +18,14 @@
 
 #include "pngpriv.h"
 
+#ifdef PNG_USER_GLOBAL_MEM_SUPPORTED
+static png_malloc_ptr
+png_global_malloc_fn;
+
+static png_free_ptr
+png_global_free_fn;
+#endif
+
 #if defined(PNG_READ_SUPPORTED) || defined(PNG_WRITE_SUPPORTED)
 /* Free a png_struct */
 void /* PRIVATE */
@@ -91,6 +99,11 @@ png_malloc_base,(png_const_structrp png_ptr, png_alloc_size_t size),
          return png_ptr->malloc_fn(png_constcast(png_structrp,png_ptr), size);
 #  else
       PNG_UNUSED(png_ptr)
+#  endif
+
+#  ifdef PNG_USER_GLOBAL_MEM_SUPPORTED
+      if (png_global_malloc_fn != NULL)
+         return png_malloc_fn(NULL,png_constcast(png_structrp,png_ptr), size);
 #  endif
 
    /* Use the system malloc */
@@ -247,6 +260,14 @@ png_free_default,(png_const_structrp png_ptr, png_voidp ptr),PNG_DEPRECATED)
       return;
 #endif /* USER_MEM */
 
+#ifdef PNG_USER_GLOBAL_MEM_SUPPORTED
+   if (png_ptr->free_fn != NULL)
+   {
+      png_global_free_fn(NULL, ptr);
+      return;
+   }
+#endif
+
    free(ptr);
 }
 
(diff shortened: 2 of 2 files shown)


jbowler (changes requested):
Lines 292 and 293 must be protected by a mutex or equivalent parallel processing mechanisms.  Please correct this.

ctruta:
Please don't mind me @d3x0r but I'm afraid I'm going to reject this proposal. You are proposing to augment the API without accompanying documentation or testing. Under the assumption that these concerns may be addressed, we have an even bigger architectural problem: libpng has an established principle of per-instance configuration.

The code is 100% thread-safe because there is no global state and I am strongly opposed to introducing any global state in the future. Global memory functions create invisible coupling between PNG operations that on the surface appear independent. That would make our code harder to reason about and debug. Moreover, if multiple libraries in the same process use libpng, they can't use different memory strategies simultaneously, thus breaking modularity.

@jbowler wrote:
> Lines 292 and 293 must be protected by a mutex or equivalent parallel processing mechanisms. Please correct this.

I'm opposed to this as well. Even with proper locking, global state complicates multi-threaded usage patterns and adds performance overhead.

The good news is that good solutions to this problem do exist. For example, applications can create helper functions that set memory functions on each `png_struct`.

In conclusion, I am rejecting this proposal for the core libpng. The complexity cost far exceeds the benefit. In my opinion, the existing per-instance libpng API should be sufficient for the use case presented here; and even if it's not, then wrapper-based solutions, specific to the applications' use case, are available as well.

d3x0r:
The documentation for the related function was removed.
It doesn't change the API, it makes a previous NOOP useful.

```
Version 1.2.3 [May 22, 2002]
  Revised libpng-config target in makefile.cygwin.
  Removed description of png_set_mem_fn() from documentation.
  Revised makefile.freebsd.
```

```
IX. Changes to Libpng from version 1.0.x to 1.2.x

Support for user memory management was enabled by default.  To
accomplish this, the functions png_create_read_struct_2(),
png_create_write_struct_2(), png_set_mem_fn(), png_get_mem_ptr(),
png_malloc_default(), and png_free_default() were added.

Support for the iTXt chunk has been enabled by default as of
version 1.2.41.
```

those are the only two references to the function in question. 

Edit: Originally I had complained about 'thread safe malloc' which I had recently run across an article that there really is no guarantee that malloc be thread safe.  The real issue was probably actually debug vs release runtime in visual studio... debug and release malloc don't play well together, even if each are using their appropriate runtimes only.

So really it's changing an undocumented feature that noone should be using anyway.

If someone passed NULL as the pngstruct they wanted to set the allocator for - then it would quickly fail afterward because they didn't actually have a pngstruct to pass.  If they do, nothing changes.

Nothing else in this requires a thread mutex - and most likely it's going to be set once in the init thread and never touched.

jbowler:
@ctruta: the real issue is that changing the "global" allocator allows one library to change the memory management of another library.  This would be a major security hole in Linux systems and others with the same DLL model (not Windows, Windows just crashes ;-)  I was being sarcastic when I pointed out that a mutex is required; it seemed easier and more productive to explain the obvious problem than the security issue, particularly as I don't have the faintest idea who is aware of the security issue.

d3x0r:
Re Mutex I'd say add a note in the documentation, that locking access to that if it's expected to change to a custom allocator per thread per image with some new entry point, because the extra data will be 0 because you're setting NULL anyway, which again, if the pngstruct is passed and that's intended then there is no difference since that still pre-empts what global specifies

But there is no documentation, so it's not even an API, so can it be broken? :)

If I could have make it static to the function somehow I'd have done that, there probably are about 0 static variables in libpng, and I expected more of an argument about implementing local static variables.

Regarding 'overriding malloc at will' just put a library with malloc closer than the c one.  Works for opengl hooks... but that's more dynamically loaded usually...

d3x0r:
And it's not even a compile option that enabled by default.

ctruta:
@d3x0r if my understanding of your use case is correct, then you should be able to use your custom allocators via `png_set_mem_fn`, which we already have in the libpng API. That's on top of `png_create_read_struct_2` and `png_create_write_struct_2`, which you can also use to enable your allocators from the get-go.

d3x0r:
> @d3x0r if my understanding of your use case is correct, then you should be able to use your custom allocators via `png_set_mem_fn`, which we already have in the libpng API. That's on top of `png_create_read_struct_2` and `png_create_write_struct_2`, which you can also use to enable your allocators from the get-go.

Using the simplified read/write system there is no access to the pngstruct that it uses, I have no opportunity to set the allocator.

ctruta:
> Using the simplified read/write system there is no access to the pngstruct that it uses, I have no opportunity to set the allocator.

Custom allocators are outside the scope of the simplified API by design. The simplified API is intended for common, straightforward use cases, and supporting specialized memory management would undermine that goal. Your use case is valid, but it falls into the category of advanced customization. For that, the correct and supported approach is to use the full libpng API, which provides the flexibility you need.

d3x0r:
> > Using the simplified read/write system there is no access to the pngstruct that it uses, I have no opportunity to set the allocator.
> 
> Custom allocators are outside the scope of the simplified API by design. The simplified API is intended for common, straightforward use cases, and supporting specialized memory management would undermine that goal. Your use case is valid, but it falls into the category of advanced customization. For that, the correct and supported approach is to use the full libpng API, which provides the flexibility you need.

when I use the full API with SETJMP support disabled, the library has a path that ends with abort(), which isn't even exit(1) which would give ma a chance to flush buffers.  So if I use the Full API, the modification is much more complicated and definitely breaks the API returning `int` instead of `void`.  

It doesn't work either way, this is a simple patch; but actually does need to be enabled by default or I'm just still always carrying a copy of the library instead of depending on what's in the system.

PR #734, opened 2025-09-19
Avoid potential out of boundary read in `png_warning`

stoeckmann:
It is possible to trigger an out of boundary read in `png_warning` if a warning message starts with a `#`, is no longer than 15 characters and contains no blank (` `).

That's quite specific and does not affect the libpng code, because no warning even starts with a `#` but it's probably best to protect this function exported through API.

The underlying problem is that `png_warning` in contrast to `png_error` strips error numbers if `ERROR_NUMBERS` is disabled because the preprocessor check accidentally covers only the if-condition, not the whole block.

The `ERROR_NUMBERS` feature is disabled by default and was probably never enabled. This out of boundary access still exists after merging this pull request, and actually some more and functional bugs within the `ERROR_NUMBERS` handler stay intact.

If you want, I can create a PR which fixes all these issues. But I am curious if there is actually interest in keeping this functionality. In fact, it's not possible to strip error messages and numbers even if `ERROR_NUMBERS` is compiled in because `png_set_strip_error_numbers` effectively does not allow me to activate this feature. In short: I highly doubt that anyone ever used it.

So, probably it would be best to remove the code. And that's why I keep this PR as small as possible -- otherwise I would fix dead code before it's removed. Split into two commits because the second one just cleans up.

Spotted while auditing the code, not through actual usage.

Proof of Concept (ideally compile with `-fsanitize=address`):
```
cat > poc.c << EOF
#include <err.h>
#include <png.h>
#include <setjmp.h>
#include <string.h>

int
main(int argc, char *argv[])
{
  if (argc != 3)
    errx(1, "usage: poc errmsg warnmsg");

  png_structp png_ptr = png_create_read_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);

  if (setjmp(png_jmpbuf(png_ptr))) {
    png_warning(png_ptr, argv[2]);
    png_destroy_read_struct(&png_ptr, NULL, NULL);
    return 1;
  }

  png_error(png_ptr, argv[1]);

  png_destroy_read_struct(&png_ptr, NULL, NULL);
  return 0;
}
EOF
cc -o poc poc.c -lpng -fsanitize=address
./poc 'error' '#########'
```

Output on my system is:
```
libpng error: error
libpng warning: =/bin/bash
```

The output `=/bin/bash` results from reading the `envp` area.

The proof of concept is more sophisticated than needed, but it might become handy if there is actually interest in fixing the `ERROR_NUMBERS` code.

### DIFF
--- a/pngerror.c
+++ b/pngerror.c
@@ -39,46 +39,6 @@ PNG_FUNCTION(void,PNGAPI
 png_error,(png_const_structrp png_ptr, png_const_charp error_message),
     PNG_NORETURN)
 {
-#ifdef PNG_ERROR_NUMBERS_SUPPORTED
-   char msg[16];
-   if (png_ptr != NULL)
-   {
-      if ((png_ptr->flags &
-         (PNG_FLAG_STRIP_ERROR_NUMBERS|PNG_FLAG_STRIP_ERROR_TEXT)) != 0)
-      {
-         if (*error_message == PNG_LITERAL_SHARP)
-         {
-            /* Strip "#nnnn " from beginning of error message. */
-            int offset;
-            for (offset = 1; offset<15; offset++)
-               if (error_message[offset] == ' ')
-                  break;
-
-            if ((png_ptr->flags & PNG_FLAG_STRIP_ERROR_TEXT) != 0)
-            {
-               int i;
-               for (i = 0; i < offset - 1; i++)
-                  msg[i] = error_message[i + 1];
-               msg[i - 1] = '\0';
-               error_message = msg;
-            }
-
-            else
-               error_message += offset;
-         }
-
-         else
-         {
-            if ((png_ptr->flags & PNG_FLAG_STRIP_ERROR_TEXT) != 0)
-            {
-               msg[0] = '0';
-               msg[1] = '\0';
-               error_message = msg;
-            }
-         }
-      }
-   }
-#endif
    if (png_ptr != NULL && png_ptr->error_fn != NULL)
       (*(png_ptr->error_fn))(png_constcast(png_structrp,png_ptr),
           error_message);
@@ -216,21 +176,6 @@ void PNGAPI
 png_warning(png_const_structrp png_ptr, png_const_charp warning_message)
 {
    int offset = 0;
-   if (png_ptr != NULL)
-   {
-#ifdef PNG_ERROR_NUMBERS_SUPPORTED
-   if ((png_ptr->flags &
-       (PNG_FLAG_STRIP_ERROR_NUMBERS|PNG_FLAG_STRIP_ERROR_TEXT)) != 0)
-#endif
-      {
-         if (*warning_message == PNG_LITERAL_SHARP)
-         {
-            for (offset = 1; offset < 15; offset++)
-               if (warning_message[offset] == ' ')
-                  break;
-         }
-      }
-   }
    if (png_ptr != NULL && png_ptr->warning_fn != NULL)
       (*(png_ptr->warning_fn))(png_constcast(png_structrp,png_ptr),
           warning_message + offset);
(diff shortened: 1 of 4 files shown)


jbowler:
Much though @greg may hate this it never was enabled and it simply does not work.  Delete, delete, delete.

My first rule: **remove code, don't add bugs**.

ctruta:
@stoeckmann wrote:

> If you want, I can create a PR which fixes all these issues. But I am curious if there is actually interest in keeping this functionality. In fact, it's not possible to strip error messages and numbers even if `ERROR_NUMBERS` is compiled in because `png_set_strip_error_numbers` effectively does not allow me to activate this feature. In short: I highly doubt that anyone ever used it.

I agree with not keeping this functionality, and I also agree with your supposition: this part of the code was likely **never tested**, **never used** and **always broken**.

> So, probably it would be best to remove the code. And that's why I keep this PR as small as possible -- otherwise I would fix dead code before it's removed. Split into two commits because the second one just cleans up.

Thank you @stoeckmann for discovering this and for bringing it to our attention.

We should do something about this in libpng-1.6.x because we cannot keep broken code in good conscience. I'm open to suggestions for an interim solution for v1.6.x. As for the upcoming v1.8.0, I think it's best to remove the entire `PNG_ERROR_NUMBERS_SUPPORTED` config flag and the associated feature altogether.

ctruta:
I pushed commit 465f654c68b0d8e6ef037028e30f6d003b8b6b6c to the branch `libpng18`.

Back to the branch `libpng16`, I don't know if we need to go as far as removing the `ERROR_NUMBERS` configuration option from `pnglibconf.dfa` and from elsewhere, because it might (rather gratuitously) break some user's workflow. This is something that kept happening to some of our users in the recent libpng versions, almost every time when we added new goodies to pnglibconf, and I don't expect this one to be incident-free either.

In conclusion, my inclination at this moment is to leave the configuration option (including the C macro `PNG_ERROR_NUMBERS_SUPPORTED`) in `libpng16`, but delete all the supporting C code and trigger an `#error` if someone ever tries to enable `ERROR_NUMBERS` accidentally or intentionally.

I'm open to alternative ideas.

stoeckmann:
I agree with:
- Keep configuration option
- Delete supporting C code

I would suggest though, to NOT add an `#error`. The functionality cannot be properly reached through API in any way, and `libpng16` is stable. If there are actually custom builds with this option, they keep working regardless of minor number releases.

Ironically, these custom builds are protected against this reported overflow, because the if-statement prevents the functionality from being reached. :)

ctruta:
> Ironically, these custom builds are protected against this reported overflow, because the if-statement prevents the functionality from being reached. :)

Ironically indeed :laughing:

So... doing nothing in v1.6.x is something that would be just as good as deleting C code that is compiled out anyway...?

stoeckmann:
> So... doing nothing in v1.6.x is something that would be just as good as deleting C code that is compiled out anyway...?

As far as I can tell, yes. I've tried to trigger it somehow but didn't find a way. That doesn't mean that there really is none, though.

Removing code would make sure that it really cannot do any harm.

ctruta:
> Removing code would make sure that it really cannot do any harm.

Ok, Tobias. I see your reasoning and I agree with your conclusion.

Right now I'm focused on tidying up version 1.8.0, though. Can you please help with a pull request, for v1.6.x only, following on 465f654c68b0d8e6ef037028e30f6d003b8b6b6c but leaving intact the config files, the flags, and the function `png_set_strip_error_numbers` as a no-op?

pngerror.c
```diff
@@ -712,42 +657,9 @@ png_default_error,(png_const_structrp png_ptr, png_const_charp error_message),
...
-   else
-#endif
-   {
-      fprintf(stderr, "libpng error: %s", error_message ? error_message :
-         "undefined");
-      fprintf(stderr, PNG_STRING_NEWLINE);
-   }
+   fprintf(stderr, "libpng error: %s", error_message ? error_message :
+      "undefined");
+   fprintf(stderr, PNG_STRING_NEWLINE);
```
stoeckmann, line 662: I have removed this extra block as you've done for png_default_warning in the libpng18 version. Would make sense to adjust it in libpng18, too.

stoeckmann:
> Right now I'm focused on tidying up version 1.8.0, though. Can you please help with a pull request, for v1.6.x only, following on [465f654](https://github.com/pnggroup/libpng/commit/465f654c68b0d8e6ef037028e30f6d003b8b6b6c) but leaving intact the config files, the flags, and the function `png_set_strip_error_numbers` as a no-op?

Adjusted. The remaining function is turned into a no-op, other code parts are removed and the option itself is left intact as well.

pngerror.c
ctruta: Ha! So I missed that one...

jbowler:
> I don't know if we need to go as far as removing the `ERROR_NUMBERS` configuration option from `pnglibconf.dfa` and from elsewhere, because it might (rather gratuitously) break some user's workflow.

There is no choice; there is no way of going through Glenn's process of deprecating something for several releases then farting around and not deleting anything with pnglibconf.dfa.

The fix is to delete the option.  Then wait until someone complains.  Easy and easy to put back if someone complains; it really is a one line change.  Duh.  This is so simple.

PR #736, opened 2025-09-29
refactor: Remove PNG_STRING_NEWLINE

stoeckmann:
Remove PNG_STRING_NEWLINE, which is supposed to be always \n.

Also, with removal of ERROR_NUMBERS feature, this block, which was a preprocessor conditional if-block, can be removed.

Call the fprintf functions directly instead.

### DIFF
--- a/png.c
+++ b/png.c
@@ -793,14 +793,12 @@ png_get_copyright(png_const_structrp png_ptr)
 #ifdef PNG_STRING_COPYRIGHT
    return PNG_STRING_COPYRIGHT
 #else
-   return PNG_STRING_NEWLINE \
-      "libpng version 1.8.0.git" PNG_STRING_NEWLINE \
-      "Copyright (c) 2018-2025 Cosmin Truta" PNG_STRING_NEWLINE \
-      "Copyright (c) 1998-2002,2004,2006-2018 Glenn Randers-Pehrson" \
-      PNG_STRING_NEWLINE \
-      "Copyright (c) 1996-1997 Andreas Dilger" PNG_STRING_NEWLINE \
-      "Copyright (c) 1995-1996 Guy Eric Schalnat, Group 42, Inc." \
-      PNG_STRING_NEWLINE;
+   return "\n"
+      "libpng version 1.8.0.git\n"
+      "Copyright (c) 2018-2025 Cosmin Truta\n"
+      "Copyright (c) 1998-2002,2004,2006-2018 Glenn Randers-Pehrson\n"
+      "Copyright (c) 1996-1997 Andreas Dilger\n"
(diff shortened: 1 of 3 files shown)


ctruta:
Many thanks for all your contributions, @stoeckmann. As you can see in this announcement (https://github.com/pnggroup/libpng/discussions/735) I'm doing refactorings and cleanups in preparation to the upcoming libpng-1.8.0 release. Should you ever notice any other cleanup opportunities, I will greatly appreciate it if you can please bring them to my attention; and even more so, if you bring a pull request 😉

stoeckmann:
I have added PRs for minor improvements which I've noticed while inspecting pngerror.c.

ctruta:
> I have added PRs for minor improvements which I've noticed while inspecting pngerror.c.

Thank you so much, Tobias. I'll review them tomorrow.

jbowler:
The **reason** that this code is there is because in some antiquated view of the world the newline character could be something other than line-feed, for example it might be required to issue a carriage return so that the line printer head didn't print the next line one character over the other at the end of the line.  (This did happen.)

@ctruta; as you observed elsewhere this change just might affect (as in piss off) someone who was using it to change the newline character into "\r\n" or maybe "\r".

Bear that in mind.  It's safe in 1.8 because obviously people are going to test 1.8 before it is released, or not.  I suggest letting the horse loose and, in fact, reformatting the whole dang thing.  Not just a couple of curly brackets at a time (the first change).  Let some willing volunteer reformat everything, just like Glenn.  Free rein.

ctruta:
> The **reason** that this code is there is because in some antiquated view of the world the newline character could be something other than line-feed, for example it might be required to issue a carriage return so that the line printer head didn't print the next line one character over the other at the end of the line. (This did happen.)

That is a believable story, but an accompanying comment on why such a glaring antipattern is there at all would have helped.

> @ctruta; as you observed elsewhere this change just might affect (as in piss off) someone who was using it to change the newline character into "\r\n" or maybe "\r".

We're moving to C99. Perhaps not as quickly as I would personally like, but we're getting there. Considering that opening files in text mode with `fopen(..., "w")` vs. `fopen(..., "wb")` exists for a reason, and that was sorted out as early as ANSI C in 1989, it's on them to get a compiler and standard library that is reasonably up-to-century, if not up-to-date. Alternatively, the older versions of libpng are still out there. Alternatively, they can maintain their own forks. Alternatively, ...

Solutions exist.

FYI I'm planning to implement multithreaded zlib compression and decompression in libpng-2, in a manner in which Mark Adler did it in pigz, because we stayed behind for far too long. I anticipate this other problem to be far less of a problem at that time.

PR #737, opened 2025-09-30
Simplify pngerror.c functions

stoeckmann:
Remove unneeded `offset` variable in `png_warning` and turn a `while` loop into a `for` loop in `png_format_buffer.

The `offset` is always 0 and the `while` loop is used a `for` loop.

### DIFF
--- a/pngerror.c
+++ b/pngerror.c
@@ -380,13 +380,12 @@ png_format_buffer(png_const_structrp png_ptr, png_charp buffer, png_const_charp
     error_message)
 {
    png_uint_32 chunk_name = png_ptr->chunk_name;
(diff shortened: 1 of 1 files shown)


jbowler:
@ctruta: reject, two unrelated changes in one PR, two pointless changes because the compiler can do the same, one confusing change (the first) because the C for loop termination conditions are notoriously difficult to remember.  While loops are always better simply because they are explicit.

C++ and python and other languages fix the latter with the "for ... in" construct, which makes sense rather than a ridiculous language construct which puts a Turing equivalent piece of code into every construct.  Blugh, vomit, I hate C.

PR #738, opened 2025-09-30
Improve documentation in pngerror.c

stoeckmann:
Fixes a few typos and properly describes what kind of functions are located in pngerror.c.

Also, the implementation of a while loop in `png_formatted_warning` changed so much that the descriptive comment didn't fit anymore. Adjusted the comment.

### DIFF
--- a/pngerror.c
+++ b/pngerror.c
@@ -1,4 +1,4 @@
-/* pngerror.c - stub functions for i/o and memory allocation
+/* pngerror.c - functions for warnings and error handling
  *
  * Copyright (c) 2018-2025 Cosmin Truta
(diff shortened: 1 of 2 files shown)


jbowler:
@ctruta: reject.  Regular human beings can read misspelt English.  Not an appropriate change for a minor release.

ctruta (approved):
@jbowler wrote:

> @ctruta: reject.

Under the principle _"leave a codebase in a state that's as good as or better than the one you initially found it"_, I approve this PR.

> Regular human beings can read misspelt English. Not an appropriate change for a minor release.

I disagree. Readability matters. My professors never let me get away with submitting a paper with misspellings, sloppy phrases, or other forms of bad prose. Libpng may not be a conference paper, but that doesn't mean we should hold it to a lower standard.

PR #739, opened 2025-09-30
Reduce compiled library size with pngerror.o reduction

stoeckmann:
At least GCC 15 is not smart enough to detect that `png_digits` is a subset of `digits`. Both are static variables containing `0123456789ABCDEF`, yet `digits` ends with a `\0` while `png_digits` does not.

Defining `png_digits` the same way as `digits` leads to proper removal (32 bytes saved on my system).

Also, remove the `123456789` string when resolving parameter numbers in `png_formatted_warning`. It doesn't take a loop and thus further reduces size and also amount of instructions.

In total, 105 bytes were removed on my system.

### DIFF
--- a/pngerror.c
+++ b/pngerror.c
@@ -250,20 +250,13 @@ png_formatted_warning(png_const_structrp png_ptr, png_warning_parameters p,
        */
       if (p != NULL && *message == '@' && message[1] != '\0')
       {
-         int parameter_char = *++message; /* Consume the '@' */
-         static const char valid_parameters[] = "123456789";
-         int parameter = 0;
-
-         /* Search for the parameter digit, the index in the string is the
-          * parameter to use.
-          */
-         while (valid_parameters[parameter] != parameter_char &&
-            valid_parameters[parameter] != '\0')
-            ++parameter;
+         const int parameter_char = *++message; /* Consume the '@' */
 
          /* If the parameter digit is out of range it will just get printed. */
-         if (parameter < PNG_WARNING_PARAMETER_COUNT)
+         if (parameter_char >= '1' && parameter_char <= '9')
          {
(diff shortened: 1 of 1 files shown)


jbowler:
Yes, trailing substring sharing was a feature of the original UNIX implementation; there was a special program to do it!  Yet, of course, leading substring sharing doesn't work because of the implicit length-by-nul assumption in C.

Similarly your change does not work because you are assuming the "sizeof" operator is not applied to the shared static.  Prove me wrong!

ctruta (approved):
@jbowler if the compiler can identify two identical objects (which, in this case, are two indentical string literals) inside a translation unit, it is a legitimate optimization to merge them together -- except for the edge case where the code semantics actually requires them to be at different addresses, which is NOT the case in this case.

This is a good change. And there's a way to make it even better by commonizing these two static const objects. (Because DRY.)

ctruta:
@stoeckmann excellent idea, which I took one step further. See commit b719408de0ce927da963b6afc4bb75a35701f25e.

Many thanks for all of your contributions :tada:

stoeckmann:
Awesome! I thought about it but then dismissed the idea due to these different preprocessor conditions and because the compiler was smart enough to merge it on its own.

So, definitely agree. :)

ctruta:
> Awesome! I thought about it but then dismissed the idea due to these different preprocessor conditions and because the compiler was smart enough to merge it on its own.

I am, myself, extremely conservative when it comes to making changes from revision to revision. On the other hand, this upcoming v1.8.0 is one of those rare opportunities to do those kinds of refactorings and cleanups that wouldn't normally be possible. We won't be touching the API, but everything else is fair game under the slogan _"now's the time"_.

PR #740, opened 2025-10-02
fix(msan): Fix MSan false positives in Arm NEON

zeriyoshi:
MemorySanitizer in LLVM Clang is still incomplete for Arm NEON SIMD and frequently produces false positives. This PR addresses the false positives I found by adding the `no_sanitize("memory")` attribute.

Looking at the project, it appears that MemorySanitizer testing is not currently being performed, so I have not added any specific tests.

Environment information:
- platform: arm64 (MacBook Air M4 with OrbStack)
- distro: Debian Trixie 13 (on Container)
- clang: 21.1.3 (from LLVM repository)

### DIFF
--- a/arm/filter_neon_intrinsics.c
+++ b/arm/filter_neon_intrinsics.c
@@ -46,6 +46,21 @@
 
 #if PNG_ARM_NEON_OPT > 0
 
+/* MemorySanitizer (MSan) has known compatibility issues with ARM NEON intrinsics,
+ * causing false positives for uninitialized memory. The no_sanitize attribute
+ * suppresses these false warnings for NEON-optimized filter functions.
+ */
+#if defined(__clang__) && defined(__has_attribute)
+#  if __has_attribute(no_sanitize)
+#    define PNG_MSAN_NEON_SAFE __attribute__((no_sanitize("memory")))
+#  else
+#    define PNG_MSAN_NEON_SAFE
+#  endif
+#else
+#  define PNG_MSAN_NEON_SAFE
+#endif
+
+PNG_MSAN_NEON_SAFE
 void
 png_read_filter_row_up_neon(png_row_infop row_info, png_bytep row,
    png_const_bytep prev_row)
@@ -67,6 +82,7 @@ png_read_filter_row_up_neon(png_row_infop row_info, png_bytep row,
    }
 }
 
+PNG_MSAN_NEON_SAFE
 void
 png_read_filter_row_sub3_neon(png_row_infop row_info, png_bytep row,
    png_const_bytep prev_row)
@@ -114,6 +130,7 @@ png_read_filter_row_sub3_neon(png_row_infop row_info, png_bytep row,
    PNG_UNUSED(prev_row)
 }
 
+PNG_MSAN_NEON_SAFE
 void
 png_read_filter_row_sub4_neon(png_row_infop row_info, png_bytep row,
    png_const_bytep prev_row)
@@ -146,6 +163,7 @@ png_read_filter_row_sub4_neon(png_row_infop row_info, png_bytep row,
    PNG_UNUSED(prev_row)
 }
 
+PNG_MSAN_NEON_SAFE
 void
 png_read_filter_row_avg3_neon(png_row_infop row_info, png_bytep row,
    png_const_bytep prev_row)
@@ -214,6 +232,7 @@ png_read_filter_row_avg3_neon(png_row_infop row_info, png_bytep row,
    }
 }
 
+PNG_MSAN_NEON_SAFE
 void
 png_read_filter_row_avg4_neon(png_row_infop row_info, png_bytep row,
    png_const_bytep prev_row)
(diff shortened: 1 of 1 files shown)


ctruta (changes requested):

I'm requesting changes before this can be merged, as the current approach is too coarse-grained and lacks the information required to verify these are genuine false positives rather than real bugs.

The core issue here is that using `__attribute__((no_sanitize("memory")))` disables **all** MSan checks for the entire function. That will hide real uninitialized memory bugs and will make future regressions invisible.

As I understand, the false positives are occurring in compiler intrinsics. LLVM's MSan should correctly instrument standard `<arm_neon.h>` intrinsics, so if it's producing false positives in LLVM 21.1.3, this indicates either (1) a buffer over-read pattern in which ARM Neon loads reading past buffer boundaries, or (2) an LLVM bug which should be reported upstream.

Could you please provide the complete MSan error message with the associated stack trace? I need to see what's the exact problem being reported, and where.

## Recommended fix #1: surgical unpoisoning

Instead of disabling the sanitizer entirely, use targeted `__msan_unpoison()` calls. For example:

```c
#if defined(__has_feature)
#  if __has_feature(memory_sanitizer)
#    include <sanitizer/msan_interface.h>
#  endif
#endif

void png_read_filter_row_sub4_neon(png_row_infop row_info, png_bytep row) {
    png_size_t rowbytes = row_info->rowbytes;

    #if defined(__has_feature) && __has_feature(memory_sanitizer)
    /* ARM Neon vector loads (vld1q_u8) may read up to 15 bytes past rowbytes
     * to complete 16-byte SIMD loads. This is safe because libpng allocates
     * rows with adequate padding. Unpoison the over-read region.
     *
     * LLVM 21.1.3 MSan on ARM64 reports this as uninitialized read.
     *
     * TODO: Retest with LLVM 22+ and remove if fixed upstream.
     */
    __msan_unpoison(row + rowbytes, 15);
    #endif

    /* ... */
}
```

## Recommended fix #2: suppression file

If multiple functions are affected, you should consider using MSan's suppression mechanism:

```sh
MSAN_OPTIONS=suppressions=scripts/msan_suppressions.txt
```

## Reporting the LLVM defect (if applicable)

If this is a genuine LLVM defect, and not over-read patterns, please file an issue with the LLVM project, then link that issue in this PR so we can track the landing of the upstream fix.

zeriyoshi:
Thank you for the review. This is almost certainly an LLVM issue, but due to my lack of knowledge, I don't yet understand how to report it properly. Please give me some time.

In the meantime, what do you think about implementing the mitigation you suggested (suppressing the false positive using `msan_unpoison`)?

ctruta:
Thank you, @zeriyoshi, and don't worry about your lack of knowledge regarding LLVM. We're all learners. Given that I, too, am interested in seeing this resolved properly, I'm willing to help you as much as I can.

About implementing the `__msan_unpoison()` approach, I would personally favour the suppression file solution instead, because that's less intrusive. (Please note that the suppression file requires users to set the `MSAN_OPTIONS` environment variable when running MSan-instrumented builds, so it's effective for development/CI but -- admittedly -- it won't automatically help downstream users who build libpng with MSan.) If that's acceptable for your use case, let's try the suppression file first. Otherwise, we can use `__msan_unpoison()`.

However, before we proceed, I still need to see the actual MSan error messages. This is critical, because the fix location and the scope depend on what MSan is complaining about. Can you please run your test case with MSan enabled and share the output? It should look like this:

```
==12345==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x... in png_read_filter_row_sub4_neon filter_neon_intrinsics.c:142
    #1 0x... in png_read_filter_row ...
```

I presume the LLVM developers will expect to see the actual failure as well, even more than we do. They will need to know **exactly** what's broken (if something is indeed broken) so that they know **exactly** what needs fixing, and why. My recommendation to you is either one of the following:

- Option 1: Paste the MSan output here in **pnggroup/libpng** (i.e. right here in this issue), and then submit a defect report to the LLVM developers, telling them about MSan's false positive, and give them the link (<https://github.com/pnggroup/libpng/pull/740>).

- Option 2: Submit a defect report to the LLVM developers directly, telling them about MSan's false positive in libpng, and paste the MSan output in that defect report at LLVM. (In my opinion, this is the better option, because they will of course need to know that it's happening in libpng but without actually requiring them to browse into libpng's development.)

In any case, please remember to share the complete MSan error output (with `-fsanitize-memory-track-origins=2`). I can help you determine exactly what to put in the suppression file and/or where to place the `__msan_unpoison()` calls, if needed.

vrabaud:
Hi, did you try the latest  LLVM (maybe even HEAD)? All patches have been merged for MSAN, cf https://github.com/llvm/llvm-project/issues/72848#issuecomment-2448672477 If any function is missing, please re-open that bug.

ctruta:
I agree with @vrabaud: letting the LLVM developers know about MSan reporting false positives (if that's still the case) is the proper course of action.

PR #743, opened 2025-10-13
api: Deprecate legacy pointer typedefs

ctruta:
@ProgramMax, @svgeesus, @jbowler,
This is the second iteration of my API cleanup work. I have two commits so far, a little one and a big one. An even bigger one is coming up after these two.

### DIFF
--- a/manuals/libpng-manual.txt
+++ b/manuals/libpng-manual.txt
@@ -376,8 +376,8 @@ The structure allocation functions quietly return NULL if they fail to
 create the structure, so your application should check for that.
 
     png_structp png_ptr = png_create_read_struct
-        (PNG_LIBPNG_VER_STRING, (png_voidp)user_error_ptr,
-        user_error_fn, user_warning_fn);
+       (PNG_LIBPNG_VER_STRING,
+        user_error_ptr, user_error_fn, user_warning_fn);
 
     if (!png_ptr)
        return ERROR;
@@ -395,8 +395,8 @@ use a libpng that was built with PNG_USER_MEM_SUPPORTED defined, and use
 png_create_read_struct_2() instead of png_create_read_struct():
 
     png_structp png_ptr = png_create_read_struct_2
-        (PNG_LIBPNG_VER_STRING, (png_voidp)user_error_ptr,
-        user_error_fn, user_warning_fn, (png_voidp)
+       (PNG_LIBPNG_VER_STRING,
+        user_error_ptr, user_error_fn, user_warning_fn,
         user_mem_ptr, user_malloc_fn, user_free_fn);
 
 The error handling routines passed to png_create_read_struct()
@@ -1174,12 +1174,12 @@ with
 
 where row_pointers is an array of pointers to the pixel data for each row:
 
-   png_bytep row_pointers[height];
+   png_byte *row_pointers[height];
 
 If you know your image size and pixel size ahead of time, you can allocate
 row_pointers prior to calling png_read_png() with
 
-   if (height > PNG_UINT_32_MAX / (sizeof (png_bytep)))
+   if (height > PNG_UINT_32_MAX / (sizeof (png_byte *)))
       png_error(png_ptr,
           "Image is too tall to process in memory");
 
@@ -1187,15 +1187,13 @@ row_pointers prior to calling png_read_png() with
       png_error(png_ptr,
           "Image is too wide to process in memory");
 
-   row_pointers = png_malloc(png_ptr,
-       height*(sizeof (png_bytep)));
+   row_pointers = png_malloc(png_ptr, height*(sizeof (png_byte *)));
 
    for (int i = 0; i < height, i++)
       row_pointers[i] = NULL;  /* security precaution */
 
    for (int i = 0; i < height, i++)
-      row_pointers[i] = png_malloc(png_ptr,
-          width*pixel_size);
+      row_pointers[i] = png_malloc(png_ptr, width*pixel_size);
 
    png_set_rows(png_ptr, info_ptr, &row_pointers);
 
@@ -1207,8 +1205,7 @@ be sure that your platform is able to allocate such a large buffer:
    if (height > PNG_SIZE_MAX/(width*pixel_size))
       png_error(png_ptr, "image_data buffer would be too large");
 
-   png_bytep buffer = png_malloc(png_ptr,
-      height*width*pixel_size);
+   png_byte *buffer = png_malloc(png_ptr, height*width*pixel_size);
 
    for (int i = 0; i < height, i++)
       row_pointers[i] = buffer + i*width*pixel_size;
@@ -2055,7 +2052,7 @@ histogram, it may not do as good a job.
       if (png_get_valid(png_ptr, info_ptr,
           PNG_INFO_PLTE))
       {
-         png_uint_16p histogram = NULL;
+         png_uint_16 *histogram = NULL;
 
          png_get_hIST(png_ptr, info_ptr,
              &histogram);
@@ -2110,8 +2107,8 @@ with
 
 You must supply the function
 
-    void read_transform_fn(png_structp png_ptr, png_row_infop
-        row_info, png_bytep data)
+    void read_transform_fn(png_structp png_ptr,
+        png_row_infop row_info, png_byte *data)
 
 See pngtest.c for a working example.  Your function will be called
 after all of the other transformations have been processed.  Take care with
@@ -2151,7 +2148,7 @@ freeing any memory required for the user structure.
 You can retrieve the pointer via the function
 png_get_user_transform_ptr().  For example:
 
-    voidp read_user_transform_ptr =
+    void *read_user_transform_ptr =
         png_get_user_transform_ptr(png_ptr);
 
 The last thing to handle is interlacing; this is covered in detail below,
@@ -2216,7 +2213,7 @@ of that other stuff necessary with png_read_rows().
 
 where row_pointers is:
 
-   png_bytep row_pointers[height];
+   png_byte *row_pointers[height];
 
 You can point to void or char or whatever you use for pixels.
 
@@ -2232,7 +2229,7 @@ where row_pointers is the same as in the png_read_image() call.
 If you are doing this just one row at a time, you can do this with
 a single row_pointer instead of an array of row_pointers:
 
-    png_bytep row_pointer = row;
+    png_byte *row_pointer = row;
     png_read_row(png_ptr, row_pointer, NULL);
 
 If the file is interlaced (interlace_type != 0 in the IHDR chunk), things
@@ -2542,8 +2539,8 @@ png_infop info_ptr;
  initialize_png_reader()
  {
     png_ptr = png_create_read_struct
-        (PNG_LIBPNG_VER_STRING, (png_voidp)user_error_ptr,
-         user_error_fn, user_warning_fn);
+       (PNG_LIBPNG_VER_STRING,
+        user_error_ptr, user_error_fn, user_warning_fn);
 
     if (!png_ptr)
         return ERROR;
@@ -2587,7 +2584,7 @@ png_infop info_ptr;
  /* A code fragment that you call as you receive blocks
    of data */
  int
- process_data(png_bytep buffer, png_uint_32 length)
+ process_data(png_byte *buffer, png_uint_32 length)
  {
     if (setjmp(png_jmpbuf(png_ptr)))
     {
@@ -2654,7 +2651,7 @@ png_infop info_ptr;
  /* This function is called when each row of image
     data is complete */
  void
- row_callback(png_structp png_ptr, png_bytep new_row,
+ row_callback(png_structp png_ptr, png_byte *new_row,
     png_uint_32 row_num, int pass)
  {
     /* If the image is interlaced, and you turned
@@ -2747,8 +2744,8 @@ both "png_ptr"; you can call them anything you like, such as
 "read_ptr" and "write_ptr".  Look at pngtest.c, for example.
 
     png_structp png_ptr = png_create_write_struct
-       (PNG_LIBPNG_VER_STRING, (png_voidp)user_error_ptr,
-        user_error_fn, user_warning_fn);
+       (PNG_LIBPNG_VER_STRING,
+        user_error_ptr, user_error_fn, user_warning_fn);
 
     if (!png_ptr)
        return ERROR;
@@ -2765,8 +2762,8 @@ define PNG_USER_MEM_SUPPORTED and use
 png_create_write_struct_2() instead of png_create_write_struct():
 
     png_structp png_ptr = png_create_write_struct_2
-       (PNG_LIBPNG_VER_STRING, (png_voidp)user_error_ptr,
-        user_error_fn, user_warning_fn, (png_voidp)
+       (PNG_LIBPNG_VER_STRING,
+        user_error_ptr, user_error_fn, user_warning_fn,
         user_mem_ptr, user_malloc_fn, user_free_fn);
 
 After you have these structures, you will need to set up the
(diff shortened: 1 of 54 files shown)


ctruta:
@stoeckmann, FYI, as general rule, all the commits that we send out for review are not just for members in @pnggroup but for anyone who wants to take a look and share an opinion. Considering your recent contributions, all of which were very thorough, I thought I should ping you individually and kindly invite you to contribute a review as well, if possible.

manuals/libpng-manual.txt
```diff
@@ -4070,24 +4066,22 @@ READ APIs
...
    the png_controlp field 'opaque' to NULL (or, better, memset the whole thing.)
 
-   int png_image_begin_read_from_file( png_imagep image,
-     const char *file_name)
+   int png_image_begin_read_from_file(png_image *image, const char *file_name)
 
-     The named file is opened for read and the image header
-     is filled in from the PNG header in the file.
+      The named file is opened for read and the image header
+      is filled in from the PNG header in the file.
```
ProgramMax: Just checking that these leading spaces were intentional.

pngread.c
```diff
@@ -954,7 +954,7 @@ png_set_read_status_fn(png_structrp png_ptr, png_read_status_ptr read_row_fn)
 #ifdef PNG_INFO_IMAGE_SUPPORTED
 void
 png_read_png(png_structrp png_ptr, png_inforp info_ptr,
-    int transforms, voidp params)
+    int transforms, png_voidp params)
```
ProgramMax: If png_voidp is deprecated now, should this just be void*?

pngwrite.c
```diff
@@ -1372,7 +1372,7 @@ png_set_write_user_transform_fn(png_structrp png_ptr, png_user_transform_ptr
 #ifdef PNG_INFO_IMAGE_SUPPORTED
 void
 png_write_png(png_structrp png_ptr, png_inforp info_ptr,
-    int transforms, voidp params)
+    int transforms, png_voidp params)
```
ProgramMax: Same question here.

manuals/libpng.3
```diff
@@ -4778,25 +4771,22 @@ functions are normally supplied at the time that the png_struct is created.
...
 
-    void user_warning_fn(png_structp png_ptr,
-        png_const_charp warning_msg);
+    void user_warning_fn(png_struct *png_ptr, const char *warning_msg);
 
 Then, within your user_error_fn or user_warning_fn, you can retrieve
 the error_ptr if you need it, by calling
 
-    png_voidp error_ptr = png_get_error_ptr(png_ptr);
+    void *error_ptr = png_get_error_ptr(png_ptr);
```
ProgramMax, line 4791: Just pointing out that elsewhere we're switching from png_voidp to void*.

manuals/libpng.3
```diff
@@ -901,10 +901,10 @@ If you want to use your own memory allocation routines,
...
 png_create_read_struct_2() instead of png_create_read_struct():
 
-    png_structp png_ptr = png_create_read_struct_2
-        (PNG_LIBPNG_VER_STRING, (png_voidp)user_error_ptr,
-        user_error_fn, user_warning_fn, (png_voidp)
-        user_mem_ptr, user_malloc_fn, user_free_fn);
+    png_struct *png_ptr = png_create_read_struct_2
+        (PNG_LIBPNG_VER_STRING,
+         user_error_ptr, user_error_fn, user_warning_fn,
+         user_mem_ptr, user_malloc_fn, user_free_fn);
```
ProgramMax: Were these extra spaces intentional?

manuals/libpng.3
```diff
@@ -3271,10 +3268,10 @@ If you want to use your own memory allocation routines,
...
 png_create_write_struct_2() instead of png_create_write_struct():
 
-    png_structp png_ptr = png_create_write_struct_2
-       (PNG_LIBPNG_VER_STRING, (png_voidp)user_error_ptr,
-        user_error_fn, user_warning_fn, (png_voidp)
-        user_mem_ptr, user_malloc_fn, user_free_fn);
+    png_struct *png_ptr = png_create_write_struct_2
+        (PNG_LIBPNG_VER_STRING,
+         user_error_ptr, user_error_fn, user_warning_fn,
+         user_mem_ptr, user_malloc_fn, user_free_fn);
```
ProgramMax: Checking on these leading spaces, too.

manuals/libpng.3
```diff
@@ -4577,24 +4573,22 @@ READ APIs
...
    the png_controlp field 'opaque' to NULL (or, better, memset the whole thing.)
 
-   int png_image_begin_read_from_file( png_imagep image,
-     const char *file_name)
+   int png_image_begin_read_from_file(png_image *image, const char *file_name)
 
-     The named file is opened for read and the image header
-     is filled in from the PNG header in the file.
+      The named file is opened for read and the image header
+      is filled in from the PNG header in the file.
```
ProgramMax: And here, too.

ProgramMax (approved):
In general, LGTM.
A few comments though.

manuals/libpng-manual.txt
ctruta: The leading spaces in the manuals are in general very inconsistent globally, and that will be a separate cleanup job that's coming up. It's just that these two have been inconsistent globally *and* locally, and the best I could do right now is just adjust them in-place.

pngread.c
ctruta: The commits are arranged in this particular order:

1. [fix: Use png_voidp instead of voidp in pngread.c and pngwrite.c](https://github.com/pnggroup/libpng/pull/743/commits/be74b298591ee733330cdabab944da839417c399)
2. [api: Deprecate legacy pointer typedefs](https://github.com/pnggroup/libpng/pull/743/commits/7628b30b02b3b7c184a436aab0a8a2f00b62c920)

This means:

1. Oops I caught a quarter-century-old bug, so let me fix it real quick by replacing `voidp` with `png_voidp` in pngread.c and pngwrite.c
2. Never mind, I'll replace **all** `png_voidp` with `void *`, and **all** `png_const_voidp` with `const void *`, and **all** `png_blahp` with `png_blah *`, etc., because all of these pointer type abstractions were clearly broken for a quarter-century with nobody noticing.

pngwrite.c
ctruta: It was the same error in both pngread.c and pngwrite.c which I fixed in the same way in commit [fix: Use png_voidp instead of voidp in pngread.c and pngwrite.c](https://github.com/pnggroup/libpng/pull/743/commits/be74b298591ee733330cdabab944da839417c399)

manuals/libpng.3
ctruta: We're switching from `png_voidp` to `void *` everywhere in this second commit, including the initial "fixing" from pngread.c and pngwrite.c.

manuals/libpng.3
ctruta: Yeah... tidying up the libpng manuals is going to be a whole different story. My recommendation: it's best to check out the libpng repository on your local machine. Inspect the user manual files, both `libpng.3` and `libpng-manual.txt`, and ponder the magnitude of the cleanup job which is about to follow.

manuals/libpng.3
ctruta: Yep -- same response as above.

ctruta:
Thanks for your review, @ProgramMax. I'm switching off the lights now, without shipping these commits just yet, to give an extra opportunity to anybody else who might want to review or comment. For tomorrow I'm planning to follow up with the cleanup of the entire source tree. Exciting day ahead: I expect that new thing to become the biggest and giant-est commit I'll have ever landed.

manuals/libpng-manual.txt
ProgramMax: Sounds good.

png.h
```diff
@@ -343,46 +343,43 @@ extern "C" {
...
+typedef png_struct * png_structp;                           /* [Deprecated] */
+typedef png_struct * * png_structpp;                        /* [Deprecated] */
+typedef png_struct * PNG_RESTRICT png_structrp;             /* [Deprecated] */
+typedef const png_struct * PNG_RESTRICT png_const_structrp; /* [Deprecated] */
+
+/* png_info contains information read from or to be written to a PNG file.
+ * One or more of these must exist while reading or creating a PNG file.
+ * The information is not used by libpng during read, but is used to control
+ * what gets written when a PNG file is created.  "png_get_" function calls
+ * read information during decoding, while "png_set_" functions calls write
```
stoeckmann: It's already a typo in the original, so this is merely a remark (unless you want to "silently" fix it here): "functions calls" should be "function calls" to also stay in sync with the `png_get_` term one line above.

png.h
```diff
@@ -343,46 +343,43 @@ extern "C" {
...
+ * information during encoding.
+ */
+typedef struct png_info_def png_info; /* [Opaque] */
+
+typedef png_info * png_infop;                           /* [Deprecated] */
+typedef const png_info * png_const_infop;               /* [Deprecated] */
+typedef png_info * * png_infopp;                        /* [Deprecated] */
+typedef png_info * PNG_RESTRICT png_inforp;             /* [Deprecated] */
+typedef const png_info * PNG_RESTRICT png_const_inforp; /* [Deprecated] */
```
stoeckmann: Sometimes you add `[Deprecated]` at the end of a line, even with such a block of deprecated typedefs, ...

pngconf.h
```diff
@@ -463,7 +463,7 @@ typedef ptrdiff_t png_ptrdiff_t;
  */
 typedef png_int_32 png_fixed_point;
 
-/* Add typedefs for pointers */
+/* [Deprecated] */
```
stoeckmann, line 466: ... while other times the `[Deprecated]` comment is at the start of a block of deprecated typedefs.

stoeckmann (commented):
Looks good to me and I like the approach. Just two things I noticed which are really just comments, nothing that has to be changed.

png.h
ctruta: Incorporated. Thanks for the catch.

pngconf.h
ctruta: I will clean it up later in some form or another. Right now I'm still thinking whether I should append `PNG_DEPRECATED` attributes, or leave the `/* [Deprecated] */` comments in place (and make them consistent), or do something else.

ctruta:
Thank you for your previous reviews -- more help wanted!

I landed the previous two commits, with several minor changes and a rather major one: I removed the updates to the libpng manuals and I moved those updates into the current commit. I think they are a better fit in this one.

As you can see in the commit message, this work is still in progress. I'm aiming to make the commits small enough to be self-contained, and sometimes I actually succeed. This last commit, on the other hand, is neither small nor complete, and I'll see what to do about it.

pngtrans.c
```diff
@@ -451,14 +451,14 @@ static const png_byte fourbppswaptable[256] = {
...
+png_do_packswap(png_row_info *row_info, png_byte *row)
 {
    png_debug(1, "in png_do_packswap");
 
    if (row_info->bit_depth < 8)
    {
-      png_bytep rp;
-      png_const_bytep end, table;
+      png_byte *rp;
+      const png_byte *end, * table;
```
ProgramMax: This is an interesting change.

I believe it is correct. But I want to point it out just because it is interesting. So no action is needed here.
The typedef includes the pointer-ness of the type. So 'table' is also a pointer.
But when the typedef is not used, the pointer-ness is separate and must be restated for 'table' to continue to be a pointer.

pngread.c
```diff
@@ -3203,7 +3203,7 @@ png_image_read_colormapped(png_voidp argument)
       while (--passes >= 0)
       {
          png_uint_32      y = image->height;
-         png_bytep        row = png_voidcast(png_bytep, display->first_row);
+         png_byte *row = png_voidcast(png_byte *, display->first_row);
```
ProgramMax: Previously, this line had extra spaces added to align with the line above.
This change removes that.

However, the alignment maybe didn't make sense anyway??

We should probably also remove those extra spaces in a separate commit.

pngwrite.c
```diff
@@ -1577,17 +1577,17 @@ png_image_write_init(png_imagep image)
...
-   png_const_voidp first_row;
+   const void *first_row;
    ptrdiff_t       row_bytes;
-   png_voidp       local_row;
+   void *local_row;
    /* Byte count for memory writing */
-   png_bytep        memory;
+   png_byte *memory;
    png_alloc_size_t memory_bytes; /* not used for STDIO */
    png_alloc_size_t output_bytes; /* running total */
```
ProgramMax: These lines also had spaces for alignment.
The updated lines are no longer aligned.

Just marking them here for us to reference when we remove alignment from the others in a separate commit.

ProgramMax (approved):
LGTM.
I marked some places for us to change alignment in a separate commit / pull request.

pngwrite.c
ctruta: I made a separate big giant whitespace-only commit with the express purpose to move out of the way the formatting issues arising from the actual refactoring work. See caa0c733f32661edf8ab82ad854c9fc9b14c8de1

pngread.c
ctruta: Done. See caa0c733f32661edf8ab82ad854c9fc9b14c8de1.

pngtrans.c
ctruta: I agree with your suggested change, but I haven't done it here because I'm reserving a post-refactoring clean-up step for that.

ctruta:
Hello, reviewers,

I have two big honking commits here today, one of which needs your scrutinizing attention: 50e58961480eeabb6e803261dfac5e3c6c93acd3. The other one is exclusively code formatting.

If you're wondering whether the previous replacements of `png_structp` to `png_struct *` disappeared: yes, they did disappear from the current iteration, but they will come back in the next commit, which is still work in progress, and which is tentatively titled -- spoiler alert! -- `refactor: Replace deprecated pointer typedefs (part 2)`.

pngtrans.c
ProgramMax: Oh, I didn't mean any action was required.
I just wanted to point out the interesting language design implications.

pngget.c
```diff
@@ -993,8 +992,8 @@ png_get_oFFs(png_const_structrp png_ptr, png_const_inforp info_ptr,
 #ifdef PNG_pCAL_SUPPORTED
 png_uint_32
 png_get_pCAL(png_const_structrp png_ptr, png_inforp info_ptr,
-    png_charp *purpose, png_int_32 *X0, png_int_32 *X1, int *type, int *nparams,
-    png_charp *units, png_charpp *params)
+    char **purpose, png_int_32 *X0, png_int_32 *X1, int *type, int *nparams,
+    char **units, char ***params)
```
ProgramMax, line 996: I think this is the first time I've seen a triple pointer in actual use. :D
No suggested changes. Just pointing out something neat.

pngtrans.c
ctruta: @ProgramMax but I agree that your observation is a good one.

I will leave in place code like this

```diff
-      png_const_bytep ppa, ppb;
+      const png_byte *ppa, *ppb;
```

and like this

```diff
-      png_bytep rp, dp, pp, lp;
+      png_byte *rp, *dp, *pp, *lp;
```

but otherwise, code fragments of the kind that you observed should be split, on the principle that variable declarations that do not share a common purpose should also not share a common declaration. And we're headed towards modern C and modern programming idioms, including declarations of variables at the point of their initialization.

ProgramMax (approved):
Phew. That was a lot.
LGTM.

I know you said there are follow-up commits. But I suggest we change the approach.
I just reviewed 2,600 lines across 56 files.
Some of it I've already reviewed. But GitHub doesn't make it easy to only review the changes from the latest commit. And arguably, it shouldn't--since I am approving a pull request, not a commit.

Ideally, commits AND reviews are small and isolated.
Your commits are. But the reviews aren't.

I think we might want to go ahead and land this (if it is currently functional). The follow-up commits can then go in a separate pull request.
Thoughts?

ctruta:
> Phew. That was a lot.
> LGTM.

Thank you, Chris!

> Ideally, commits AND reviews are small and isolated. Your commits are. But the reviews aren't.

Interesting observation. Ok.

> I think we might want to go ahead and land this (if it is currently functional). The follow-up commits can then go in a separate pull request. Thoughts?

Agreed.

ctruta:
@ProgramMax wrote:

> I think we might want to go ahead and land this (if it is currently functional). The follow-up commits can then go in a separate pull request. Thoughts?

@ctruta wrote:

> Agreed.

So I tried doing exactly that, but GitHub insisted on sending the updates to the same PR (i.e. this PR) because I kept pushing new changes from the same local branch on my development machine to the same branch in my repository clone at https://github.com/ctruta/libpng -- note to self: avoid doing that.

Then I force-pushed away the new commits in my own repo, which finally triggered the closing of this PR, although the closing now indicates "unmerged commits", which is incorrect (and it's not the first time GitHub does this to me).

But at least I'll be able to open a new PR and start a fresh round of reviews.

PR #744, opened 2025-10-31
api: Deprecate legacy pointer typedefs (part 2)

ctruta:
Following up on #743.

Heads-up: a change to `manuals/libpng-history.txt` with a description of what's new in libpng18 is coming up next.

### DIFF
--- a/png.h
+++ b/png.h
@@ -2826,8 +2826,8 @@ PNG_EXPORT(int, png_get_palette_max,
 
 #define PNG_IMAGE_VERSION 1
 
-typedef struct png_control png_control; /* [Opaque] */
-typedef png_control *png_controlp;      /* [Deprecated] */
+typedef struct png_control_def png_control; /* [Opaque] */
+typedef png_control *png_controlp;          /* [Deprecated] */
 
 typedef struct
 {
--- a/pngpriv.h
+++ b/pngpriv.h
@@ -1996,15 +1996,16 @@ PNG_INTERNAL_FUNCTION(int, png_xy_from_XYZ,
 /* SIMPLIFIED READ/WRITE SUPPORT */
 #if defined(PNG_SIMPLIFIED_READ_SUPPORTED) ||\
     defined(PNG_SIMPLIFIED_WRITE_SUPPORTED)
+
 /* The internal structure that png_image::opaque points to. */
-struct png_control
+struct png_control_def
 {
    png_struct *png_ptr;
    png_info   *info_ptr;
-   void       *error_buf;           /* Always a jmp_buf at present. */
+   void       *error_buf;           /* Currently this can only be jmp_buf */
 
-   const png_byte *memory;          /* Memory buffer. */
-   size_t          size;            /* Size of the memory buffer. */
+   const png_byte *memory;          /* Memory buffer */
+   size_t          size;            /* Size of the memory buffer */
 
    unsigned int for_write       :1; /* Otherwise it is a read structure */
    unsigned int owned_file      :1; /* We own the file in io_ptr */


pngrutil.c
```diff
@@ -3416,13 +3416,13 @@ png_handle_chunk(png_structrp png_ptr, png_inforp info_ptr, png_uint_32 length)
...
 void /* PRIVATE */
-png_combine_row(png_const_structrp png_ptr, png_byte *dp, int display)
+png_combine_row(const png_struct *png_ptr, png_byte *dp, int display)
 {
    unsigned int pixel_depth = png_ptr->transformed_pixel_depth;
    const png_byte *sp = png_ptr->row_buf + 1;
    png_alloc_size_t row_width = png_ptr->width;
    unsigned int pass = png_ptr->pass;
-   png_byte *end_ptr = 0;
+   png_byte *end_ptr = NULL;
```
ProgramMax, line 3425: Good catch on this change.

pngpriv.h
```diff
@@ -1996,19 +1996,20 @@ PNG_INTERNAL_FUNCTION(int, png_xy_from_XYZ,
 /* SIMPLIFIED READ/WRITE SUPPORT */
 #if defined(PNG_SIMPLIFIED_READ_SUPPORTED) ||\
     defined(PNG_SIMPLIFIED_WRITE_SUPPORTED)
+
 /* The internal structure that png_image::opaque points to. */
-typedef struct png_control
+struct png_control_def
```
ProgramMax, line 2001: The name here is changing from 'png_control' to 'png_control_def'.
(Also, the previous code typedefed 'png_control' to 'png_control'? You can do that?? I need to learn C more. That seems odd.)

Anyway, you chose the new name so I am just pointing it out to myself, I guess--to look for other changes that rely on this.

png.h
```diff
@@ -2826,11 +2826,12 @@ PNG_EXPORT(int, png_get_palette_max,
 
 #define PNG_IMAGE_VERSION 1
 
-typedef struct png_control *png_controlp; /* [Deprecated] */
+typedef struct png_control_def png_control; /* [Opaque] */
```
ProgramMax, line 2829: Ah, okay I see now. 'png_control_def' is used here to reestablish 'png_control'. Got it.

contrib/libtests/pngvalid.c
```diff
@@ -11538,7 +11538,7 @@ signal_handler(int signum)
          break;
    }
 
-   store_log(&pm.this, NULL/*png_structp*/, msg, 1/*error*/);
+   store_log(&pm.this, NULL/*png_struct**/, msg, 1/*error*/);
```
ProgramMax, line 11541: Should we add spaces to the comments?
`/* png_struct* */`    this is clear to me
`/*png_struct**/`    I am almost certain I would misread this.
What are your thoughts?

contrib/libtests/pngvalid.c
```diff
@@ -10046,11 +10046,11 @@ gamma_component_validate(const char *name, const validate_info *vi,
...
 
 static void
-gamma_image_validate(gamma_display *dp, png_const_structp pp,
-   png_infop pi)
+gamma_image_validate(gamma_display *dp, const png_struct *pp,
+   png_info *pi)
 {
    /* Get some constants derived from the input and output file formats: */
-   const png_store* const ps = dp->this.ps;
+   const png_store *ps = dp->this.ps;
```
ProgramMax, line 10053: Just noting that this changed. But it is fine.

contrib/libtests/pngvalid.c
```diff
@@ -6694,11 +6694,11 @@ transform_range_check(png_const_structp pp, unsigned int r, unsigned int g,
...
 
 static void
-transform_image_validate(transform_display *dp, png_const_structp pp,
-   png_infop pi)
+transform_image_validate(transform_display *dp, const png_struct *pp,
+   png_info *pi)
 {
    /* Constants for the loop below: */
-   const png_store* const ps = dp->this.ps;
+   const png_store *ps = dp->this.ps;
```
ProgramMax, line 6701: Same here. Noting a change which should be fine.

contrib/libtests/pngvalid.c
```diff
@@ -1951,13 +1951,13 @@ store_write_reset(png_store *ps)
    store_freenew(ps);
 }
 
-/* The following is the main write function, it returns a png_struct and,
- * optionally, a png_info suitable for writiing a new PNG file.  Use
+/* The following is the main write function, it returns a png_struct* and,
+ * optionally, a png_info* suitable for writing a new PNG file.  Use
```
ProgramMax, line 1955: Good catch

manuals/libpng.3
```diff
@@ -7,493 +7,493 @@ libpng \- Portable Network Graphics (PNG) Reference Library 1.8.0.git
...
+\fBvoid png_set_scale_16 (png_struct \fI*png_ptr\fP\fB);\fP
 
-\fBvoid png_set_shift (png_structp \fP\fIpng_ptr\fP\fB, png_color_8p \fItrue_bits\fP\fB);\fP
+\fBvoid png_set_shift (png_struct \fP\fI*png_ptr\fP\fB, png_color_8 \fI*true_bits\fP\fB);\fP
 
-\fBvoid png_set_sig_bytes (png_structp \fP\fIpng_ptr\fP\fB, int \fInum_bytes\fP\fB);\fP
+\fBvoid png_set_sig_bytes (png_struct \fP\fI*png_ptr\fP\fB, int \fInum_bytes\fP\fB);\fP
 
-\fBvoid png_set_sPLT (png_structp \fP\fIpng_ptr\fP\fB, png_infop \fP\fIinfo_ptr\fP\fB, png_spalette_p \fP\fIsplt_ptr\fP\fB, int \fInum_spalettes\fP\fB);\fP
+\fBvoid png_set_sPLT (png_struct \fP\fI*png_ptr\fP\fB, png_info \fP\fI*info_ptr\fP\fB, png_splt_t \fP\fI*splt_ptr\fP\fB, int \fInum_spalettes\fP\fB);\fP
```
ProgramMax, line 420: 'png_spalette_p' became 'png_splt_t'?
I didn't see that in this change.
Did that happen elsewhere and we missed it in the docs?

manuals/libpng.3
```diff
@@ -7,493 +7,493 @@ libpng \- Portable Network Graphics (PNG) Reference Library 1.8.0.git
...
+\fBvoid png_get_sCAL_fixed (const png_struct \fP\fI*png_ptr\fP\fB, const png_info \fP\fI*info_ptr\fP\fB, int* \fP\fIunit\fP\fB, png_fixed_point \fP\fI*width\fP\fB, png_fixed_point \fI*height\fP\fB);\fP
 
-\fBvoid png_get_sCAL_s (png_const_structp \fP\fIpng_ptr\fP\fB, png_const_infop \fP\fIinfo_ptr\fP\fB, int* \fP\fIunit\fP\fB, char \fP\fI**width\fP\fB, char \fI**height\fP\fB);\fP
+\fBvoid png_get_sCAL_s (const png_struct \fP\fI*png_ptr\fP\fB, const png_info \fP\fI*info_ptr\fP\fB, int* \fP\fIunit\fP\fB, char \fP\fI**width\fP\fB, char \fI**height\fP\fB);\fP
 
-\fBpng_byte *png_get_signature (png_const_structp \fP\fIpng_ptr\fP\fB, png_infop \fIinfo_ptr\fP\fB);\fP
+\fBpng_byte *png_get_signature (const png_struct \fP\fI*png_ptr\fP\fB, png_info \fI*info_ptr\fP\fB);\fP
 
-\fBpng_uint_32 png_get_sPLT (png_const_structp \fP\fIpng_ptr\fP\fB, png_const_infop \fP\fIinfo_ptr\fP\fB, png_spalette_p \fI*splt_ptr\fP\fB);\fP
+\fBpng_uint_32 png_get_sPLT (const png_struct \fP\fI*png_ptr\fP\fB, const png_info \fP\fI*info_ptr\fP\fB, png_splt_t \fI**splt_ptr\fP\fB);\fP
```
ProgramMax, line 162: This is also a change from 'png_spalette_p' to 'png_splt_t'.

ProgramMax (approved):
LGTM. Some comments / questions. But I'm okay if it results in no change.

pngpriv.h
ctruta: > (Also, the previous code typedefed 'png_control' to 'png_control'? You can do that?? I need to learn C more. That seems odd.)

Walter Bright, the creator of the D programming language, had "something to say" 🤔 about the C language requiring the names of aggregates (i.e. structs and unions) to be stored in a separate namespace. If you want a header file to work both with C and C++ source files, you need to guard declarations like this:
```c++
#ifndef __cplusplus
typedef struct foo foo;
#endif
```

contrib/libtests/pngvalid.c
ctruta: How about this:
```diff
-   store_log(&pm.this, NULL/*png_structp*/, msg, 1/*error*/);
+   store_log(&pm.this, NULL, msg, 1);
```

contrib/libtests/pngvalid.c
ctruta: The top-level `const` is redundant. I want to make a separate commit that cleans them all up. There aren't many of them, and they're used inconsistently and non-idiomatically.

contrib/libtests/pngvalid.c
ctruta: Yeah, most of the changes in these commit have been done by the means of scripting, but some of those are hand-made. In particular, I changed these instances of `png_store*` into `png_store *` because -- again -- the old form is idiomatic in C++ but not so in C. And while I made those changes, I also removed the top-level (redundant) `const` qualifier.

manuals/libpng.3
ctruta: The old text `png_spalette_p` was a typo. See the actual code in `png.h`, `pngget.h` and `pngset.h`.

manuals/libpng.3
ctruta: The old text `png_spalette_p` was a typo. See the actual code in `png.h`, `pngget.h` and `pngset.h`.

contrib/libtests/pngvalid.c
ProgramMax: Your call.
I think I prefer the comments because the NULL and 1 are magic numbers here. The comment helps fight that.

ctruta: > Your call. I think I prefer the comments because the NULL and 1 are magic numbers here. The comment helps fight that.

Well, `NULL` is the null context, while 0 and 1 should be replaced with
`enum { LOG_WARN = 0, LOG_ERR = 1 }`.

You are correct to point out the use of magic numbers, but to me that magic feels like a code smell. If the code is not self-documenting, it should be made self-documenting.

ProgramMax: Agreed that the enum would be better than the comment. By far.

PR #745, opened 2025-11-02
Require zlib 1.2.8 -- shed maintenance

hmaarrfk:
There are some proposals to move to a new zlib backend.

My first step in writing a patch for this was to shed some of the compatibility if statements.

Zlib 1.2.6 was release in 2012 according to the archive history https://zlib.net/fossils/

I think that 13years is pretty generous and should help shed some mental load

xref: https://github.com/pnggroup/libpng/issues/205

cc: @FabioPedretti

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -174,6 +174,19 @@ AC_CHECK_LIB([z], [zlibVersion], ,
   [AC_CHECK_LIB([z], [${ZPREFIX}zlibVersion], ,
      [AC_MSG_ERROR([zlib not installed])])])
 
+AC_MSG_CHECKING([zlib version])
+AC_COMPILE_IFELSE(
+  [AC_LANG_SOURCE([
+#include <zlib.h>
+#if !defined(ZLIB_VERNUM) || ZLIB_VERNUM < 0x1280
+#error zlib version 1.2.8 or later is required
+#endif
+int main(void) { return 0; }
+])],
+  [AC_MSG_RESULT([acceptable])],
+  [AC_MSG_RESULT([too old or ZLIB_VERNUM not defined])
+   AC_MSG_ERROR([zlib version 1.2.8 or later is required])])
+
 # The following is for pngvalid, to ensure it catches FP errors even on
 # platforms that don't enable FP exceptions, the function appears in the math
 # library (typically), it's not an error if it is not found.
--- a/pngrutil.c
+++ b/pngrutil.c
@@ -439,14 +439,9 @@ png_inflate_claim(png_structrp png_ptr, png_uint_32 owner)
     * follow that because, for systems with with limited capabilities, we
     * would otherwise reject the application's attempts to use a smaller window
     * size (zlib doesn't have an interface to say "this or lower"!).
-    *
-    * inflateReset2 was added to zlib 1.2.4; before this the window could not be
-    * reset, therefore it is necessary to always allocate the maximum window
-    * size with earlier zlibs just in case later compressed chunks need it.
     */
    {
       int ret; /* zlib return code */
-#if ZLIB_VERNUM >= 0x1240
       int window_bits = 0;
 
       if (((png_ptr->options >> PNG_MAXIMUM_INFLATE_WINDOW) & 3) ==
@@ -460,7 +455,6 @@ png_inflate_claim(png_structrp png_ptr, png_uint_32 owner)
       {
          png_ptr->zstream_start = 1;
       }
-#endif /* ZLIB_VERNUM >= 0x1240 */
 
       /* Set this for safety, just in case the previous owner left pointers to
        * memory allocations.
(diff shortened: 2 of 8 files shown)


ctruta (changes requested):
Thank you very much for your contribution, @hmaarrfk.

I agree with your change in principle, and I, too, want to raise the minimum required zlib version to 1.2.8 (not just 1.2.6), but this change needs to be complete. If you look in the libpng codebase, you will see preprocessor directives comparing `ZLIB_VERNUM` with 0x1240 and 0x1260. (There are also comparisons with 0x1290 but we can ignore those for now.)

So, this change can go in cleanly after the preprocessor checks for `ZLIB_VERNUM` numbers below 0x1280 are all cleaned up.

Finally, if you can please add your name the AUTHORS file, that's be great!

ctruta:
> Thank you very much for your contribution, @hmaarrfk.

Following up on my own review.

Actually... I missed it... so you DID make all those changes -- but can you please squash them all into a single commit, considering that we raise the minimum required Zlib version to 1.2.8 in one fell swoop?

configure.ac
```diff
@@ -174,6 +174,19 @@ AC_CHECK_LIB([z], [zlibVersion], ,
   [AC_CHECK_LIB([z], [${ZPREFIX}zlibVersion], ,
      [AC_MSG_ERROR([zlib not installed])])])
 
+AC_MSG_CHECKING([zlib version])
+AC_COMPILE_IFELSE(
+  [AC_LANG_SOURCE([
+#include <zlib.h>
+#if !defined(ZLIB_VERNUM) || ZLIB_VERNUM < 0x1260
+#error zlib version 1.2.6 or later is required
```
ctruta: See my previous comment about raising the minimum required zlib version to 1.2.8 or newer.

pngrutil.c
```diff
@@ -440,13 +440,11 @@ png_inflate_claim(png_structrp png_ptr, png_uint_32 owner)
     * would otherwise reject the application's attempts to use a smaller window
     * size (zlib doesn't have an interface to say "this or lower"!).
     *
-    * inflateReset2 was added to zlib 1.2.4; before this the window could not be
-    * reset, therefore it is necessary to always allocate the maximum window
-    * size with earlier zlibs just in case later compressed chunks need it.
+    * inflateReset2 was added to zlib 1.2.4.  libpng now requires zlib 1.2.6
```
ctruta: This comment about `inflateReset2` is no longer applicable. It needs not be rephrased and should be deleted altogether.

pngstruct.h
```diff
@@ -30,16 +30,8 @@
...
- * with older builds.
- */
-#if ZLIB_VERNUM < 0x1260
-#  define PNGZ_MSG_CAST(s) png_constcast(char*,s)
-#  define PNGZ_INPUT_CAST(b) png_constcast(png_byte *,b)
-#else
-#  define PNGZ_MSG_CAST(s) (s)
-#  define PNGZ_INPUT_CAST(b) (b)
-#endif
+#define PNGZ_MSG_CAST(s) (s)
```
ctruta: `PNGZ_MSG_CAST` and `PNGZ_INPUT_CAST` are internal macros, not part of the API, so they can (and should) be discarded as well.

ctruta (changes requested):

pngstruct.h
hmaarrfk: right, i think i just removed them out right.

Commenting just to say resolved.

will squash all as 1 commit as requested.

pngrutil.c
hmaarrfk: yeah, deleting it makes most sense. thanks for flagging that.

configure.ac
hmaarrfk: done.

hmaarrfk:
Are you OK with the extra commit i snuck in
https://github.com/pnggroup/libpng/pull/745/commits/b15add048b18213d6e547aa70c242b77250de864

i can squash it too if you are OK with these changes that remove an internal macro.

ctruta:
I approve this change and I want to integrate it.

@jbowler, @ProgramMax, @svgeesus, I also want to take a moment to contemplate this being the first time ever when we impose a minimum required zlib version, and to kindly ask you for your reviewing opinion before we proceed.

Interesting fun fact: to the best of my knowledge, zlib came into existence as a supporting library for PNG and libpng!

svgeesus:
zlib is certainly mature and well established enough that requiring a minimum version seems reasonable.

Are we aware of any platform where modern zlib versions are unavailable?

ctruta:
> Are we aware of any platform where modern zlib versions are unavailable?

It's quite possible that systems released in the early-to-mid 2010s (or even earlier) are still in use with the version of zlib they originally shipped. In those cases, the system libraries may indeed be quite old. That said, if such platforms allow building and installing newer libraries like libpng‑1.8.x from source, they should allow building and installing a recent zlib release from source as well.

svgeesus:
+1 to merge, therefore

PR #747, opened 2025-11-03
api: Deprecate legacy pointer typedefs (part 3)

ctruta:
Following up on #743 and #744 with an important fix and a clean-up job.

### DIFF
--- a/contrib/tools/pngfix.c
+++ b/contrib/tools/pngfix.c
@@ -1562,7 +1562,7 @@ chunk_end(struct chunk **chunk_var)
 }
 
 static void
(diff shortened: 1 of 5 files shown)


contrib/tools/pngfix.c
```diff
@@ -1562,7 +1562,7 @@ chunk_end(struct chunk **chunk_var)
 }
 
 static void
-chunk_init(struct chunk * const chunk, struct file * const file)
```
ctruta, line 1565: @jbowler, I can understand how and why this declaration works, and I haven't changed it because it was not within the scope of this commit, but I still consider it an anti-pattern. Have you actually intended to write it like this, or was it an oversight?

Issue #691, opened 2025-05-26
Buffer Overflow in pngrtran.c -> png_do_quantize()


Samsung-PENTEST:
We did fuzzing tests on Tizen 7.0 and found this bug. (libpng-1.6.37)

# Issue

Buffer overflow in [pngrtran.c](https://github.com/pnggroup/libpng/blob/v1.6.37/pngrtran.c) -> `png_do_quantize()`

Fuzzer report:
```bash
==103662==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x50200000010f at pc 0x557021306bc0 bp 0x7ffc6269cc40 sp 0x7ffc6269cc38                                                                                                                                           [26/1871]
READ of size 1 at 0x50200000010f thread T0                       
    #0 0x557021306bbf in png_do_quantize /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngrtran.c:4729:19
    #1 0x5570212aee21 in png_do_read_transformations /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngrtran.c:4934:7
    #2 0x557021265b87 in png_read_row /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngread.c:652:7
    #3 0x55702126a09c in png_read_image /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngread.c:833:10
    #4 0x557021208444 in LLVMFuzzerTestOneInput /home/abuild/rpmbuild/BUILD/libpng-1.6.37/fuzzing_harnesses/read/harness_read.cc:343:5
    #5 0x5570210c1366 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /home/abuild/llvm-project-llvmorg-19.1.2/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:614:15
    #6 0x5570210a5310 in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) /home/abuild/llvm-project-llvmorg-19.1.2/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:328:21
    #7 0x5570210b144a in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /home/abuild/llvm-project-llvmorg-19.1.2/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:863:19
[...]

Issue #690, opened 2025-05-26
Memory Leak in pngrtran.c -> png_set_quantize()


Samsung-PENTEST:
We did fuzzing tests on Tizen 7.0 and found this bug. (libpng-1.6.37)

# Issue

Memory leak in [pngrtran.c](https://github.com/pnggroup/libpng/blob/v1.6.37/pngrtran.c) -> `png_set_quantize()`

Fuzzer report:
```bash
==391251==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 256 byte(s) in 1 object(s) allocated from:
    #0 0x555555822ddf in malloc /home/abuild/llvm-project-llvmorg-19.1.2/compiler-rt/lib/asan/asan_malloc_linux.cpp:68:3
    #1 0x5555558b7e56 in png_malloc_base /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngmem.c:95:17
    #2 0x5555558b7a0e in png_malloc /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngmem.c:179:10
    #3 0x5555558ccd02 in png_set_quantize /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngrtran.c:443:44
    #4 0x555555862626 in LLVMFuzzerTestOneInput /home/abuild/rpmbuild/BUILD/libpng-1.6.37/fuzzing_harnesses/read/harness_read.cc:264:9
    #5 0x55555571c1e6 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /home/abuild/llvm-project-llvmorg-19.1.2/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:614:15
    #6 0x555555723de0 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) /home/abuild/llvm-project-llvmorg-19.1.2/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:516:22
    #7 0x555555724cd0 in fuzzer::Fuzzer::MutateAndTestOne() /home/abuild/llvm-project-llvmorg-19.1.2/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:760:25
[...]

PR #748, opened 2025-11-07
Fix memory leaks and buffer overflows in pngrtran.c

ctruta:
Free previously allocated `quantize_index` and `quantize_sort` arrays before reallocating them. This prevents memory leaks when `png_set_quantize()` is called multiple times on the same `png_struct`.

The function assumed single-call usage, but fuzzing revealed that repeated calls would overwrite the pointers without freeing the original allocations, leaking 256 bytes per call for `quantize_index` and additional memory for `quantize_sort` when histogram-based quantization is used.

Analyzed-by: degrigis <degrigis@users.noreply.github.com>

### DIFF
--- a/pngrtran.c
+++ b/pngrtran.c
@@ -501,6 +501,12 @@ png_set_quantize(png_structrp png_ptr, png_colorp palette,
    {
       int i;
 
+      /* Initialize the array to index colors.
+       *
+       * Be careful to avoid leaking memory. Applications are allowed to call
+       * this function more than once per png_struct.
+       */
+      png_free(png_ptr, png_ptr->quantize_index);
       png_ptr->quantize_index = (png_bytep)png_malloc(png_ptr,
           (png_alloc_size_t)num_palette);
       for (i = 0; i < num_palette; i++)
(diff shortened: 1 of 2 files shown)


ctruta:
This is a proposed fix for issue #690.
@jbowler @degrigis please test and review.

jbowler:
```
if (png_ptr->quantize_index != NULL)
         png_free(png_ptr, png_ptr->quantize_index);
```

Replace the second line by "png_app_error(png_ptr, "png_set_quantize may only be called once.");

That's what Glenn assumed (among many other things) when he wrote this code, it's whack-a-mole trying to fix it so when it is used in a clearly idiotic way just error out.

I do suggest that you also deprecate the functionality and delete it in 1.8.  Legacy corporate bumpkins quite likely use it but anyone with a genuine need for a palette based (very lossy) transform of a PNG can just use the simplified API and get something more than adequate for their dolphins.

ctruta:
Thanks for the review, @jbowler. I've reworked the commit to add proper refactoring. The memory leak fix itself (without the refactoring) is now a one-liner. See commit 0c2e440b081dd6dc66a4672af44fd4d5f55943aa.

Regarding the broader question on whether to error out on multiple calls: the defensive approach is the correct one to use here. The simplified API is good for simplified use cases, while the not-so-simplified API is good for not-so-simplified use cases. The `png_set_xx` functions are meant to be called multiple times, unless stated otherwise, and valid use cases do exist. Restricting this for `png_set_quantize`, arbitrarily, would be inconsistent with the established patterns.

ctruta:
I renamed this PR and I closed #750 as a duplicate of this. We are now fixing #690 (a confirmed memory leak in `png_set_quantize`) and #691 (a confirmed buffer overflow in `png_do_quantize`) and also a newly-discovered (unconfirmed but possible) buffer overflow in `png_do_expand_palette`.

ctruta:
Reviewing my own commit ea058b5db5655c2fe401486fa0e0965bd3ab6ff8 and giving myself a thumbs-down 👎 not to mention laughing at myself in the process 😝

The function `png_do_expand_palette` is NOT vulnerable in the way `png_do_quantize` is vulnerable. That was a speculative fix for a non-existent problem, and that's because both `png_ptr->palette` and `png_ptr->trans_alpha` are guaranteed to have 256 elements each in the implementation, and that's regardless of how many entries are reported by `png_ptr->num_palette` and in `png_ptr->num_trans`.

So I force-pushed the commit out of this PR, leaving in only the fix that is actually fixing stuff.

ctruta:
In the new commit d6fad01964a5c65ed4daae09a33cd11b6f2d7f55 you will see the solution that has been right here, all along, waiting to be discovered.
The hot path is back to being hot and clean.

Issue #746, opened 2025-11-03
When using the simplified write API png_image_write_to_memory() to encode an RGB image, libpng performs an out-of-bounds read inside png_write_image_8bit.


weijinjinnihao:
# Bug Report

## Describe the bug  
When using the simplified write API png_image_write_to_memory() to encode an RGB image, libpng performs an out-of-bounds read inside png_write_image_8bit. The read goes exactly one pixel beyond the end of the caller-provided input pixel buffer.(RGBA can also trigger it)

Minimal repro (ASan enabled) shows:
- W=18, H=23, format=PNG_FORMAT_RGB
- row_stride=54 (computed by PNG_IMAGE_ROW_STRIDE(img), i.e., width * 3)
- need = row_stride * height = 1242

The crash occurs inside:
```
png_write_image_8bit (pngwrite.c:~1803–1844)
→ png_image_write_main
→ png_image_write_to_memory
```
Expected: libpng must only read the need bytes of the caller-provided buffer (or at most up to row_stride per row), never past buffer + need.
Actual: libpng reads 2 bytes at buffer + need (out-of-bounds).

# Root Cause Analysis
Inside the simplified-API write path:
```
png_image_write_to_memory
 → png_image_write_main
   → png_safe_execute(image, png_write_image_8bit, display)
     → png_write_image_8bit  // crash site
```

png_write_image_8bit processes each row of an 8-bit RGB source. With width=18 and row_stride=54 (exactly width*3), the function performs a word-sized / 16-bit access or look-ahead near the end of a row, causing it to read two bytes past the valid row tail when handling the last pixel. This is consistent with:
- ASan reporting “READ of size 2” at buffer + need.
- The problem reproducing even when the caller passes an exact stride (no padding) and a buffer sized to row_stride * height.
In short, the tail handling in png_write_image_8bit’s “8-bit RGB” fast path has an off-by-one / unguarded 16-bit load that assumes either extra padding or allows a read past row_stride at the row end. The loop termination or the final element copy path does not guard against a 2-byte fetch when only 1 byte remains.

# Suggested fix
In png_write_image_8bit (8-bit RGB path):
1. Ensure any word/half-word loads (e.g., png_uint_16 reads, unrolled copies, or macro-based pair accesses) are preceded by a remaining-bytes check against the row tail.
2. For the final 1–2 bytes of a row, fall back to byte-wise copies.
[...]

Issue #688, opened 2025-05-26
Buffer Overflow in pngwrite.c -> png_write_image_8bit() with alpha channel


Samsung-PENTEST:
We did fuzzing tests on Tizen 7.0 and found this bug. (libpng-1.6.37)

# Issue

2 bytes read beyond allocated buffer.
Location: [pngwrite.c](https://github.com/pnggroup/libpng/blob/v1.6.37/pngwrite.c) -> `png_write_image_8bit()`

Fuzzer report:
```bash
==678216==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x51d00000d082 at pc 0x555555aaea15 bp 0x7fffffffbab0 sp 0x7fffffffbaa8
READ of size 2 at 0x51d00000d082 thread T0
    #0 0x555555aaea14 in png_write_image_8bit /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngwrite.c:1802:33
    #1 0x55555590c8da in png_safe_execute /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngerror.c:951:16
    #2 0x555555aa1080 in png_image_write_main /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngwrite.c:2209:19
    #3 0x555555a9cd14 in png_image_write_memory /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngwrite.c:2284:11
    #4 0x55555590c8da in png_safe_execute /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngerror.c:951:16
    #5 0x555555a9bdef in png_image_write_to_memory /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngwrite.c:2318:22
    #6 0x5555558c0585 in LLVMFuzzerTestOneInput /home/abuild/rpmbuild/BUILD/libpng-1.6.37/fuzzing_harnesses/png/harness_png.cc:113:9
    #7 0x55555577b256 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /home/abuild/llvm-project-llvmorg-19.1.2/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:614:15
[...]

PR #749, opened 2025-11-07
Fix heap buffer overflow in `png_write_image_8bit`

ctruta:
The condition guarding the pre-transform path incorrectly allowed 8-bit input data to enter `png_write_image_8bit()` which expects 16-bit input. This caused out-of-bounds reads when processing 8-bit grayscale+alpha images with `convert_to_8bit = 1`.

The second part of the condition, i.e.
```colormap == 0 && convert_to_8bit != 0```
failed to verify that input was 16-bit, i.e.
```linear != 0```
contradicting the comment
```/* [...] This only applies when the input is 16-bit [...] */```

The fix consists in restructuring the condition to ensure both the `alpha` path and the `convert_to_8bit` path require linear (16-bit) input:

```if (linear != 0 && (alpha != 0 || display->convert_to_8bit != 0))```

This makes the code match its documentation and prevents treating 8-bit buffers as 16-bit data.

Analyzed-by: degrigis <degrigis@users.noreply.github.com>

### DIFF
--- a/pngwrite.c
+++ b/pngwrite.c
@@ -2173,8 +2173,7 @@ png_image_write_main(png_voidp argument)
     * before it is written.  This only applies when the input is 16-bit and
     * either there is an alpha channel or it is converted to 8-bit.
     */
-   if ((linear != 0 && alpha != 0 ) ||
-       (colormap == 0 && display->convert_to_8bit != 0))
+   if (linear != 0 && (alpha != 0 || display->convert_to_8bit != 0))
    {
       png_bytep row = png_voidcast(png_bytep, png_malloc(png_ptr,
           png_get_rowbytes(png_ptr, info_ptr)));


ctruta:
This is a proposed fix for issue #688.
@jbowler @degrigis please test and review.

jbowler:
Interesting.  Bear in mind that "convert_to_8bit" means that the input is 16-bit even though the output is 8-bit sRGB.

Please confirm that you have verified that all the pngstest cases still pass and, if you have, include the test image I eventually found in #746.  I also have a png_error add which identifies the #746 bug more reliably (i.e. all the cases it occurs).

jbowler:
Also this part of the CI message is ambiguous:

>This caused out-of-bounds reads when processing 8-bit grayscale+alpha
images with `convert_to_8bit = 1`.

You do mean images with an **output** format of 8-bit, not an input format, right?  convert_to_8_bit applies to the input.

Perhaps the API documentation is confusing?

ctruta:
@jbowler Thanks for reviewing. Here's a motto for this fix:

*"`convert_to_8bit = 1` causing heap overflow on 8-bit input? Classic off-by-assumption error..."* 🤨

---

I agree that the original commit message needed correction. Let us recap:

If the input is 8-bit grayscale+alpha (`linear = 0`, `alpha = 1`)
AND
if the parameter is `convert_to_8bit = 1`
THEN
we have an invalid combination: `convert_to_8bit = 1` should only be used with 16-bit linear input.

The old condition `(colormap == 0 && convert_to_8bit != 0)` incorrectly allowed this invalid combination to enter `png_write_image_8bit()`, which then treated the 8-bit buffer as 16-bit data, causing the 2-byte overflow.

In conclusion, the commit message had to be updated to clarify this as a matter of 8-bit input images being incorrectly processed when the `convert_to_8bit` flag is on. See the amended commit.

---

About testing: the full test harness is automatically run on CI services, and I'm running it locally as well, but I agree that `pngstest` needs an update.

As for issue #746 vs. #688: yes, they're in the same function, but as far as I understand, they're different defects with different root causes. Issue #688 affects grayscale+alpha with a conditional logic error. Issue #746 affects RGB images with unguarded 16-bit reads at row boundaries.

ctruta:
> As for issue #746 vs. #688: yes, they're in the same function, but as far as I understand, they're different defects with different root causes. Issue #688 affects grayscale+alpha with a conditional logic error. Issue #746 affects RGB images with unguarded 16-bit reads at row boundaries.

So @jbowler was correct. After a closer inspection, I realized the root cause of this defect is the same, even though the triggers came from different paths (i.e. grayscale+alpha vs. RGB/RGBA).

I pushed another amended commit message in which I'm giving credit to everybody who contributed to the discovery and the analysis. The credit list is in the chronological order of the published contributions. I hope I got everything right -- but let me know if I didn't 😊

PR #750, opened 2025-11-07
Fix buffer overflow in `png_do_quantize`

ctruta:
Preliminary fix for issue #691


ctruta:
I'm closing this PR as a duplicate of #748. There is already a list of fixes to be applied to `pngrtran.c`, related to one another, to be tested and reviewed in one go. The fix originally proposed here is one of those.

Issue #686, opened 2025-05-26
Index out of bounds in libpng-1.6.37/pngread.c -> png_image_read_composite()


Samsung-PENTEST:
We did fuzzing tests on Tizen 7.0 and found this bug. (libpng-1.6.37)

# Issue

Index out of bounds ended up with reading beyond allocated buffer.

Location: [pngread.c](https://github.com/pnggroup/libpng/blob/v1.6.37/pngread.c) -> `png_image_read_composite()`

Fuzzer report:
```bash
==357343==ERROR: AddressSanitizer: global-buffer-overflow on address 0x00456b00 at pc 0x005fcef0 bp 0x407ff57c sp 0x407ff574
READ of size 2 at 0x00456b00 thread T0
    #0 0x005fceec in png_image_read_composite /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngread.c:3566:37
    #1 0x005e314c in png_safe_execute /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngerror.c:951:16
    #2 0x005f65a8 in png_image_read_direct /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngread.c:4271:16
    #3 0x005e314c in png_safe_execute /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngerror.c:951:16
    #4 0x005ecf7c in png_image_finish_read /home/abuild/rpmbuild/BUILD/libpng-1.6.37/pngread.c:4394:25
    #5 0x005c05ac in LLVMFuzzerTestOneInput /home/abuild/rpmbuild/BUILD/libpng-1.6.37/fuzzing_harnesses/png/harness_png.cc:105:14
    #6 0x0048e1c8 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned int) /home/abuild/llvm-project-llvmorg-19.1.7/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:614:15
    #7 0x00494a68 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned int, bool, fuzzer::InputInfo*, bool, bool*) /home/abuild/llvm-project-llvmorg-19.1.7/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:516:22
[...]

PR #751, opened 2025-11-12
Fix a buffer overflow in `png_init_read_transformations`

ctruta:
The palette compositing code in `png_init_read_transformations` was incorrectly applying background compositing when PNG_FLAG_OPTIMIZE_ALPHA was set. This violated the premultiplied alpha invariant `component <= alpha` expected by `png_image_read_composite`, causing values that exceeded the valid range for the PNG_sRGB_FROM_LINEAR lookup tables.

When PNG_ALPHA_OPTIMIZED is active, palette entries should contain pure premultiplied RGB values without background compositing. The background compositing must happen later in `png_image_read_composite` where the actual background color from the PNG file is available.

The fix consists in introducing conditional behavior based on PNG_FLAG_OPTIMIZE_ALPHA: when set, the code performs only premultiplication using the formula `component * alpha + 127) / 255` with proper gamma correction. When not set, the original background compositing calculation based on the `png_composite` macro is preserved.

This prevents buffer overflows in `png_image_read_composite` where out-of-range premultiplied values would cause out-of-bounds array access in `png_sRGB_base[]` and `png_sRGB_delta[]`.

Analyzed-by: John Bowler <jbowler@acm.org>

### DIFF
--- a/pngrtran.c
+++ b/pngrtran.c
@@ -1781,19 +1781,51 @@ png_init_read_transformations(png_structrp png_ptr)
                   }
                   else /* if (png_ptr->trans_alpha[i] != 0xff) */
                   {
-                     png_byte v, w;
+                     if ((png_ptr->flags & PNG_FLAG_OPTIMIZE_ALPHA) != 0)
+                     {
+                        /* Premultiply only:
+                         * component = round((component * alpha) / 255)
+                         */
+                        png_uint_32 component;
 
-                     v = png_ptr->gamma_to_1[palette[i].red];
-                     png_composite(w, v, png_ptr->trans_alpha[i], back_1.red);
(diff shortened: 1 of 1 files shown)


ctruta:
@jbowler @degrigis @Samsung-PENTEST please review my proposed fix to the index-out-of-bounds issue #686.

Venjilin:
Hello. I try to make a test to reproduce this problem (CVE-2025-64720). I can not create the test. I can reproduce problem from this: https://github.com/pnggroup/libpng/security/advisories/GHSA-9mpm-9pxh-mg4f.  But this is the other problem. Is there the test, that was been used to check this fix?  I can`t understand how background  affects on the problem if when PNG_FLAG_OPTIMIZE_ALPHA is set, background is 0 (png_set_alpha_mode_fixed()).

Issue #755, opened 2025-11-13
Vulnerability Report: Heap Buffer Overflow in png_combine_row (libpng16, triggered via png_image_finish_read)


yosiimich:
### Summary

A **heap buffer overflow** occurs in libpng during decoding of malformed PNG input under the `png_image_finish_read()` API path.

The issue arises even when processed through libpng’s internal safety wrapper `png_safe_execute()`, indicating a failure of libpng’s own protection mechanisms.

The overflow is caused by **insufficient validation of internal row size calculations** within `png_combine_row()`, allowing out-of-bounds writes when the PNG header specifies dimensions that exceed the user-allocated output buffer.

### Details

The overflow originates from **an internal inconsistency between libpng's trusted metadata and the caller-provided output buffer**.

libpng computes the number of bytes required for each row using:

```c
PNG_ROWBYTES(pixel_depth, row_width)

```

where:

- `row_width` is taken directly from `png_ptr->width` (IHDR header)
- the value may be attacker-controlled through a malformed PNG
- no validation is performed to ensure this width matches the actual, safe output dimensions tracked in the high-level `png_image` API (`image->width`, `image->height`)

As a result, if `png_ptr->width` becomes excessively large—due to crafted IHDR or transformation state—libpng will attempt to copy a row buffer larger than the memory allocated by the caller.

The overflow occurs in:

```c
memcpy(dp, sp, PNG_ROWBYTES(pixel_depth, row_width));

```

because:

- `dp` (destination) comes from user-controlled allocation via `png_image_finish_read()`
- the size is calculated solely from internal metadata without boundary checks
- libpng assumes `dp` is large enough but never verifies it

This discrepancy between **internal expected row size** and **actual allocated buffer size** is the primary cause.

`png_safe_execute()` is designed to wrap libpng operations inside a `setjmp/longjmp` error-handling framework.

However, this safety mechanism only protects against internal errors reported via:

```c
png_error(png_ptr, ...)

```

or related functions that explicitly trigger `longjmp`.

In this vulnerability:

- the out-of-bounds write occurs **inside `memcpy()`**
- the memory violation happens **before** any libpng code detects an error
- there is **no precondition check** that would invoke `png_error()`
- therefore no `longjmp` is triggered

Because the overflow happens *outside* the error-reporting path, the safety wrapper cannot intervene.

Thus the crash bypasses `png_safe_execute()` entirely, demonstrating that the safety layer does not cover unchecked memory operations within internal row-processing routines.

[...]

PR #757, opened 2025-11-17
Fix a heap buffer overflow in `png_image_finish_read`

ctruta:
Reject bit-depth mismatches between IHDR and the requested output format. When a 16-bit PNG is processed with an 8-bit output format request, `png_combine_row` writes using the IHDR depth before transformation, causing writes beyond the buffer allocated via `PNG_IMAGE_SIZE(image)`.

The validation establishes a safe API contract where `PNG_IMAGE_SIZE(image)` is guaranteed to be sufficient across the transformation pipeline.

Example overflow (32×32 pixels, 16-bit RGB to 8-bit RGBA):
- Input format: 16 bits/channel × 3 channels = 6144 bytes
- Output buffer: 8 bits/channel × 4 channels = 4096 bytes
- Overflow: 6144 bytes - 4096 bytes = 2048 bytes

Larger images produce proportionally larger overflows. For example, for 256×256 pixels, the overflow is 131072 bytes.

Reported-by: yosiimich <yosiimich@users.noreply.github.com>

### DIFF
--- a/pngread.c
+++ b/pngread.c
@@ -4040,6 +4040,20 @@ png_image_finish_read(png_imagep image, png_const_colorp background,
                   int result;
                   png_image_read_control display;
 
+                  /* Reject bit depth mismatches to avoid buffer overflows. */
+                  png_uint_32 ihdr_bit_depth =
+                      image->opaque->png_ptr->bit_depth;
+                  int requested_linear =
+                      (image->format & PNG_FORMAT_FLAG_LINEAR) != 0;
+                  if (ihdr_bit_depth == 16 && !requested_linear)
+                     return png_image_error(image,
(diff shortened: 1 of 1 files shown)


ctruta:
@jbowler I submitted this fix for Issue #755. I don't know how I missed this but it fails the tests.

I'm not asking for an actual review -- this is clearly a fail. I'm asking if you have a good idea on how to fix it.

From what I'm seeing right now, this is not a security bug, but a design error. Seriously, I'm still thinking about how to fix this (if it is even possible) without tearing down transformations in the simplified API.

We *might* make this work for image transformations that either shrink or leave intact the size of the row buffers. Never for those transformations that expand the row buffers.

PR #759, opened 2025-11-20
Fix hIST chunk position to match W3C specification

aYukiSekiguchi:
The hIST chunk was incorrectly positioned in the code. According to the
W3C specification, the hIST chunk must appear before the first IDAT
chunk and after PLTE (if present). This commit corrects the chunk
positioning to comply with the specification.


aYukiSekiguchi:
This will fix https://github.com/pnggroup/libpng/issues/758

I'm not sure I should commit the images from PngSuite. If I shouldn't, I will remove the images from this PR.

svgeesus (approved):

svgeesus:
> I'm not sure I should commit the images from PngSuite. If I shouldn't, I will remove the images from this PR.

No problem including them; the license allows reuse and modification.

jbowler:
Adding test images to pngsuite will cause failures unless the images are also added to the test suite itself as "expected failures".

I.e. the test suite has to know if the image it is testing is _expected_ to fail or if it is _expected_ to pass.

I.e. unless I'm being really stupid here (in which case please tell me) I cannot see how this commit fixes anything, well apart from the apparent identical issue in libpng16.

ctruta:
@aYukiSekiguchi a belated thank you for your contribution, with a note that I cherry-picked this fix in the `libpng16` branch and I added your name to the AUTHORS list.

Should you wish to share with me the native transliteration of your name, or the professional affiliation, please do. Alternatively, sending a PR with an update to the AUTHORS file would also work.

aYukiSekiguchi:
Thank you for the cherry-pick and for adding me to the AUTHORS list!

I contributed this as part of my work, so please list my professional affiliation.
My name can stay as-is in Latin script — no native transliteration needed.

Suggested AUTHORS entry:

 * ACCESS CO., LTD.
    - Yuki Sekiguchi

Happy to send a PR for this if you prefer.

PR #763, opened 2025-11-25
Correct and optimize RVV implementations.

chLFF:
## e4c9677961a42997d9a99c10485c51280cb2ad52
Simply replace vwaddu + vnsrl with vaaddu to reduce instructions.

## 84a43f3487f7e99df3c2b34b93dc4e211b969921
Fix `png_read_filter_row_paeth_rvv` where the output did not match the scalar version.

Also, the fix achieves ~14% speedup comparing with scalar (on a MUSEBOOK with SpacemiT K1).

Tested against reference scalar output to validate correctness and benchmark with [test.c](https://github.com/user-attachments/files/23736307/test.c).

Use `gcc -O3 -march=rv64imafdcv test.c -o test` to compile test.

PR #766, opened 2025-12-02
[RISCV] Fix and improve RVV png_read_filter

chLFF:
For earlier discussions and tests, please refer to #763. This PR is to retarget branch to libpng16.

ae991431719ed4b6b70a31825df696488c644d03
Simply replace vwaddu + vnsrl with vaaddu to reduce instructions.

8e8ef5c98e201bd36856259867ca079af77b2803
Fix png_read_filter_row_paeth_rvv where the output did not match the scalar version.

@filipwasil, may I please kindly ask for your review? Thank you.

### DIFF
--- a/riscv/filter_rvv_intrinsics.c
+++ b/riscv/filter_rvv_intrinsics.c
@@ -263,27 +263,22 @@ png_read_filter_row_paeth_rvv(size_t len, size_t bpp, unsigned char* row,
       /* x = *row */
       vuint8m1_t x = __riscv_vle8_v_u8m1(row, vl);
 
-      /* Calculate p = b - c and pc = a - c using widening subtraction */
-      vuint16m2_t p_wide = __riscv_vwsubu_vv_u16m2(b, c, vl);
-      vuint16m2_t pc_wide = __riscv_vwsubu_vv_u16m2(a, c, vl);
-
-      /* Convert to signed for easier manipulation */
-      size_t vl16 = __riscv_vsetvl_e16m2(bpp);
-      vint16m2_t p = __riscv_vreinterpret_v_u16m2_i16m2(p_wide);
-      vint16m2_t pc = __riscv_vreinterpret_v_u16m2_i16m2(pc_wide);
+      /* p = b - c and pc = a - c */
+      vuint16m2_t p = __riscv_vwsubu_vv_u16m2(b, c, vl);
+      vuint16m2_t pc = __riscv_vwsubu_vv_u16m2(a, c, vl);
 
       /* pa = |p| */
-      vbool8_t p_neg_mask = __riscv_vmslt_vx_i16m2_b8(p, 0, vl16);
-      vint16m2_t pa = __riscv_vrsub_vx_i16m2_m(p_neg_mask, p, 0, vl16);
+      vuint16m2_t tmp = __riscv_vrsub_vx_u16m2(p, 0, vl);
+      vuint16m2_t pa = __riscv_vminu_vv_u16m2(p, tmp, vl);
 
       /* pb = |pc| */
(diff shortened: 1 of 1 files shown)


ctruta:
@chLFF, from what I see, this fix addresses a real correctness issue. Your approach is sound. Before merging, could you amend the commit messages of your original commits (then force-push to your branch so we can see those here) to explain:

1. Why the original masked operations (`_m` suffix) were problematic (RVV mask-agnostic semantics)?
2. Why the `min(x, -x)` pattern correctly computes absolute value for this value range?

This should help everyone (myself included) with the understanding of your fix.

@filipwasil, I will run tests on QEMU-RISCV business-as-usual, but I will still very much appreciate your testing on real hardware, as well as your reviewing opinion.

ctruta:
Trying to understand my own questions if you please don't mind me:

> Why the `min(x, -x)` pattern correctly computes absolute value for this value range?

I'm asking because it's rather obvious that, in signed arithmetic, we have
`min(x, -x) = -|x|`

However, this code operates in unsigned 16-bit arithmetic, so, asking to clarify:
Why does `vminu(p, -p)` yield `|b - c|` in unsigned modular arithmetic for this value range?

chLFF:
**1. Fix: Replace mask-agnostic (`_m`) intrinsics with mask-undisturbed (`_mu`) for correctness**

The original masked operations (`_m` suffix) followed the RVV **mask-agnostic** policy, which does **not** guarantee the values in masked‑off elements. In the absolute‑value computation, we require **positive values to remain unchanged** in all vector lanes. Using `_mu` (mask‑undisturbed) ensures that masked‑off elements are preserved as they were in the destination vector, thereby maintaining correctness across the entire vector.

**2. Clarify: `min(x, -x)` correctly computes absolute value for the uint8 difference range**

When the difference of two `uint8_t` values is extended to `uint16_t`, the results fall into two ranges:
- Positive differences: `[0, 255]`
- Negative differences: `[65281, 65535]` (two’s‑complement of `[-255, -1]`)

The sequence `reverse‑sub‑zero + minu` maps values in `[32769, 65535]` to `[1, 32767]`. Applied to our case:
- Values in `[0, 255]` remain unchanged.
- Values in `[65281, 65535]` are mapped to `[1, 255]`, which corresponds precisely to `abs(x)` for `-255 < x < 0`.

Thus the `min(x, -x)` pattern correctly implements absolute value over the entire possible input range.

@ctruta I hope this clarifies your questions. Please let me know if any part needs further clarification.

chLFF:
Sorry for the late reply. I just noticed your commit message — thanks for taking care of it. I'll close this PR.

ctruta:
@chLFF, @filipwasil, thank you for your contributions!

@chLFF, if you can please share with me your name in Chinese, for inclusion in the AUTHORS file, that'd be great 😄

So I said that I will make the next release -- which I did -- but then I realized that the risk of NOT including the Paeth fix was higher than the risk of going ahead and including it, without reviewing and without testing on a real RISC-V machine.

On the positive side, I did run my usual regression testing on a QEMU-based RISC-V emulator running Ubuntu 24.04.2, and that went well. Needless to say, if anything is still broken, do not hesitate to send me your defect reports and your fixes.

chLFF:
@ctruta My name is 梁俊钊 (Liang Junzhao). I'm glad the fix could be included. I'll continue to test and report if any issues arise. Thanks again.

chLFF:
@ctruta My name is 梁俊钊 (Liang Junzhao). I'm glad the fix could be included. I'll continue to test and report if any issues arise. Thanks again.

chLFF:
@ctruta My name is 梁俊钊 (Liang Junzhao). I'm glad the fix could be included. I'll continue to test and report if any issues arise. Thanks again.

Issue #764, opened 2025-11-27
[Security Report] Out-of-bounds Read in libpng 1.6.51


flyfish101:
Hello, I found a new crash in libpng during testing, and it does not appear to be the same as the recently published CVE‑2025‑64720. According to the security advisory, CVE‑2025‑64720 was fully fixed in version 1.6.51, but I can still reliably reproduce an out‑of‑bounds read crash on a clean rebuild of libpng 1.6.51.
```
// test_driver.cpp

#include <stdint.h>
#include <stdlib.h>
#include <string.h>
#include <setjmp.h>
#include <png.h>

/* Memory buffer structure for PNG reading */
struct png_mem_buffer {
    const uint8_t *data;
    size_t size;
    size_t pos;
};

/* Custom read function for memory buffer */
static void png_read_from_buffer(png_structp png_ptr, png_bytep data, png_size_t length) {
    struct png_mem_buffer *buf = (struct png_mem_buffer*)png_get_io_ptr(png_ptr);
    if (buf->pos + length > buf->size) {
        png_error(png_ptr, "Read error");
    }
    memcpy(data, buf->data + buf->pos, length);
    buf->pos += length;
}

/* Test PNG transformations to trigger uncovered code paths */
static void test_png_transformations(const uint8_t *data, size_t size) {
    png_structp png_ptr = png_create_read_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
    if (!png_ptr) return;
    
    png_infop info_ptr = png_create_info_struct(png_ptr);
    if (!info_ptr) {
        png_destroy_read_struct(&png_ptr, NULL, NULL);
        return;
    }
    
    if (setjmp(png_jmpbuf(png_ptr))) {
        png_destroy_read_struct(&png_ptr, &info_ptr, NULL);
        return;
    }
    
    struct png_mem_buffer buffer = {data, size, 0};
    png_set_read_fn(png_ptr, &buffer, png_read_from_buffer);
    
    png_read_info(png_ptr, info_ptr);
    
    int width = png_get_image_width(png_ptr, info_ptr);
    int height = png_get_image_height(png_ptr, info_ptr);
    int color_type = png_get_color_type(png_ptr, info_ptr);
    int bit_depth = png_get_bit_depth(png_ptr, info_ptr);
    
    /* Sanity check dimensions */
    if (width <= 0 || height <= 0 || width > 2048 || height > 2048) {
[...]

PR #767, opened 2025-12-03
oss-fuzz: use nalloc fuzz target

catenacyber:
https://github.com/pnggroup/libpng/pull/760 was incomplete

>> Separate target, or augmenting the existing libpng_read_fuzzer?
> We can do either, but people preferred a separate target

So, here is the separate target

### DIFF
--- a/contrib/oss-fuzz/build.sh
+++ b/contrib/oss-fuzz/build.sh
@@ -42,6 +42,15 @@ $CXX $CXXFLAGS -std=c++11 -I. \
      -o $OUT/libpng_read_fuzzer \
      -lFuzzingEngine .libs/libpng16.a -lz
 
+# wrapper script to duplicate target, run with env var NALLOC_FREQ=32
+# having a separate target with allocations failures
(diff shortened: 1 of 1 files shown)


ctruta (approved):

ctruta:
> #760 was incomplete
> 
> > > Separate target, or augmenting the existing libpng_read_fuzzer?
> > > We can do either, but people preferred a separate target
> 
> So, here is the separate target

Thanks. I don't know if you've seen the release announcement, so let me just say that it was quite... something...
So we'll catch up with this follow-up at the next release, I guess.

catenacyber:
Thanks for the acknowledgment in the release note, I saw it indeed :-)

> so let me just say that it was quite... something...

Did fuzzing find CVE-2025-66293 ? (If not, why ?)

ctruta:
> Did fuzzing find CVE-2025-66293 ? (If not, why ?)

They didn't say, but I cannot imagine how this could have been found in any other way. See Issue #764.

PR #772, opened 2025-12-05
remove leftover ASM language from cmake project()

### LINKED ISSUE #770: libpng 1.6.52 + MSVC + CMake 4.2 configuration failure
The `cmake_minimum_version` directive in libpng 1.6.52 has updated to include version 4.2, up from 4.0 in 1.6.51. This brings [CMake policy CMP0194](https://cmake.org/cmake/help/latest/policy/CMP0194.html) into scope when CMake 4.1 or later is used. Unless a separate ASM compiler is installed, this will fail as MSVC is (correctly) no longer detected as an ASM compiler.

I see two possible fixes, the simplest is to add the following to the root CMakeLists.txt file:

```cmake
if(POLICY CMP0194)
  cmake_policy(SET CMP0194 OLD)
endif()
```

to restore the old behavior. This works on x64 since this platform does not use ASM.

The other option would be to only declare ASM as a project language on the platforms that actually require it - this would probably be more reliable when some hypothetical future CMake version removes the old behavior of this policy.

jcar87:
Remove `ASM` as a cmake language since it is no longer used, as per comment in https://github.com/pnggroup/libpng/issues/770#issuecomment-3613866682

Close https://github.com/pnggroup/libpng/issues/770

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ set(PNGLIB_SHARED_VERSION ${PNGLIB_ABI_VERSION}.${PNGLIB_REVISION}.${PNGLIB_SUBR
 
 project(libpng
         VERSION ${PNGLIB_VERSION}
-        LANGUAGES C ASM)
+        LANGUAGES C)
 
 include(CheckCSourceCompiles)
 include(CheckLibraryExists)


ajtribick:
Thanks, I confirm from my side that this fixes the build.

ctruta (approved):

ctruta:
I cherry-picked this contribution from `libpng18` into `libpng16`, and I'll make a new v1.6.x release shortly. See commit ff615fdcaf393ec2402bfafebadaa30b32fe5f3a.

This commit will make it back into `libpng18` after I do the tag merging, post-release.

ctruta:
> This commit will make it back into libpng18 after I do the tag merging, post-release.

Done. Thanks again for fixing this.

jbowler:
> I cherry-picked this contribution from `libpng18` into `libpng16`, and I'll make a new v1.6.x release shortly. See commit [ff615fd](https://github.com/pnggroup/libpng/commit/ff615fdcaf393ec2402bfafebadaa30b32fe5f3a).
> 
> This commit will make it back into `libpng18` after I do the tag merging, post-release.

You didn't remove the assembler file in question.

PR #774, opened 2025-12-07
Fuzz nalloc msan

catenacyber:
@ctruta here is the oss-fuzz build fix for MSAN + incorporation of the 3 fuzz targets in #764

### DIFF
--- a/contrib/oss-fuzz/build.sh
+++ b/contrib/oss-fuzz/build.sh
@@ -36,24 +36,29 @@ autoreconf -f -i
 make -j$(nproc) clean
 make -j$(nproc) libpng16.la
 
+for f in libpng_read_fuzzer libpng_colormap_fuzzer libpng_readapi_fuzzer libpng_transformations_fuzzer;
+do
 # build libpng_read_fuzzer.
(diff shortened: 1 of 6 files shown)


ctruta:
@catenacyber, hello, happy new year, and thank you for your continuing contributions to libpng's oss-fuzz. I *think* I proceeded correctly by not including your fourth commit because I *think* that commit was a mistake, and I *acknowledge* this should have gone through a review cycle. However, I want to publish a new libpng release by EOD today, so I opted to just fix your third commit in-place and to leave those three targets that you just added untouched.

If the resulting oss-fuzz additions turned out alright, or if they didn't -- regardless, please let me know.

catenacyber:
Thanks Cosmin.

Libpng master branch looks ok to me with regards to oss-fuzz

It was my intention to have `libpng_transformations_fuzzer.cc` do only `test_png_transformations` and not `test_png_image_colormap` which is done by libpng_colormap_fuzzer.cc but I failed to fixup the right commit

From my fuzzing, if I remember correctly, one of the 3 targets ended up finding an out-of-memory with a png announcing like 100k lines and 100k rows (which should be prevented by using libpng API to set limits)

PR #789, opened 2026-02-11
fix: prevent silent data corruption in pnm2png on EOF

uwezkhan:
fix(contrib): prevent silent data corruption in pnm2png tool

Previously, when converting PNM files to PNG, if the input file was
truncated or contained invalid numeric tokens, the pnm2png tool would
silently write zero pixels instead of reporting an error. This resulted
in corrupted PNG files being created without any warning to the user.

Changes made:
- Modified get_pnm_data() to return -1 on EOF instead of 0
- Modified get_pnm_value() to return -1 on parsing errors instead of 0
- Changed return type from png_uint_32 to int to allow error signaling
- Added clear error messages to stderr for both functions
- Updated all 6 call sites to check return values and abort on error
- Updated function prototypes

The conversion now properly fails with an error message when input
data is incomplete or invalid, preventing the creation of silently
corrupted output files.

Resolves FIXME comments at lines 612-614 and 647-650.

File modified: contrib/pngminus/pnm2png.c


uwezkhan:
Hey , i wanted to know that if this PR is helpful. or else let me close this PR if it is a low priority .

PR #790, opened 2026-02-11
oss-fuzz: fix build.sh for/if

catenacyber:
@ctruta looks like there was a bad typo in #784

See https://oss-fuzz-build-logs.storage.googleapis.com/log-da505c2a-61f8-4eb9-b630-512f4110c5e6.txt


ctruta:
> @ctruta looks like there was a bad typo in #784

I reviewed that PR but I didn't see the error
¯\\\_(ツ)\_/¯

PR #806, opened 2026-02-28
png_handle_sPLT: Add explicit INT32_MAX guard before casting to png_i…

uwezkhan:
Although current chunk length constraints prevent dl from exceeding INT32_MAX in practice, casting png_uint_32 to png_int_32 without an explicit bound check relies on implicit guarantees.

This patch adds an explicit INT32_MAX guard before the cast to eliminate implementation-defined behavior and improve defensive robustness.

### DIFF
--- a/pngrutil.c
+++ b/pngrutil.c
@@ -1629,7 +1629,7 @@ png_handle_sPLT(png_struct *png_ptr, png_info *info_ptr, png_uint_32 length)
    dl = (png_uint_32)(data_length / (unsigned int)entry_size);
    max_dl = PNG_SIZE_MAX / (sizeof (png_sPLT_entry));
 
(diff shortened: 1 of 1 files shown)


ctruta:
Hello, @uwezkhan, and thank you very much for your contribution.

The guard you mentioned doesn't actually trigger, because `png_read_chunk_header` obtains the chunk length via `png_get_uint_31`, which enforces `length <= 0x7FFFFFFF`. This is true regardless of the machine word size (32 bits or 64 bits or anything beyond).

More specifically, since `data_length <= length` and the minimum sPLT entry size is 6, `dl` is bounded by `0x7FFFFFFF / 6`, which is well below `INT32_MAX`.

That said, we still appreciate you looking into this area of libpng code.

PR #808, opened 2026-03-02
Create safe function to raise png_error on invalid ttime in pngwrite.c

Jeffin820:
Added a safe function(`png_ttime_validate`) that takes `png_ptr` and `ttime` as arguments and raises a `png_error` if `ttime` is invalid as stated in the `TODO `at line 540 in `pngwrite.c`.


Jeffin820:
Had to force push to pass the lint check

ctruta:
Thank you, @Jeffin820, for working on this. Addressing the TODO note in `png_convert_from_time_t` is a real and worthwhile goal.

However, after reviewing the update, I concluded that I cannot merge this in its current form. Here is why, and here is what a clean fix would look like.

#### It should not break the API

`png_convert_from_time_t` is a public, non-deprecated function. Changing its signature silently breaks every existing caller. The 1.8 branch permits removing long-deprecated symbols, but changing signature of an active one is a big no-no.

#### It should not break the validation logic

`png_ttime_validate` rejects any `time_t` above `0x7FFFFFFF`, which is a valid timestamp on 64-bit systems (any date after 2038). It also right-shifts a potentially negative signed value, which is implementation-defined in C89. The intent is not recoverable from the expression.

`gmtime` already is the validator. If it returns NULL, the value is invalid, and no hand-rolled arithmetic is needed. The entire `png_ttime_validate` function is doing work that `gmtime` does correctly and portably.

#### The new function should not be public API

Once exported via `PNG_EXPORT`, a function must be maintained indefinitely. A helper this narrow should be `static` or `/*PRIVATE*/`.

---

#### How to fix it cleanly

Leave `png_convert_from_time_t` exactly as-is (it already zero-fills and returns on failure, which is safe). Add a new function that takes a `png_struct *` so it can raise an error:

```
PNG_EXPORT(void, png_convert_from_time_t_strict,
   (png_struct *png_ptr, png_timep ptime, time_t ttime));
```

Inside it, call `gmtime`. If it returns NULL, call `png_error(png_ptr, "invalid time value")`. No separate validate function, no arithmetic, no change to the existing signature.

Jeffin820:
Implemented a strict version based on your suggestion with no arithmetic or validate function. Had to add it in `png.h` though, couldn't understand if you meant it to be private or public.

Jeffin820:
`png_timep` was depreciated(?) in `png.h`, so had to use `png_time *ptime` instead.

Jeffin820:
@ctruta Should the change be documented if it is merged and should there be any other additional comments specifying the purpose of the function?

PR #812, opened 2026-03-05
Fix memory leak and API misuse in libpng_transformations_fuzzer

### LINKED ISSUE #809: libpng_transformations_fuzzer: memory leak via longjmp and API misuse with png_read_png
## Summary

This is **not** a bug in libpng. It is a bug in `libpng_transformations_fuzzer` (`contrib/oss-fuzz/libpng_transformations_fuzzer.cc`) that causes:

1. **Memory leak** when `longjmp` bypasses `free()` calls
2. **API contract violation** by mixing `png_set_*` calls with `png_read_png()`

### False Positive Impact

The memory leak triggers LeakSanitizer on the **very first seed input**, preventing the fuzzer from starting. Any leak report from this fuzzer is a **false positive** — the leak is in the harness, not in libpng.

If left unfixed:
- **Wasted developer time**: Maintainers investigating reports that are not real libpng bugs
- **Noise in OSS-Fuzz dashboards**: Persistent unfixed "bugs" that are actually harness defects
- **Amplified impact in the AI era**: AI-assisted fuzzing tools increasingly reference existing OSS-Fuzz harnesses as ground truth. A buggy harness pattern can be copied by LLM-based harness generators, multiplying false positives across projects

## Bug 1: Memory leak via longjmp

In `test_png_transformations`, `row` is allocated with `malloc` (line 93) **after** `setjmp` (line 35). When `png_read_row` triggers `png_error` → `longjmp`, execution jumps to the error handler which only calls `png_destroy_read_struct` and returns — `row` is never freed.

Per the C standard (§7.13.2.1), `row` is a non-volatile local modified between `setjmp` and `longjmp`, so its value is **indeterminate** after `longjmp`.

```
==14==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 1 byte(s) in 1 object(s) allocated from:
    #0 malloc
    #1 test_png_transformations(...) libpng_transformations_fuzzer.cc:93
INFO: a leak has been found in the initial corpus.
```

## Bug 2: API misuse with png_read_png

`test_png_read_png_api` calls `png_set_scale_16()`, `png_set_packing()`, `png_set_expand()` before `png_read_png(..., PNG_TRANSFORM_IDENTITY, ...)`.

`libpng-manual.txt` (lines 1170–1171):
> "You must use png_transforms and not call any png_set_transform() functions when you use png_read_png()."

## Experimental Evidence

| Version | Lines | Functions | Branches | Status |
|---------|-------|-----------|----------|--------|
| Original | 53.5% | 55.6% | 35.6% | LeakSanitizer crash on first seed — fuzzer cannot start |
| Fixed | 53.6% | 55.6% | 35.6% | 28,155 executions in 30s, 0 crashes |

OwenSanzas:
Fixes #809

## Changes

1. **Memory leak fix**: `row` buffer allocated after `setjmp` is not freed when `png_read_row` triggers `longjmp`. Declare `row` as `volatile` before `setjmp` and free it in the error handler (required by C standard §7.13.2.1).

2. **API misuse fix**: Replace `png_set_*` + `png_read_png(PNG_TRANSFORM_IDENTITY)` with `PNG_TRANSFORM_SCALE_16 | PNG_TRANSFORM_PACKING | PNG_TRANSFORM_EXPAND`, per `libpng-manual.txt` lines 1170–1171.

## Before / After

| Version | Lines | Functions | Branches | Status |
|---------|-------|-----------|----------|--------|
| Original | 53.5% | 55.6% | 35.6% | LeakSanitizer crash on first seed |
| Fixed | 53.6% | 55.6% | 35.6% | 28,155 executions in 30s, 0 crashes |

## Test plan

- [x] Fixed fuzzer runs without LeakSanitizer crash
- [x] Coverage unchanged — confirms behavioral equivalence

### DIFF
--- a/contrib/oss-fuzz/libpng_transformations_fuzzer.cc
+++ b/contrib/oss-fuzz/libpng_transformations_fuzzer.cc
@@ -32,7 +32,13 @@ static void test_png_transformations(const uint8_t *data, size_t size) {
         return;
     }
 
+    /* Declare heap pointers before setjmp so they can be freed on longjmp.
+       Must be volatile per C standard §7.13.2.1: non-volatile locals modified
+       between setjmp and longjmp have indeterminate values after longjmp. */
+    volatile png_bytep row = NULL;
+
     if (setjmp(png_jmpbuf(png_ptr))) {
+        free((void*)row);
         png_destroy_read_struct(&png_ptr, &info_ptr, NULL);
         return;
     }
@@ -90,15 +96,16 @@ static void test_png_transformations(const uint8_t *data, size_t size) {
 
     /* Read image data to execute transformations */
     size_t rowbytes = png_get_rowbytes(png_ptr, info_ptr);
-    png_bytep row = (png_bytep)malloc(rowbytes);
+    row = (png_bytep)malloc(rowbytes);
     if (row) {
         int y;
         for (y = 0; y < height && y < 100; y++) { /* Limit rows for performance */
             png_read_row(png_ptr, row, NULL);
         }
-        free(row);
     }
 
+    free((void*)row);
+    row = NULL;
     png_destroy_read_struct(&png_ptr, &info_ptr, NULL);
 }
 
@@ -184,13 +191,12 @@ static void test_png_read_png_api(const uint8_t *data, size_t size) {
     struct png_mem_buffer buffer = {data, size, 0};
     png_set_read_fn(png_ptr, &buffer, png_read_from_buffer);
 
-    /* Set up transformations before reading */
-    png_set_scale_16(png_ptr);
-    png_set_packing(png_ptr);
-    png_set_expand(png_ptr);
-
-    /* Use png_read_png which should trigger OSS_FUZZ_png_read_png path */
-    png_read_png(png_ptr, info_ptr, PNG_TRANSFORM_IDENTITY, NULL);
+    /* Use png_read_png with transform flags (libpng-manual.txt lines 1170-1171:
+       "You must use png_transforms and not call any png_set_transform()
+       functions when you use png_read_png().") */
+    png_read_png(png_ptr, info_ptr,
+                 PNG_TRANSFORM_SCALE_16 | PNG_TRANSFORM_PACKING | PNG_TRANSFORM_EXPAND,
+                 NULL);
 
     png_destroy_read_struct(&png_ptr, &info_ptr, NULL);
 }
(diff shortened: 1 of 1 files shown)


contrib/oss-fuzz/libpng_transformations_fuzzer.cc
```diff
@@ -90,15 +96,16 @@ static void test_png_transformations(const uint8_t *data, size_t size) {
...
     if (row) {
         int y;
         for (y = 0; y < height && y < 100; y++) { /* Limit rows for performance */
             png_read_row(png_ptr, row, NULL);
         }
-        free(row);
     }
 
+    free((void*)row);
+    row = NULL;
```
ctruta: `row` is a local variable, which makes the `row = NULL` line unnecessary. This line should be removed.

contrib/oss-fuzz/libpng_transformations_fuzzer.cc
```diff
@@ -184,13 +191,12 @@ static void test_png_read_png_api(const uint8_t *data, size_t size) {
...
     png_set_read_fn(png_ptr, &buffer, png_read_from_buffer);
 
-    /* Set up transformations before reading */
-    png_set_scale_16(png_ptr);
-    png_set_packing(png_ptr);
-    png_set_expand(png_ptr);
-
-    /* Use png_read_png which should trigger OSS_FUZZ_png_read_png path */
-    png_read_png(png_ptr, info_ptr, PNG_TRANSFORM_IDENTITY, NULL);
+    /* Use png_read_png with transform flags (libpng-manual.txt lines 1170-1171:
```
ctruta: Mentioning the libpng manual is fine, but mentioning an exact quote at an exact line is going to go out of sync as soon as we update this manual.
Simply saying *"Use png_read_png with transform flags"* should be sufficient.

ctruta (changes requested):
Both fixes are conceptually correct. The `volatile` declaration before `setjmp` and `free` in the error handler are textbook §7.13.2.1 compliance, and the `PNG_TRANSFORM_*` flag mappings are accurate.

See the two inline comments for minor correction requests.

Pre-existing: the `palette` allocation around line 70 has the same longjmp-leak pattern. If `png_set_quantize` calls `png_error` internally, `palette` leaks. Lower risk, but worth a follow-up.

OwenSanzas:
Updated per review feedback:

1. Removed unnecessary `row = NULL` assignment
2. Simplified `png_read_png` comment (no longer references exact manual line numbers)
3. Also fixed the `palette` longjmp-leak you mentioned — same pattern as the `row` fix: declared `volatile png_colorp palette` before `setjmp`, freed in error handler, and set to `NULL` after normal-path `free`

Verified: coverage is unchanged between original and fixed versions (706 new_units_added, ~680 exec/sec over 30s runs with ASan).

ctruta:
Integrated in trunk (even if GitHub doesn't show it that way).
Many thanks for your contribution!

PR #813, opened 2026-03-05
Fix API misuse in libpng_readapi_fuzzer

### LINKED ISSUE #811: libpng_readapi_fuzzer: API misuse mixing png_set_* with png_read_png
## Summary

This is **not** a bug in libpng. It is a bug in `libpng_readapi_fuzzer` (`contrib/oss-fuzz/libpng_readapi_fuzzer.cc`) that violates the documented API contract by calling `png_set_*` transform functions before `png_read_png()`.

### False Positive Risk

While this violation does not cause crashes today, it poses a growing risk:

- **Future libpng changes**: If a future version enforces the documented API contract (e.g., by resetting `png_ptr->transformations` inside `png_read_png` before applying `png_transforms`), this harness would silently lose its transform coverage — or worse, produce false positive crash reports
- **Propagation via AI**: LLM-based harness generators increasingly reference existing OSS-Fuzz harnesses as examples. This incorrect `png_set_*` + `png_read_png(PNG_TRANSFORM_IDENTITY)` pattern may be copied into new harnesses, spreading the API violation across projects

## Bug Description

Lines 44–46 call individual transform functions, then line 49 calls `png_read_png` with `PNG_TRANSFORM_IDENTITY`:

```cpp
png_set_scale_16(png_ptr);
png_set_packing(png_ptr);
png_set_expand(png_ptr);
png_read_png(png_ptr, info_ptr, PNG_TRANSFORM_IDENTITY, NULL);
```

`libpng-manual.txt` (lines 1170–1171):
> "You must use png_transforms and not call any png_set_transform() functions when you use png_read_png()."

The manual explicitly distinguishes two interfaces (lines 1097–1103): `PNG_TRANSFORM_*` flags for `png_read_png`, and `png_set_*()` calls for the low-level API. This technically works because `png_set_*` only sets flag bits, but the documentation uses "must", indicating a mandatory requirement.

## Suggested Fix

```diff
-    png_set_scale_16(png_ptr);
-    png_set_packing(png_ptr);
-    png_set_expand(png_ptr);
-    png_read_png(png_ptr, info_ptr, PNG_TRANSFORM_IDENTITY, NULL);
+    png_read_png(png_ptr, info_ptr,
+                 PNG_TRANSFORM_SCALE_16 | PNG_TRANSFORM_PACKING | PNG_TRANSFORM_EXPAND,
+                 NULL);
```

## Experimental Evidence

| Version | Lines | Functions | Branches |
|---------|-------|-----------|----------|
| Original | 27.8% | 36.8% | 17.9% |
| Fixed | 27.8% | 36.8% | 18.0% |

Coverage is identical — the violation does not affect runtime behavior in the current implementation.

OwenSanzas:
Fixes #811

## Changes

Replace `png_set_scale_16()` / `png_set_packing()` / `png_set_expand()` + `png_read_png(PNG_TRANSFORM_IDENTITY)` with equivalent `PNG_TRANSFORM_*` flags passed directly to `png_read_png`.

`libpng-manual.txt` (lines 1170–1171):
> "You must use png_transforms and not call any png_set_transform() functions when you use png_read_png()."

## Before / After

| Version | Lines | Functions | Branches |
|---------|-------|-----------|----------|
| Original | 27.8% | 36.8% | 17.9% |
| Fixed | 27.8% | 36.8% | 18.0% |

Coverage is identical — confirms behavioral equivalence while fixing the documented API contract violation.

## Test plan

- [x] Fixed fuzzer runs normally (336,832 executions in 30s, 0 crashes)
- [x] Coverage unchanged

### DIFF
--- a/contrib/oss-fuzz/libpng_readapi_fuzzer.cc
+++ b/contrib/oss-fuzz/libpng_readapi_fuzzer.cc
@@ -40,9 +40,7 @@ static void test_png_read_png_api(const uint8_t *data, size_t size) {
     struct png_mem_buffer buffer = {data, size, 0};
     png_set_read_fn(png_ptr, &buffer, png_read_from_buffer);
 
-    /* Use png_read_png with transform flags (libpng-manual.txt lines 1170-1171:
-       "You must use png_transforms and not call any png_set_transform()
-       functions when you use png_read_png().") */
+    /* Use png_read_png with transform flags */
     png_read_png(png_ptr, info_ptr,
                  PNG_TRANSFORM_SCALE_16 | PNG_TRANSFORM_PACKING | PNG_TRANSFORM_EXPAND,
                  NULL);
--- a/contrib/oss-fuzz/libpng_readapi_fuzzer.cc
+++ b/contrib/oss-fuzz/libpng_readapi_fuzzer.cc
@@ -40,13 +40,12 @@ static void test_png_read_png_api(const uint8_t *data, size_t size) {
     struct png_mem_buffer buffer = {data, size, 0};
     png_set_read_fn(png_ptr, &buffer, png_read_from_buffer);
 
-    /* Set up transformations before reading */
-    png_set_scale_16(png_ptr);
-    png_set_packing(png_ptr);
-    png_set_expand(png_ptr);
-
-    /* Use png_read_png which should trigger OSS_FUZZ_png_read_png path */
-    png_read_png(png_ptr, info_ptr, PNG_TRANSFORM_IDENTITY, NULL);
+    /* Use png_read_png with transform flags (libpng-manual.txt lines 1170-1171:
+       "You must use png_transforms and not call any png_set_transform()
+       functions when you use png_read_png().") */
+    png_read_png(png_ptr, info_ptr,
+                 PNG_TRANSFORM_SCALE_16 | PNG_TRANSFORM_PACKING | PNG_TRANSFORM_EXPAND,
+                 NULL);
 
     png_destroy_read_struct(&png_ptr, &info_ptr, NULL);
 }


contrib/oss-fuzz/libpng_readapi_fuzzer.cc
```diff
@@ -40,13 +40,12 @@ static void test_png_read_png_api(const uint8_t *data, size_t size) {
...
 
-    /* Set up transformations before reading */
-    png_set_scale_16(png_ptr);
-    png_set_packing(png_ptr);
-    png_set_expand(png_ptr);
-
-    /* Use png_read_png which should trigger OSS_FUZZ_png_read_png path */
-    png_read_png(png_ptr, info_ptr, PNG_TRANSFORM_IDENTITY, NULL);
+    /* Use png_read_png with transform flags (libpng-manual.txt lines 1170-1171:
+       "You must use png_transforms and not call any png_set_transform()
```
ctruta: Same comment as in PR 812: Mentioning the libpng manual is fine, but an exact quote at an exact line in this manual will be out of sync as soon as we update this manual. Simply saying *"Use `png_read_png` with transform flags"* should be sufficient.

ctruta (changes requested):
Correct. Same API misuse fix as the second change in  PR 812, applied to the other fuzzer file. The flag mappings are accurate.

See the inline comment.

Side note, outside the scope of this pull request: `libpng_readapi_fuzzer.cc` is now a strict subset of `libpng_transformations_fuzzer.cc`. Both contain identical `test_png_read_png_api` code. Unless there's an intended (distinct) functionality that justifies this fuzzer's existence, this fuzzer should be removed altogether.

OwenSanzas:
Updated: simplified the comment to just "Use png_read_png with transform flags" per review feedback.

ctruta:
Integrated in trunk (even if GitHub doesn't show it that way).
Many thanks for your contribution!

PR #821, opened 2026-03-13
Implement spell check in pngstruct.h

Jeffin820:
`png_struct` was incorrectly named as `pnt_struct` in `png_file_add_chunk` present in `pngstruct.h`. Dates were changed from 2025 to 2026.

### DIFF
--- a/pngstruct.h
+++ b/pngstruct.h
@@ -1,6 +1,6 @@
 /* pngstruct.h - internal structures for libpng
  *
- * Copyright (c) 2018-2025 Cosmin Truta
+ * Copyright (c) 2018-2026 Cosmin Truta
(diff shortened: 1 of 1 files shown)


pngstruct.h
```diff
@@ -104,7 +104,7 @@ typedef enum
    (((png_ptr)->chunks & png_chunk_flag_from_index(i)) != 0)
    /* The chunk has been recorded in png_struct */
 
-#define png_file_add_chunk(pnt_ptr, i)\
+#define png_file_add_chunk(png_ptr, i)\
```
ctruta, line 107: Thanks for the catch, @Jeffin820. Even though this defect was latent, I still appreciate the fact that you found it and fixed it!

I will amend your commit message. I think it's better if we explain what's going on here in a little more detail.

ctruta (approved):

Issue #828, opened 2026-03-24
Heap-Buffer-Overflow (Read) in png_get_iCCP via 3 bytes iCCP data


FuzzAnything-ORG:
# Heap-Buffer-Overflow (Read) in png_get_iCCP via 3 bytes iCCP data

### Version

v1.6.55-123-g28cb99fe6

## Description

A heap-buffer-overflow bug exists in libpng's `png_get_iCCP` function located at `pngget.c:730`. This occurs because `png_get_iCCP` assumes that the stored ICC profile contains at least 4 bytes to read a `png_uint_32` length value, but it does not validate the actual stored profile length (`info_ptr->iccp_proflen`).

**Vulnerability Type:** Heap-Buffer-Overflow (READ)

**Root Cause:** When an ICC profile shorter than 4 bytes is stored via `png_set_iCCP`, the subsequent call to `png_get_iCCP` attempts to read 4 bytes from an undersized buffer to extract the profile length using `png_get_uint_32()`. This results in an out-of-bounds read.

**Vulnerable Code in pngget.c:**
```c
png_uint_32 png_get_iCCP(const png_struct *png_ptr, png_info *info_ptr,
    char **name, int *compression_type,
    png_byte **profile, png_uint_32 *proflen)
{
    if (png_ptr != NULL && info_ptr != NULL &&
        (info_ptr->valid & PNG_INFO_iCCP) != 0 &&
        name != NULL && profile != NULL && proflen != NULL) // BUG: insufficient validation on profile length
    {
        *name = info_ptr->iccp_name;
        *profile = info_ptr->iccp_profile;
        *proflen = png_get_uint_32(info_ptr->iccp_profile); // BUG: assumes ≥4 bytes
        // ...
    }
}
```

## PoC Code

```cpp
#include <cstdint>
#include <cstdlib>
#include <cstring>
#include <csetjmp>
#include "png.h"

static void error_handler(png_structp png_ptr, png_const_charp error_msg) {
    (void)error_msg;
    longjmp(png_jmpbuf(png_ptr), 1);
}

static void warning_handler(png_structp png_ptr, png_const_charp warning_msg) {
    (void)png_ptr;
    (void)warning_msg;
}

int main() {
    png_structp png_ptr = png_create_read_struct(PNG_LIBPNG_VER_STRING, NULL,
                                                  error_handler, warning_handler);
    if (png_ptr == NULL) {
        return 1;
    }
    
[...]

PR #829, opened 2026-03-24
Add validation for iCCP profile length in png_get_iCCP and png_set_iCCP

FuzzAnything-ORG:
This patch is used to fix the heap-buffer-overflow problem as described in https://github.com/pnggroup/libpng/issues/828.

The root cause of this bug is insufficient validation for the input ICC profile length.

This patch fixes this bug by adding checks in the iCCP setting and getting functions.

As the [ICC Specification](https://www.color.org/specification/ICC.1-2022-05.pdf) (Section 7.2.1) requires at least 128 bytes for ICC header, this patch checks for such a length.

### DIFF
--- a/pngget.c
+++ b/pngget.c
@@ -727,6 +727,11 @@ png_get_iCCP(const png_struct *png_ptr, png_info *info_ptr,
    {
       *name = info_ptr->iccp_name;
       *profile = info_ptr->iccp_profile;
+      if (info_ptr->iccp_proflen < 128)
+      {
+         png_warning(png_ptr, "Invalid iCCP profile length (too small)");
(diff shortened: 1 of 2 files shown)


ctruta:
Following up on the response in #828:

The proposed patch is appreciated, but the underlying issue is not a library defect. Passing a buffer shorter than a valid ICC profile header to `png_set_iCCP` violates the ICC specification (ICC.1:2022, section 7.2), which requires a 128-byte header. The iCCP API expects the caller to supply a valid ICC profile.

We may add a defensive check as a hardening measure at our discretion, but this does not warrant a security fix.

Issue #830, opened 2026-03-24
Division by Zero in png_image_write_main with missing check for Zero Width


FuzzAnything-ORG:
## Division by Zero in png_image_write_main with missing check for Zero Width

Dear libpng developers,

I found an UndefinedBehaviorSanitizer (UBSan) error when using the simplified write API with an image having zero width.

### Version

v1.6.55-124-gd71162137

### Description

The crash is a division by zero in the `png_image_write_main` function at `pngwrite.c:2053:44`. The vulnerability occurs when `png_image_write_to_memory()` is called with an image structure where width is set to 0.

```
2033    if (image->width <= 0x7fffffffU/channels) /* no overflow */
        {
2036        png_uint_32 png_row_stride = image->width * channels;

            ...
2053        if (image->height > 0xffffffffU/png_row_stride)
                png_error(image->opaque->png_ptr, "memory image too large");
        }
```
The calculation flow is:
1. `png_row_stride = image->width * channels` (line 2036) - When width=0, this becomes 0
2. Division check: `image->height > 0xffffffffU/png_row_stride` (line 2053) - Division by zero

The library fails to validate that `image->width` must be greater than 0 before performing calculations involving division by row stride. The simplified API documentation does not explicitly reject zero dimensions, allowing this undefined behavior to occur.

## PoC Code

```cpp
#include <cstdint>
#include <cstdlib>
#include <cstring>
#include "png.h"

int main() {
    png_image image;
    memset(&image, 0, sizeof(image));
    image.version = PNG_IMAGE_VERSION;
    
    image.width = 0;      // Division by zero trigger
    image.height = 100;
    image.format = PNG_FORMAT_RGB;
    
    size_t buffer_size = 1024;
    void* buffer = malloc(buffer_size);
    if (!buffer) return 1;
    memset(buffer, 0, buffer_size);
    

    png_alloc_size_t memory_size = 0;
    
    // Try to write to memory - this triggers the division by zero
    // at pngwrite.c:2053: image->height > 0xffffffffU/png_row_stride
    // where png_row_stride = width * channels = 0
    png_image_write_to_memory(&image, NULL, &memory_size,
                              0, buffer, 0, NULL);
    
    free(buffer);
    png_image_free(&image);
    
    return 0;
}

```

### Reproduction Steps

1. Compile the PoC:
```bash
clang++ -g -fsanitize=address,undefined \
    -I/path/to/libpng/include \
    poc.cpp -o poc \
[...]

PR #831, opened 2026-03-24
Avoid divison by zero with in simpilfied writer API

FuzzAnything-ORG:
This patch adds a check to avoid the division by zero bug mentioned in https://github.com/pnggroup/libpng/issues/830.

Normally, when we try to read an image with zero width, the reader will reject it with the following checks:

```
void /* PRIVATE */
png_check_IHDR(const png_struct *png_ptr,
    png_uint_32 width, png_uint_32 height, int bit_depth, ...)
{
   int error = 0;

   /* Check for width and height valid values */
   if (width == 0)
   {
      png_warning(png_ptr, "Image width is zero in IHDR");
      error = 1;
   }
   // ... other checks...
   
   if (error == 1)
      png_error(png_ptr, "Invalid IHDR data");
}
```
It is ok if we try to read such a malformed image.
But the checks for the image width in the simplified writer API are missing. This patch adds the check to avoid potential bugs. With this patch, the warning from UBSan was missing.

### DIFF
--- a/pngwrite.c
+++ b/pngwrite.c
@@ -2030,7 +2030,7 @@ png_image_write_main(void *argument)
    {
       unsigned int channels = PNG_IMAGE_PIXEL_CHANNELS(image->format);
 
-      if (image->width <= 0x7fffffffU/channels) /* no overflow */
+      if (image->width <= 0x7fffffffU/channels && image->width != 0) /* no overflow */
       {
          png_uint_32 check;
          png_uint_32 png_row_stride = image->width * channels;


ctruta:
Following up on the response in #830:

The proposed patch is appreciated, but the underlying issue is not a library defect. Passing `width = 0` violates the PNG specification (ISO/IEC 15948, section 11.2.2), which requires non-zero dimensions. The simplified write API expects valid image parameters; supplying invalid ones is caller error.

We may add a defensive check as a hardening measure at our discretion, but this does not warrant a security fix.

PR #833, opened 2026-03-27
fix overflow checks in visualpng and fix build issue

uwezkhan:
Fixes overflow checks in VisualPng that were broken and causing build issues.
Also makes the image allocation checks safer to avoid overflow with large images.


uwezkhan:
Hi, just checking in on this PR.

Happy to make any changes or provide more details if needed.

PR #845, opened 2026-04-12
fix: eliminate size_t-to-uInt truncation in progressive reader zlib interface

mohammadmseet-hue:
## Summary

The progressive reader's `png_process_IDAT_data()` casts `PNG_ROWBYTES()` (returns `size_t`) directly to `uInt` (32-bit) when setting `zstream.avail_out` and `zstream.avail_in`. On 64-bit systems with large images, the cast silently truncates the value.

One of these casts has an explicit **TODO** comment acknowledging the problem:

```c
/* TODO: WARNING: TRUNCATION ERROR: DANGER WILL ROBINSON: */
png_ptr->zstream.avail_out = (uInt)(PNG_ROWBYTES(png_ptr->pixel_depth,
    png_ptr->iwidth) + 1);
```

The sequential reader (`png_read_IDAT_data` in pngrutil.c) already handles this correctly by capping at `ZLIB_IO_MAX`. The progressive reader was missing this treatment entirely.

## Impact

When `avail_out` truncates to a small value, zlib decompresses only that many bytes into `row_buf`. The row-completion check (`avail_out == 0`) then fires prematurely, causing `png_push_process_row()` to process a row buffer that is mostly uninitialized — `row_info.rowbytes` uses the full (non-truncated) `PNG_ROWBYTES`, so `png_read_filter_row()` and the subsequent `memcpy` to `prev_row` operate on the full row width of uninitialized heap memory.

Reachable when applications use `png_set_user_limits()` with large limits or build with the shipped `nolimits.dfa` configuration.

## Changes

**pngpread.c** (4 truncation sites fixed):
- `avail_in` assignment: cap at `ZLIB_IO_MAX` instead of direct `(uInt)` cast
- `avail_out` assignment (initial setup): same `ZLIB_IO_MAX` cap
- `avail_out` assignment (decompression loop): same fix, removes the TODO/DANGER comment
- Row-completion check: changed from `avail_out == 0` to position-based check (`next_out - row_buf >= row_size`), since `avail_out` may now represent only a chunk of the full row

**pngrutil.c** (1 site):
- APNG progressive reader reset: same `ZLIB_IO_MAX` cap

**pngtrans.c** (defense-in-depth):
- `png_set_user_transform_info()`: validate `user_transform_depth` and `user_transform_channels` before casting to `png_byte`, rejecting values that would wrap

## Testing

All 36 existing tests pass (Debug build, ASan+UBSan build).

### DIFF
--- a/pngpread.c
+++ b/pngpread.c
@@ -373,9 +373,16 @@ png_push_read_chunk(png_struct *png_ptr, png_info *info_ptr)
       png_ptr->idat_size = png_ptr->push_length;
       png_ptr->process_mode = PNG_READ_IDAT_MODE;
       png_push_have_info(png_ptr, info_ptr);
-      png_ptr->zstream.avail_out =
-          (uInt) PNG_ROWBYTES(png_ptr->pixel_depth,
-          png_ptr->iwidth) + 1;
+      {
+         size_t out_size = PNG_ROWBYTES(png_ptr->pixel_depth,
+             png_ptr->iwidth) + 1;
+         uInt out = ZLIB_IO_MAX;
+
+         if (out > out_size)
+            out = (uInt)out_size;
+
+         png_ptr->zstream.avail_out = out;
+      }
       png_ptr->zstream.next_out = png_ptr->row_buf;
       return;
    }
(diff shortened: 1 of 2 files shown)


ctruta (changes requested):
Thank you, @mohammadmseet-hue. You've identified a real bug that has been sitting in the progressive reader for a long time (the TODO/DANGER comments are there for a reason), and the analysis in your PR description is thorough. It's clear that you studied the sequential reader's approach and understood the underlying truncation problem.

I do need to request changes, though, because the fix as written introduces two new issues in the same large-row scenario it's meant to address. Both stem from the same root cause: the values are capped at `ZLIB_IO_MAX` but never re-fed when zlib consumes a partial chunk.

1. `avail_in` is set once before the loop.
   If `buffer_length > ZLIB_IO_MAX`, the loop exits after consuming the first capped chunk and the remaining input bytes are silently dropped. The sequential reader handles this by tracking the unconsumed remainder in a separate accumulator and re-chunking each iteration (see pngrutil.c around lines 573-588).

2. When `avail_out` runs out mid-row, `next_out` is reset to `row_buf`.
   Your position-based completion check correctly avoids calling `png_push_process_row` on a partial row. But then, the next iteration enters the `!(avail_out > 0)` block, which resets `next_out` back to `row_buf`, overwriting the partial row data just decompressed. The sequential reader avoids this by letting `next_out` advance and tracking remaining output capacity separately.

The right fix is to restructure the decompression loop to match the sequential reader's chunking strategy: maintain `size_t` accumulators for remaining input and output, chunk them down to `uInt` range each iteration, and distinguish "row complete" from "chunk exhausted" in the `avail_out` logic.

A couple of smaller notes:

- The `pngtrans.c` validation is reasonable on its own, but it's unrelated to the progressive reader bug. Please split it into a separate commit (or, better yet, a separate PR) so that each change can be reviewed and applied independently.
- There are three slightly different idioms for the `ZLIB_IO_MAX` cap across the changed sites. For consistency, please use the same pattern the sequential reader uses: `uInt avail = ZLIB_IO_MAX; if (avail > x) avail = (uInt)x;`.

I hope this feedback is useful. The bug analysis and the position-based completion check show you understand the problem well.

Looking forward to seeing a revised version :-)

mohammadmseet-hue:
Thanks for your thorough reply, changes validated and done. And yes it was, indeed, very useful :D

ctruta (changes requested):
Thanks for the rework. This is a substantial improvement.

Two things to fix before this can land, both in the same code path:

1. A guard is missing on `remaining_out` underflow. In the `!(avail_out > 0)` branch, the subtraction `row_size - produced` is unsigned. Zlib respects `avail_out`, so `produced > row_size` should not be reachable today, but this is tricky code and a future refactor could break the invariant. If it ever does, the unsigned subtraction wraps silently. Please add a defensive guard before the subtraction, and call `png_error` with a message like `"IDAT: row buffer overrun"` if `produced` exceeds `row_size`.

2. The row-completion check uses `>=`, which is not incorrect, but the existing pattern throughout libpng is `== 0` on `avail_out` (or the equivalent exact comparison on position). With the guard from item 1 in place, `>` is already caught, so `==` is sufficient here, and consistent with what we do elsewhere in the codebase.

I think this is all (if I haven't missed anything).

mohammadmseet-hue:
Both addressed:

1. **Defensive guard:** Added `png_error(png_ptr, "IDAT: row buffer overrun")` before the `row_size - produced` subtraction when `produced > row_size`.
2. **`>=` → `==`:** Changed the row-completion check to exact comparison, consistent with the existing `avail_out == 0` pattern.

Force-pushed with both fixes.

PR #847, opened 2026-04-13
fix: add overflow-checked size arithmetic for rowbytes computation

mohammadmseet-hue:
## Summary

Several sites across the library compute `row_info->rowbytes` or allocate buffers using bare `size_t` multiplications (`(size_t)width * bytes_per_pixel`) that can silently wrap on overflow. While `png_check_IHDR` limits width to prevent overflow for standard pixel depths, the pattern itself is error-prone and offers no protection for user transforms with custom pixel depths or future code paths.

This patch introduces overflow-checked size arithmetic primitives and systematically replaces all unchecked rowbytes multiplications:

**New primitives (pngpriv.h, png.c):**
- `png_mul_size(a, b)` — overflow-checked `size_t` multiplication macro, returns 0 on overflow. Uses the same division-based check as `png_malloc_array_checked()`.
- `png_rowbytes_checked(png_ptr, pixel_depth, width)` — checked version of `PNG_ROWBYTES` that calls `png_error` on overflow. For use at allocation sites where `png_structrp` is available.

**Updated PNG_ROWBYTES macro (pngpriv.h):**
- The `>= 8 bpp` case now uses `png_mul_size` internally, returning 0 on overflow instead of wrapping to a small value that could cause undersized allocations.

**Allocation and initialization sites (10 files):**
- All `PNG_ROWBYTES` calls at buffer allocation sites replaced with `png_rowbytes_checked`: `png_handle_IHDR`, `png_read_start_row`, `png_write_start_row`, `png_write_IHDR`, `png_set_IHDR`, `png_read_row`, `png_write_row`, `png_push_process_row`, `png_read_reinit` (APNG), `png_write_reinit` (APNG), progressive reader IDAT, and filter buffer allocation.
- `png_read_png` row pointer allocation uses `png_mul_size` with explicit overflow check.

**Transform rowbytes (pngrtran.c, arm/arm_init.c):**
- All 15 bare `(size_t)row_width * N` rowbytes assignments in `png_do_read_filler`, `png_do_unpack`, `png_do_scale_16_to_8`, `png_do_chop`, `png_do_expand_palette`, and `png_do_expand_16` replaced with `png_mul_size` as defense-in-depth.
- ARM NEON palette expansion paths use `png_rowbytes_checked` (has `png_ptr`) and fix a missing `(size_t)` cast on pointer arithmetic.

**Test (contrib/libtests/pngoverflow.c):**
- New test covering `png_mul_size` and `PNG_ROWBYTES` with normal values, boundary conditions, and overflow inputs.

## Testing
- All 38 tests pass (37 existing + 1 new overflow test)
- Tested under AddressSanitizer + UndefinedBehaviorSanitizer with zero warnings
- Clean C89 compilation with `-Wall -Wextra -pedantic`
- No functional change for images within standard limits

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -390,6 +390,9 @@ set(pngstest_sources
 set(pnggetset_sources
     contrib/libtests/pnggetset.c
 )
+set(pngoverflow_sources
+    contrib/libtests/pngoverflow.c
+)
 set(pngunknown_sources
     contrib/libtests/pngunknown.c
 )
@@ -560,6 +563,15 @@ if(PNG_TESTS AND PNG_SHARED)
 
   png_add_test(NAME pnggetset
                COMMAND pnggetset)
+
+  # pngoverflow test:
+  # Overflow-checking primitives: png_mul_size, PNG_ROWBYTES.
+  add_executable(pngoverflow ${pngoverflow_sources})
+  target_link_libraries(pngoverflow
+                        PRIVATE png_shared)
+
+  png_add_test(NAME pngoverflow
+               COMMAND pngoverflow)
 
   # pngvalid tests:
   # Internal validation of standard and progressive reading,
--- a/Makefile.am
+++ b/Makefile.am
@@ -13,7 +13,7 @@ ACLOCAL_AMFLAGS = -I scripts/autoconf
 
 # test programs - run on make check, make distcheck
 if ENABLE_TESTS
-check_PROGRAMS= pngtest pnggetset pngunknown pngstest pngvalid pngimage pngcp
+check_PROGRAMS= pngtest pnggetset pngunknown pngstest pngvalid pngimage pngoverflow pngcp
 if HAVE_CLOCK_GETTIME
 check_PROGRAMS += timepng
 endif
@@ -59,6 +59,9 @@ pngimage_LDADD = libpng@PNGLIB_MAJOR@@PNGLIB_MINOR@.la
 
 timepng_SOURCES = contrib/libtests/timepng.c
 timepng_LDADD = libpng@PNGLIB_MAJOR@@PNGLIB_MINOR@.la
+
+pngoverflow_SOURCES = contrib/libtests/pngoverflow.c
+pngoverflow_LDADD = libpng@PNGLIB_MAJOR@@PNGLIB_MINOR@.la
 
 pngcp_SOURCES = contrib/tools/pngcp.c
 pngcp_LDADD = libpng@PNGLIB_MAJOR@@PNGLIB_MINOR@.la
@@ -110,7 +113,8 @@ TESTS =\
    tests/pngunknown-save\
    tests/pngunknown-vpAg\
    tests/pngimage-quick\
-   tests/pngimage-full
+   tests/pngimage-full\
+   tests/pngoverflow
 endif
 
 # man pages
--- a/arm/arm_init.c
+++ b/arm/arm_init.c
@@ -145,7 +145,7 @@ png_target_do_expand_palette_neon(png_struct *png_ptr, png_row_info *row_info,
           * way despite the fact that the comments in the neon palette code
           * obfuscate what is happening.
           */
-         png_byte *dp = row + (4/*RGBA*/*row_width - 1);
+         png_byte *dp = row + (4/*RGBA*/*(size_t)row_width - 1);
 
          /* Cosmin Truta: "Sometimes row_info->bit_depth has been changed to 8.
           * In these cases, the palette hasn't been riffled."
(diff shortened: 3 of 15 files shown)


ctruta (commented):
Thumbs-up from me in principle, although I'd like to give it one more read for extra safety, plus a cherry-pick to the `libpng16` branch. (CAUTION: It won't cherry-pick cleanly.)

@ProgramMax, @svgeesus, can you please also give it a look if you have a chance?

ctruta (changes requested):
Thank you, @mohammadmseet-hue. The two-tier design is sound, the coverage is thorough, and the test is solid. I've applied it locally and it passes the full suite.

Three things to address before merging:

- **Merge order with #845.** Both PRs touch `pngpread.c` and `pngrutil.c:png_progressive_read_reset`. Your hunks here are based on the pre-#845 code and keep the `(uInt)` truncation casts that #845 fixes. You'll need a rebase once #845 lands.
- **Missing Makefile.am entry.** The `pngoverflow` test is added to CMakeLists.txt but not to configure/Makefile.am. Both build systems need to stay in sync.
- **Bit-width comments disagree.** pngpriv.h says "36 bits", while `png_rowbytes_checked` says "34 bits". The actual maximum is `(2^32 - 1) * 4 + 7 = 2^34 + 3`, so it's 35 bits.

Minor, not blocking:

- In `png_read_start_row`, `row_bytes` is `size_t` passed to a `png_uint_32` parameter. An explicit cast would silence `-Wconversion`.
- `png_mul_size` evaluates its arguments multiple times. A comment noting the hazard would be an act of kindness to future users.

mohammadmseet-hue:
Thanks for your, as always, thorough and quick review. All points addressed:

- **Bit-width comments:** Both corrected to 35 bits. The full intermediate `(2^32 - 1) * 4 + 7 = 2^34 + 3` needs 35 bits; `pngpriv.h` and `png_rowbytes_checked` now agree.
- **Makefile.am:** Added `pngoverflow` to `check_PROGRAMS`, source/link definitions, and `TESTS`. Also added the `tests/pngoverflow` shell script to match the existing test runner pattern.
- **`-Wconversion` cast:** Added explicit `(png_uint_32)` cast on `row_bytes` in `png_read_start_row`.
- **`png_mul_size` multi-eval:** Added a NOTE comment documenting the hazard.
- **Rebase on #845:** Will rebase once #845 lands.

ctruta (changes requested):
Thank you for the follow-up. The rebase on #845 is still pending, which is fine. We'll handle the rest after #845 lands.

The core machinery is correct. A few things to address before merging:

1. Several sites still use bare `PNG_ROWBYTES` or bare multiplication for `row_info->rowbytes`: `png_do_gray_to_rgb`, `png_do_expand`, `png_do_quantize`, and the user-transform path in `png_do_read_transformations` in pngrtran.c. Similarly, `png_do_pack` in pngwtran.c and `png_do_write_interlace` in pngwutil.c. These functions lack `png_ptr`, so `png_rowbytes_checked` is not available, but they should use `png_mul_size` for consistency. Alternatively, if you consider them safe because of the IHDR width limits, a brief comment at each site documenting that reasoning would suffice.

2. `pngoverflow.c` tests `png_mul_size` overflow with `SIZE_MAX`-range values, which is good, but it does not exercise the `>= 8` path of the updated `PNG_ROWBYTES` macro at all. The `png_uint_32` width constraint means overflow through `PNG_ROWBYTES` is only reachable on 32-bit targets, but a few cases confirming the `>= 8` path produces correct results at large-but-valid widths would still strengthen coverage.

Also, please squash all commits into one and force-push. The PR currently has two commits, and that's fine in general when the commits are logically independent from one another. For direct follow-up and fix-up fixes, however, just amend your top commit and then force-push to the PR.

And, finally: not blocking, but FYI, `pngoverflow.c` does not exercise the `png_rowbytes_checked` error path (the `png_error` on overflow). Testing it requires a `png_struct` with a `setjmp` handler, which is more involved, but it would strengthen the test if you have the appetite for it.

mohammadmseet-hue:
All addressed:

1. **Remaining bare multiplication sites:** Added `png_mul_size` to `png_do_gray_to_rgb`, `png_do_expand` (both gray-alpha and rgba paths), and `png_do_quantize` (both rgb and rgba paths) in pngrtran.c — all sites where pixel_depth is always >= 8 bpp. For the three sites where pixel_depth may be sub-byte (`png_do_pack` in pngwtran.c, `png_do_write_interlace` in pngwutil.c, and the user-transform path in `png_do_read_transformations`), kept `PNG_ROWBYTES` with a comment documenting why they are safe: `PNG_ROWBYTES` uses `png_mul_size` for >= 8 bpp, and sub-byte depths cannot overflow because width is limited by `png_check_IHDR`.

2. **`>= 8` path of `PNG_ROWBYTES` in test:** Added test cases exercising `PNG_ROWBYTES` at `png_uint_32` max-1 widths for 8, 24, 32, and 64 bpp. Uses a runtime `sizeof(size_t)` check (rather than `#if SIZE_MAX`) because `SIZE_MAX` is a constant expression, not a preprocessor integer constant, in many C89 environments. On 64-bit the tests confirm correct values; on 32-bit they confirm `png_mul_size` returns 0 on overflow.

3. **Squashed** into a single commit and force-pushed.

Note on `png_rowbytes_checked` error path: since it is `PNG_INTERNAL` and not exported by the shared library, it cannot be called from an external test binary. On 64-bit targets, the error path is also unreachable (`png_uint_32 * 8` always fits in a 64-bit `size_t`). Added a comment in the test documenting both constraints.

All 38 tests pass (37 existing + pngoverflow).

PR #857, opened 2026-05-07
oss-fuzz: add PNG v3 chunks to fuzzing dictionary

dhruvac29:
This updates the OSS-Fuzz PNG dictionary with PNG v3 chunk names that libpng already supports: cICP, cLLI, and mDCV.

Adding these tokens helps fuzzers generate inputs that exercise the existing PNG v3 chunk parsing paths more directly.

Validation:
- git diff --check HEAD~1..HEAD passes

I attempted a full OSS-Fuzz build locally, but the build failed before reaching libpng because the OSS-Fuzz base-builder image failed to run Perl/debconf in my local Docker environment.


ctruta:
Looks good, thank you. The three names are spelled and case-correct against the PNG v3 registrations, sorted into the right positions, and they match chunks libpng already parses on this branch, so they'll genuinely steer the fuzzer toward live code. Ready to merge.

ctruta:
One licensing heads-up: I'm contemplating a split of libpng's OSS-Fuzz integration into its own project, and a relicensing under the Apache License 2.0 (which is where it originated). This is still tentative, with no timeline.

Would you be willing to have your contribution here relicensed under Apache-2.0 if that happens?

Issue #783, opened 2026-01-19
CMAKE: WIN32: Wrong symlink created for import library


carlo-bramini:
I tried to update the libpng packages for CYGWIN and the MinGW-w64 cross compilers and I noticed something wrong.
Into the `/lib` directory, it is created a symlink to the `libpng16.dll.a` import library, but the extension is not correct:

<img width="669" height="239" alt="Image" src="https://github.com/user-attachments/assets/a23f0f23-8e05-47be-9b78-fae854b48537" />

As you can see, there is a `libpng.dll` instead of `libpng.dll.a`.
The error into `CMakeLists.txt` is located here:
https://github.com/pnggroup/libpng/blob/0f07f70aa55e4771d00b0f8a238dca992fbb60e4/CMakeLists.txt#L817-L821
Actually, `CMAKE_SHARED_LIBRARY_SUFFIX` is equal to `.dll` and not to `.dll.a`.
This is the reason because the wrong name is assigned to the symlink.
The right variable to use here should be `CMAKE_IMPORT_LIBRARY_SUFFIX`.
Technically, it should be a very easy fix, at least in theory.
However, since there is this `if()` statement:
https://github.com/pnggroup/libpng/blob/0f07f70aa55e4771d00b0f8a238dca992fbb60e4/CMakeLists.txt#L817
it is possible that this code is also executed for something different that `CYGWIN` and `MINGW` because the `NOT WIN32` condition.
So, I'm not sure how to fix it because I don't understand what you wanted to do exactly.
If you could give me some advices, I will test it and I can contribute a PR with the fix if you want.

PR #859, opened 2026-05-10
CMAKE: Fix wrong symlink created for import library on MinGW/CYGWIN

carlo-bramini:
This PR fixes issue #783.

If the target platform is `MINGW` or `CYGWIN`, then the name of the target of the symlink is built with `CMAKE_IMPORT_LIBRARY_SUFFIX`.
Otherwise, it uses `CMAKE_SHARED_LIBRARY_SUFFIX` for replicating the same behaviour of the code and not breaking current status.

This patch also needs to be backported to libpng16 because the support for MinGW and CYGWIN with CMake is broken because of this issue.


ctruta:
Approved. Thank you very much.

One minor nit: the outer comment needs fixing, too. It was stale, even before your fix. (I fixed it now.)

ctruta:
Integrated in both `libpng16` and `libpng18`. Thanks again.

PR #866, opened 2026-05-26
Add missing length check to png_handle_tIME; remove dead PNG_UNUSED(length)

### LINKED ISSUE #826: png_handle_tIME: missing length check
`png_handle_tIME` calls `png_crc_read(png_ptr, buf, 7)` without first checking `length == 7`. Every other fixed-size chunk handler in `pngrutil.c` guards against wrong lengths before reading — `png_handle_tRNS`, `png_handle_bKGD`, `png_handle_gAMA`, etc. all do:

```c
if (length != expected)
{
    png_chunk_benign_error(png_ptr, "invalid");
    return handled_error;
}
```

`png_handle_tIME` does neither. There's also dead code: `PNG_UNUSED(length)` at line 2369 appears after `return handled_ok` and is never reached.

The `CDtIME 7U, 7U` entry in `read_chunks[]` (line 3250) sets `min_length = max_length = 7`, so the dispatch layer (line 3335) rejects wrong-length tIME chunks before the handler is called — no practical impact. But the handler should still validate its own input rather than rely on that.

**Fix**: add the length guard after the debug trace and remove the dead `PNG_UNUSED(length)`.

Confirmed on `libpng18` at `28cb99fe6`.

XananasX7:
## Summary

Add the missing length guard to `png_handle_tIME()` in `pngrutil.c`, and remove the dead `PNG_UNUSED(length)` statement that follows an unconditional `return`.

## Problem

Every other fixed-size chunk handler in `pngrutil.c` validates the chunk length before reading data, for example `png_handle_tRNS`, `png_handle_bKGD`, `png_handle_gAMA`:

```c
if (length != expected)
{
   png_chunk_benign_error(png_ptr, "invalid");
   return handled_error;
}
```

`png_handle_tIME` is the only fixed-size handler that skips this check and calls `png_crc_read(png_ptr, buf, 7)` unconditionally. The dispatch layer (`CDtIME 7U, 7U`) currently rejects wrong-length chunks before reaching the handler, so there is no direct exploitable path — but relying solely on the dispatch layer for safety is fragile and inconsistent.

Additionally, `PNG_UNUSED(length)` appears after `return handled_ok` at the end of the function body, making it dead code.

## Fix

1. Add the explicit length check after the debug trace, matching every other handler.
2. Remove the dead `PNG_UNUSED(length)` macro invocation.

Fixes #826.

### DIFF
--- a/pngrutil.c
+++ b/pngrutil.c
@@ -955,7 +955,6 @@ png_handle_IHDR(png_struct *png_ptr, png_info *info_ptr, png_uint_32 length)
        color_type, interlace_type, compression_type, filter_type);
 
    return handled_ok;
-   PNG_UNUSED(length)
 }
 
 /* Read and check the palette */
@@ -1116,7 +1115,6 @@ png_handle_gAMA(png_struct *png_ptr, png_info *info_ptr, png_uint_32 length)
 #endif /*READ_GAMMA*/
 
    return handled_ok;
-   PNG_UNUSED(length)
 }
 #else
 #  define png_handle_gAMA NULL
@@ -1257,7 +1255,6 @@ png_handle_cHRM(png_struct *png_ptr, png_info *info_ptr, png_uint_32 length)
 #  endif /* READ_RGB_TO_GRAY */
 
    return handled_ok;
-   PNG_UNUSED(length)
 }
 #else
 #  define png_handle_cHRM NULL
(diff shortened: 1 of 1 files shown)


XananasX7:
Friendly ping - this adds the missing length check to `png_handle_tIME`. Happy to revise if the maintainers have any feedback.

PR #870, opened 2026-05-29
Fix double-free of struct member pointers on allocation failure

jortles:
## Summary

Four double-free vulnerabilities caused by `png_free()` not NULLing struct member pointers before a subsequent call that can `longjmp` via `png_error()`. If the allocation fails, the cleanup function `png_read_destroy()` frees the dangling pointer again.

### Fix 1: `big_row_buf` + `big_prev_row` in `png_read_start_row()` (pngrutil.c)

`png_read_destroy()` line 856 frees the dangling pointer.  Reachable from any PNG read path.

### Fix 2: `palette` in `png_set_PLTE()` (pngset.c)

`png_read_destroy()` line 873 frees the dangling pointer.  Reachable from untrusted PNG input via `png_handle_PLTE()`.

### Fix 3: `trans_alpha` in `png_set_tRNS()` (pngset.c)

`png_read_destroy()` line 881 frees the dangling pointer.  Reachable from untrusted PNG input via `png_handle_tRNS()`.  The correct pattern already exists at line 1231 of the same function, where `trans_alpha` IS NULLed after free.

### Fix 4: `quantize_index` in `png_set_quantize()` (pngrtran.c)

`png_read_destroy()` line 866 frees the dangling pointer.  Reachable when applications call `png_set_quantize()` (code at line 494 explicitly says repeated calls are allowed).

### Root cause

libpng uses `setjmp`/`longjmp` for error handling.  When `png_malloc()` fails, it calls `png_error()` which performs a `longjmp` to the application error handler.  This skips the pointer assignment that would have overwritten the freed address.  The cleanup function `png_read_destroy()` then frees the still-stored (now dangling) pointer — a double-free.

### Impact

- **Bug class:** CWE-415 (Double Free)
- **Trigger:** Crafted PNG combined with memory pressure causing `png_malloc` to fail
- **Fixes 1–3 reachable from:** Untrusted PNG input (read path)
- **Fix 4 reachable from:** Application API (`png_set_quantize()`)

### Testing

Library builds cleanly with `-Wall -Werror -fsanitize=address,undefined`.  pngtest passes.

### DIFF
--- a/pngrtran.c
+++ b/pngrtran.c
@@ -495,6 +495,7 @@ png_set_quantize(png_struct *png_ptr, png_color *palette,
        * this function more than once per png_struct.
        */
       png_free(png_ptr, png_ptr->quantize_index);
+      png_ptr->quantize_index = NULL;
       png_ptr->quantize_index = (png_byte *)png_malloc(png_ptr,
           PNG_MAX_PALETTE_LENGTH);
       for (i = 0; i < PNG_MAX_PALETTE_LENGTH; i++)
--- a/pngrutil.c
+++ b/pngrutil.c
@@ -4789,7 +4789,9 @@ defined(PNG_USER_TRANSFORM_PTR_SUPPORTED)
    if (row_bytes + 48 > png_ptr->old_big_row_buf_size)
    {
       png_free(png_ptr, png_ptr->big_row_buf);
+      png_ptr->big_row_buf = NULL;
       png_free(png_ptr, png_ptr->big_prev_row);
+      png_ptr->big_prev_row = NULL;
 
       if (png_ptr->interlaced != 0)
          png_ptr->big_row_buf = (png_byte *)png_calloc(png_ptr,
--- a/pngset.c
+++ b/pngset.c
@@ -809,6 +809,7 @@ png_set_PLTE(png_struct *png_ptr, png_info *info_ptr,
     * regardless of num_palette.
     */
    png_free(png_ptr, png_ptr->palette);
+   png_ptr->palette = NULL;
    png_ptr->palette = png_voidcast(png_color *, png_calloc(png_ptr,
        PNG_MAX_PALETTE_LENGTH * (sizeof (png_color))));
    info_ptr->palette = png_voidcast(png_colorp, png_calloc(png_ptr,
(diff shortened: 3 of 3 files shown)


jortles:
Added a second commit applying the same fix to three additional sites where the same bug pattern exists:

**pngset.c — `png_set_PLTE()`** (line 812):  
`png_ptr->palette` is freed, then `png_calloc()` can `longjmp` on OOM. `png_read_destroy()` line 873 frees the dangling pointer again → double-free.  Reachable from untrusted PNG input via `png_handle_PLTE()`.

**pngset.c — `png_set_tRNS()`** (line 1224):  
`png_ptr->trans_alpha` is freed, then `png_malloc()` can `longjmp` on OOM. `png_read_destroy()` line 881 frees the dangling pointer again → double-free.  Reachable from untrusted PNG input via `png_handle_tRNS()`.

**pngrtran.c — `png_set_quantize()`** (line 498):  
`png_ptr->quantize_index` is freed, then `png_malloc()` can `longjmp` on OOM. `png_read_destroy()` line 866 frees the dangling pointer again → double-free.

All three follow the identical pattern as the first commit: `png_free()` without NULLing, followed by a call that can `longjmp`. The fix is the same: insert `ptr = NULL;` between the free and the allocation.

The correct pattern already exists in the same file at `png_set_tRNS()` line 1231–1232, where `png_ptr->trans_alpha` IS NULLed after free.

Total patch: 5 insertions across 3 files (pngrutil.c, pngset.c, pngrtran.c).

ctruta:
Thank you, Anthony. You caught a real defect (class CWE-415). Your fix is correct, and it follows the idiom the library already uses in `png_read_buffer` and the unknown-chunk handler.

There's one correction to the framing, though. The description labels the row-buffer, `palette`, and `trans_alpha` sites as reachable from untrusted PNG input. None of them is:

- A double-free needs the member to be non-NULL before the free, hence a prior successful pass through the same site. On the first pass, the member is NULL, `png_free` is a no-op, and an OOM leaves it NULL, so the cleanup free is also a no-op.
- A second PLTE is rejected as a duplicate inside `png_handle_PLTE` before `png_set_PLTE` runs.
- A second tRNS is rejected as a duplicate by the chunk dispatcher before `png_handle_tRNS` runs at all. In the progressive (push) path, too.
- APNG does re-run `png_read_start_row` on every frame, but fcTL validation caps each frame to the first frame's dimensions, so the row-buffer reallocation guard never re-fires within one decode.
- `png_set_quantize` is an application transform, not driven by any chunk.

So a malicious PNG can't arm any of the four sites. The real trigger is an application that causes a setter to run twice on the same `png_ptr` (or reuses the struct across decodes) **and** hits an allocation failure at that exact point: a robustness defect worth fixing, not a remote-content vulnerability. Accordingly, I'm merging this as a hardening fix, without filing for a CVE ID.

Thanks again for your contribution :-)

ctruta:
Integrated in branch `libpng18` and cherry-pick in `libpng16`. Thanks again.

PR #874, opened 2026-05-31
pngrutil: add missing length checks to gAMA, sRGB, cHRM, pHYs, oFFs handlers

XananasX7:
## Summary

Five chunk handlers in `pngrutil.c` read a fixed number of bytes via `png_crc_read()` without first verifying that the chunk's declared `length` matches the expected size. This is the same class of defect fixed for `tIME` in PR #866; this PR applies the same pattern to the remaining affected handlers.

## Affected handlers

| Handler | Expected length | Risk |
|---------|----------------|------|
| `png_handle_gAMA`  | 4 bytes  | Reads 4B from potentially shorter chunk; UB in `png_crc_finish` arithmetic |
| `png_handle_sRGB`  | 1 byte   | Same |
| `png_handle_cHRM`  | 32 bytes | Same — largest unprotected fixed-size read |
| `png_handle_pHYs`  | 9 bytes  | Same |
| `png_handle_oFFs`  | 9 bytes  | Same |

## Root cause

If a malformed PNG declares a chunk length shorter than the expected fixed size, `png_crc_read()` reads fewer bytes than needed, leaving the local buffer partially uninitialized or consuming bytes from the next chunk's data. When `png_crc_finish(png_ptr, 0)` is then called, the internal `idat_size` arithmetic underflows (unsigned wrap), producing undefined-behaviour territory and a CRC mismatch result that is non-deterministic across platforms.

## Fix

Each handler now checks `length != <expected>` as its **first** action, calls `png_crc_finish(png_ptr, length)` to drain the CRC bytes cleanly, emits a benign chunk error, and returns `handled_error` — exactly the pattern introduced for `tIME` in PR #866 and `eXIf` before it.

The two now-unreachable `PNG_UNUSED(length)` macros in `pHYs` and `oFFs` are also removed.

## Testing

Manually crafted PNGs with truncated gAMA/sRGB/cHRM/pHYs/oFFs chunks (length set to 0, 1, expected−1) parsed cleanly with the new guards: benign-error logged, chunk skipped, remainder of the file processed normally. No regressions in the existing test suite (`make check`).


XananasX7:
Friendly ping - this adds missing length checks to the gAMA, sRGB, cHRM, pHYs, and oFFs handlers. Happy to revise if needed.

PR #878, opened 2026-06-11
ossfuzz: add png_write_chunks_fuzzer for pngwutil encode path

XananasX7:
## Summary

The existing OSS-Fuzz libpng harnesses cover only the **decode** path (`pngrutil.c`). The **encode/write path** — `pngwutil.c`, `pngwrite.c`, filter selection, and compression — has no dedicated fuzz coverage.

## New harness: `contrib/ossfuzz/png_write_chunks_fuzzer.c`

Derives image parameters (width 1–64, height 1–64, color type RGB/RGBA, bit depth 8/16, filter type, compression level) from the first 6 fuzz bytes, then writes a full PNG through:

- `png_write_IHDR` / `png_write_info`
- `png_write_image` row loop (all filter strategies)
- `png_write_IEND`
- Optional `png_set_gAMA` metadata chunk

Uses a custom in-memory write callback (no file I/O), covering `pngwio.c`, `pngwrite.c`, and the per-filter path in `pngwutil.c`.

### DIFF
--- a/contrib/ossfuzz/png_write_chunks_fuzzer.c
+++ b/contrib/ossfuzz/png_write_chunks_fuzzer.c
@@ -1,0 +1,82 @@
+/* png_write_chunks_fuzzer.c
+ *
+ * Fuzz target for the libpng write path (chunk serialisers in pngwutil.c).
+ * The existing OSS-Fuzz harnesses cover read/decode only (pngrutil.c).
+ * This harness constructs a minimal RGBA image from fuzz input and
+ * writes it through the full encode pipeline, exercising:
+ *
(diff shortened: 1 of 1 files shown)


XananasX7:
Friendly ping — happy to address any review comments. Thank you!

ctruta:
Thank you, @XananasX7,

Migrating the fuzz testing of newer libpng versions into a dedicated repository has been part of our plans for a while, which is why the harnesses were removed from the `libpng18` branch.

About your contribution: yes, a write-path target like this is a welcome addition. I'll close this PR for now, and I'll ping you again right here after I get the new repository `pnggroup/pngfuzz` up and running.

PR #879, opened 2026-06-11
Harden GitHub Actions workflow permissions

Alb3e3:
This tightens the GitHub Actions workflow security posture without changing the lint or verification jobs.

- set the workflow `GITHUB_TOKEN` permissions to read-only contents access
- avoid direct GitHub context interpolation in the Linux dependency-install shell condition
- pass the ref values through environment variables and quote them before use

I verified the workflow YAML still parses, `git diff --check` passes, and `zizmor` no longer reports the previous excessive-permissions or template-injection findings. The remaining findings are action pinning warnings, which I left out of this focused change.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,6 +13,9 @@ on:
       - libpng16
       - libpng18
 
+permissions:
+  contents: read
+
 jobs:
   lint:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


ctruta:
GitHub shouldn't show this as "closed with unmerged commits". I just merged it.

Thanks for the contribution :-)

ctruta:
Following up: since all autotools packages are now available in all Ubuntu runner images, that part with replacing context expressions with environment variables is no longer needed. See commit 964b4135949703b705fc760fc3fb546b86e5ab47.

PR #885, opened 2026-06-25
fix locale-dependent sCAL value parsing in png_get_sCAL

Samin061:
png_get_sCAL and png_get_sCAL_fixed convert the sCAL chunk's two ASCII numbers with atof(), which honors LC_NUMERIC. Under a locale whose decimal separator is not '.', atof() stops at the dot, so the scale read from an untrusted PNG silently loses its fractional and exponent parts (for example "3.14" becomes 3.0 and "0.5e1" becomes 0.0). The stored strings are already validated as spec-form ASCII floats by png_check_fp_number/png_check_fp_string, so a small locale-independent parser over that validated text gives the correct value regardless of the caller's locale. The conversion stays in the getter since that is where the locale leaks in.

### DIFF
--- a/pngget.c
+++ b/pngget.c
@@ -1017,6 +1017,100 @@ png_get_pCAL(const png_struct *png_ptr, png_info *info_ptr,
 #endif
 
 #ifdef PNG_sCAL_SUPPORTED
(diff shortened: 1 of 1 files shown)


Samin061:
gentle ping

PR #887, opened 2026-06-30
Add null check before closing file pointer

hgqxjj:
Please review this change. Thanks!

**Description:**

Avoid calling `fclose()` with a `NULL` file pointer.

`fopen()` may fail and return `NULL`, so check `ip` before calling `fclose(ip)`.

### DIFF
--- a/contrib/tools/checksum-icc.c
+++ b/contrib/tools/checksum-icc.c
@@ -81,7 +81,8 @@ int main(int argc, char **argv)
             printf("/* ERROR: %s */\n", argv[i]);
          }
 
-         (void)fclose(ip);
+         if (ip != NULL)
+            (void)fclose(ip);
       }
    }
 


jbowler:
Why?

hgqxjj:
Hi @jbowler, Because `fopen()` may fail and return `NULL`. In that case, calling `fclose(ip)` would effectively call `fclose(NULL)`, which is undefined behavior and may cause a crash. The `ip != NULL` check ensures we only close a successfully opened file.

jbowler:
Your pull request says:

_No description provided._

That's why I asked, "Why?"

hgqxjj:
Thanks, I misunderstood your question. I've added the rationale to the PR description.

jbowler:
Reasonable; the prior code would give a user who passed in a bogus file name a nasty rejoinder and stop listening to what he said (the failure is in a loop over the arguments).

Neither is satisfactory of course but the suggested change is reasonable.

hgqxjj:
Thanks for the review and for the feedback!

PR #888, opened 2026-07-02
fix display row pointer declaration in readpng and timepng

Samin061:
In read_png (contrib/libtests/readpng.c) and read_by_row (contrib/libtests/timepng.c) the display row is declared with `png_byte *row = NULL, display = NULL;`, so the `*` binds only to `row` and `display` ends up a single byte. The malloc result is truncated into that byte, png_read_row then writes the decoded row through the bogus pointer and free() releases it, and recent compilers reject both files with -Wint-conversion. Add the missing `*` so display is a png_bytep like row.


Samin061:
any update?

PR #889, opened 2026-07-10
fix broken image-size overflow guard in gregbook readers

Samin061:
The gregbook example readers guard the image_data allocation with a check written in size_t, but the malloc right below computes rowbytes*height with 32-bit operands, so a crafted image whose dimensions make that product wrap (for example 32768x32768 RGBA, exactly 2^32, well inside the default 1,000,000 limits) slips past the guard and undersizes the buffer that png_read_image then overruns. In readpng.c, readppm.c and rpng2-win.c the guard also has a stray comma after the fprintf that turns return into an argument to it, so those three files do not compile at all. Terminate the fprintf and compute the size in size_t so the existing guard actually covers the allocation.


Samin061:
any update?

jbowler:
Again, https://github.com/pnggroup/libpng/pull/897

Samin061:
Fair enough, if the gregbook readers are unmaintained and EOL there's no point patching them. Closing this out.

PR #890, opened 2026-07-11
check width x channels overflow in pnm2png row_bytes

Samin061:
do_pnm2png computes row_bytes as width * channels * depth in 32-bit, so a PNM header with a large width (a 0x55555556-wide P6 wraps the product down to 2) slips past the height size guard and the following per-row png_malloc is far too small, after which the pixel loop writes width*channels bytes into it and overruns the heap. Reject widths whose width x channels x depth product would overflow before the multiply, in both the packed and unpacked branches.


Samin061:
any update?

jbowler:
Same as https://github.com/pnggroup/libpng/pull/897

Unmaintained code.  EOL.

Samin061:
Makes sense. If contrib/pngminus is treated as EOL then there's no point carrying this fix upstream. No objection to closing it out.

PR #895, opened 2026-07-13
free row_buf/prev_row on frame re-entry in png_write_start_row

Samin061:
png_write_start_row allocates row_buf and prev_row every time it runs, and the APNG writer re-enters it once per frame through png_write_reset, so each frame after the first overwrites the previous frame's pointers and leaks those buffers until png_destroy_write_struct. Free and clear both members before reallocating, the way png_read_start_row already does, which also stops a failed re-allocation on re-entry from leaving a stale pointer that gets freed twice. A four-frame APNG leaks three row buffers before this change and none after.


lovell:
If you hadn't already seen, there's a similar PR to fix the leak *and* the possible heap buffer write overflow at https://github.com/pnggroup/libpng/pull/886

Samin061:
Hadn't seen it, good pointer. #886 covers strictly more than this one.

I'd only looked at the leak. try_row/tst_row are allocated only when they're NULL, so I wrote them off as unaffected, but that NULL guard is exactly what lets a stale buffer from a smaller frame get reused for a larger one, which is the overflow. And png_write_reset is the better place for it anyway since it's the only frame transition on the write side.

Closing in favour of #886.

PR #897, opened 2026-07-22
pngfix: avoid passing NULL prefix to fprintf when output name too long

### LINKED ISSUE #896: Potential Null Pointer Dereference
When looking at the source code, I noticed that https://github.com/pnggroup/libpng/blob/d1d0abeffede1cc898ddc3d0e600839cf026d749/contrib/tools/pngfix.c#L3936 doesn't check whether `prefix==NULL` like line 3944. Does it mean that at line 3936, `prefix!=NULL`? If `prefix` can be null, then the `fprintf` may cause undefined behaviors.

SAY-5:
`prefix` starts as `NULL` and is only set when `--prefix=` is given. In the "output file name too long" branch that runs before the `if (prefix != NULL)` block, it is still passed straight to `fprintf` with `%s`, which is undefined behavior when no prefix was supplied. The matching message a few lines down is inside the `prefix != NULL` guard, so only this one is exposed.

Guard it the same way `suffix` already is on that line. Fixes #896.


jbowler:
I think that's correct.  It wasn't in the original and @thealberto authored that so they need to comment.

PR #904, opened 2026-07-29
pngset: preserve pCAL state on allocation failure

### LINKED ISSUE #917: png_set_pCAL allocation failure can cause an out-of-bounds read during destroy
# `png_set_pCAL` allocation failure can cause an out-of-bounds read during destroy

## Summary

Replacing an existing pCAL value can leave `png_info` with the new
`pcal_nparams` count but the old, shorter `pcal_params` allocation when a later
replacement allocation fails. The application may correctly stop processing
after the allocation warning and call only `png_destroy_write_struct()`, but
cleanup then indexes the old array using the new count and reads beyond it.

This report is limited to the allocator-failure API path. It does not claim
that PNG stream contents directly trigger the condition.

## Reproduction

Tested against commit `d1d0abeffede1cc898ddc3d0e600839cf026d749` on the
`libpng18` branch with AppleClang AddressSanitizer and UndefinedBehaviorSanitizer.

1. Create a write struct with `png_create_write_struct_2()` and a deterministic
   custom allocator.
2. Set a valid pCAL value with one parameter.
3. Replace it with a pCAL value containing three parameters.
4. Make the allocator reject the replacement's second allocation, after the
   replacement count has been stored.
5. Perform no further parsing or setter calls; call only
   `png_destroy_write_struct()`.

The standalone reproducer used for verification is named
`pcal_oom_destroy_poc.c`. Its key sequence is:

```c
png_set_pCAL(png_ptr, info_ptr, "first", 10, 20,
    PNG_EQUATION_LINEAR, 1, "old-unit", first_params);

allocator.fail_at = allocator.calls + 2;
png_set_pCAL(png_ptr, info_ptr, "second", 30, 40,
    PNG_EQUATION_LINEAR, 3, "new-unit", second_params);

png_destroy_write_struct(&png_ptr, &info_ptr);
```

## Actual result

The unpatched build aborts with:

```text
ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 8
#0 png_free_data png.c:551
#1 png_destroy_info_struct png.c:418
#2 png_destroy_write_struct pngwrite.c:1057
#3 main pcal_oom_destroy_poc.c:75

0 bytes after 16-byte region
allocated by png_set_pCAL at pngset.c:572
```

The 16-byte allocation contains the original one parameter plus its null
terminator. Cleanup walks it using the failed replacement's count of three.

## Expected result

An allocation failure while replacing pCAL data should leave the prior value
internally consistent so the required destroy operation remains safe.

## Proposed fix

PR #904 builds each replacement in temporary storage and installs it only after
all allocations succeed. The same reproducer exits with `destroy completed`
and no ASan/UBSan finding on that branch. The PR also adds deterministic
allocator-failure regressions for pCAL, hIST, and sCAL replacement behavior.

Assisted-by: OpenAI Codex:gpt-5.6-sol

carrerasdarren-cell:
## Summary

- build replacement pCAL fields in temporary storage and install them only after every allocation succeeds
- preserve the prior pCAL value when a replacement allocation fails
- add a deterministic failing-allocator regression
- narrow the PR to the focused cleanup bug in #917 following maintainer feedback

## Root cause

`png_set_pCAL` wrote replacement fields directly into a live `png_info`. In particular, it installed the new parameter count before every replacement allocation succeeded. If the units allocation then failed, the old, shorter parameter array remained paired with the new count. An application that stopped after the allocation warning and called only `png_destroy_write_struct()` could make `png_free_data()` read beyond the old array during required cleanup.

The change treats replacement as a transaction: allocate and copy the complete new value first, clean up temporary allocations on failure, then free and replace the old value only after success. This also keeps self-aliasing replacement calls safe.

This is limited to the allocator-failure API path; it does not claim direct triggering from PNG stream contents.

## Validation

- untouched `d1d0abe` negative control: standalone ASan/UBSan reproducer reports a heap-buffer-overflow in `png_free_data()` during `png_destroy_write_struct()`
- patched standalone reproducer: exits with `destroy completed` and no sanitizer finding
- focused warnings-as-errors ASan/UBSan `pnggetset`: 1/1 pass
- full ASan/UBSan CTest suite: 37/37 pass
- `git diff --check`: pass

Fixes #917

Assisted-by: OpenAI Codex:gpt-5.6-sol


jbowler:
From your comments.  Remember that when the app gets an error callback the parse of the stream must be aborted.  A damaged stream permits no false assumptions; the only allowed operations are the various "free" or "destroy" operations and, indeed, those are required.

png_set_hIST: correct behavior.
png_set_pCAL: admissible behavior.  The "free" operation does not require a byte count.
png_set_sCAL: likewise.

Submit a bug report which shows that the value passed to ANSI-C free(3) is not the correct value.

carrerasdarren-cell:
I reproduced the incorrect cleanup access requested above on the unpatched base (`d1d0abe`) with ASan.

Minimal sequence:

1. `png_set_pCAL(..., nparams=1, ...)` succeeds.
2. A custom allocator rejects the second allocation in a replacement `png_set_pCAL(..., nparams=3, ...)` call.
3. The unpatched setter has already stored `pcal_nparams = 3`, but returns before replacing `pcal_params`, which still points to the old two-slot allocation (`params[0]` plus its NULL terminator).
4. The application calls only the allowed cleanup operation, `png_destroy_write_struct()`.

ASan then reports:

```
ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 8
#0 png_free_data png.c:551
#1 png_destroy_info_struct png.c:418
#2 png_destroy_write_struct pngwrite.c:1057
#3 main pcal_oom_destroy_poc.c:74

0x602000000140 is 0 bytes after 16-byte region
allocated by png_set_pCAL at pngset.c:572
```

So the loop at `png.c:550-551` indexes the old one-parameter array using the new count of three. It reads past the allocation while obtaining the value to pass to `png_free`; ASan stops at that out-of-bounds read before an invalid value can reach ANSI C `free(3)`.

The identical reproducer linked against this PR exits 0 with `destroy completed` and no ASan/UBSan finding. This is independent of continuing a stream parse: the only operation after the allocation warning is destruction.

jbowler:
I think that's a more specific bug, which I would say you should raise first as a bug report then, perhaps after discussion with the maintainers (last I looked there were four, none me) as a fix.

BTW I am being very hardnosed here; I think the code deserves a good rewrite, but we live in a time where Russian and Chinese contributors really will happily rewrite our badly written code to the ends of their own State.  They might fix a few bugs too.  Think about it.

carrerasdarren-cell:
Opened #917 as requested with the focused pCAL cleanup bug and a sanitizer reproduction. I also re-ran the exact negative control today: the unpatched `d1d0abe` build reports an ASan heap-buffer-overflow in `png_free_data()` during `png_destroy_write_struct()`, while this PR exits cleanly with no ASan/UBSan finding. The issue explicitly limits the claim to the allocator-failure API path.

carrerasdarren-cell:
I narrowed this PR to the specific pCAL cleanup bug reported in #917, removing the hIST and sCAL changes. The updated branch retains the sanitizer-backed before/after proof and adds only the focused pCAL allocation-failure regression. The full ASan/UBSan CTest suite passes 37/37.

PR #913, opened 2026-08-26
fix heap overflow in simpleover sprite_op

Samin061:
sprite_op composites each input PNG by looping over the decoded image width and height, but it writes into sprite->buffer, which is sized from the --sprite dimensions given on the command line. The only bounds check validates the x and y offsets, not the image extent, so a sprite PNG whose IHDR is larger than the sprite runs the loops past the buffer and overflows the heap with pixel data from the file. Stop both loops at the sprite's right and bottom edges so the input is truncated to the sprite area, which is the behavior the usage text already documents.


jbowler:
It's sample code.  I can't convince myself that any of the function in question is correct apart from the bit I meant it to be a sample for.  Making it more complicated does not help.

The sprite and the image are represented as rectangles.  It's a simple rectangle-rectangle intersection, but that was not the point of the sample code.  Sure, it's a useful piece of code and it does make a second point; how to handle sprites that overlap, or are beyond, the edge of the image.

It's not the function of this piece of code.  It could be added but maybe the file should be renamed "complexover".

PR #914, opened 2026-08-27
fix(pngimage): correct 16-bit sBIT handling in compare_read

Samin061:
compare_read fills sig_bits with one entry per colour component, but for 16-bit images the sBIT sanity check loops once per byte and the case 16 expansion writes sig_bits[2*b+1]/[2*b+0] starting from b == channels. Feeding pngimage a 16-bit RGBA file that carries an sBIT chunk drives b to 4, storing two bytes past the 8-byte sig_bits stack array; AddressSanitizer reports a stack-buffer-overflow at the write. Bound the check by the real component count and expand into sig_bits[2*b-1]/[2*b-2] from sig_bits[b-1] so the eight bytes are produced within the array.


jbowler:
The change on line 1162 makes no sense whatsoever; 'b' is incremented.

The change is also isolated from the rest of the changes; it is located within a separate loop.

The other change implies an off-by-one error but the explanation is utterly unrelated.  This looks to me like something generated by an AI.  I'm not paid to train AIs so I will not comment further.

PR #916, opened 2026-08-30
fix: Avoid NULL memcpy in png_set_text

samuelgirmametaferia:
## Summary

Avoid calling `memcpy()` with a NULL source pointer when `png_set_text()` receives an iTXt entry with an empty language or translated keyword.

`png_text` explicitly permits `lang` and `lang_key` to be NULL. The existing code computes a zero length for those fields but still passes the NULL pointer to `memcpy()`, which is undefined behavior even when the copy length is zero and is diagnosed by UBSan.

Guard the two copies on their computed lengths and extend `pnggetset` so the documented NULL iTXt language pointers are exercised.

## Validation

- `git diff --check`: pass
- CMake build: pass
- Full CTest suite: 37/37 pass
- Pristine `libpng18` with only the regression trigger: UBSan fails at `pngset.c:1108` with `null pointer passed as argument 2`
- Patched branch under the same UBSan configuration: `pnggetset` passes


samuelgirmametaferia:
Hey could you please review this!

PR #922, opened 2026-09-10
ci: restrict workflow token to read-only contents

Alb3e3:
## Summary

Set the default `GITHUB_TOKEN` permission for the CMake workflow to `contents: read`.

The workflow checks out the repository and runs the test script; it does not need repository write access. An explicit read-only default limits the impact of a compromised action or pull request workflow.

This is a Google Patch Rewards OSS-Fuzz Tier 2 hardening patch. AI assistance was used and is disclosed.

### DIFF
--- a/.github/workflows/cmake.yaml
+++ b/.github/workflows/cmake.yaml
@@ -1,0 +1,38 @@
+name: Development
+
+on: [push, pull_request]
+
+jobs:
+  test:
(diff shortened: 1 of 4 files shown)


Alb3e3:
Closing this. I opened it against libpng18 from the fork main history, so the PR includes unrelated old commits and the CMake workflow does not exist on the target branch. The active libpng18 workflows are already covered by #879.

PR #926, opened 2026-09-14
Support building for multiple archs in a single CMake call

ISSOtm:
Currently, libpng really only scans the first element in a list; this remains relevant even when in a non-SIMD build, since only the first listed architecture will have its SIMD disabled.

The diff looks huge, but it's mostly just indentation&mdash;enable ignoring whitespace and it should (hopefully) become far more palatable... also, I am unable to test it right now due to lacking a Mac to build something multi-arch on, and I will have to expend even more effort to make GitHub Actions use this branch to attempt to build the project I am trying to integrate this into (gbdev/rgbds#2083) and it's really late :P But I'll report back as soon as I can get some results&mdash;preferably some without `HARDWARE_OPTIMIZATIONS` like presently, but with any luck, some with also? :)


ISSOtm:
Turns out hardware optimisations are [not compatible](https://github.com/gbdev/rgbds/blob/cc2ff925296d4d160db9f9735319ebb96febdc78/cmake/macos-static.cmake#L15-L16) with a multi-arch build *at least in the current setup*, so no testing of that for now. I *am* vaguely feeling that it might be possible to support them by splitting the SIMD code into separate `OBJECT` targets that get specific `OSX_ARCHITECTURES` properties (akin [to this](https://github.com/gbdev/rgbds/blob/c8e8a2f52256d96a2e261b1bc23228059f2d483e/src/CMakeLists.txt#L3-L12))...

...but I don't have a Mac to attempt to develop this on, and I wouldn't benefit much from that personally, so I'm going to turn back and attempt to climb back from this rabbit hole that I've dug myself into :grin: This PR solves a problem I'm having, so let's keep it at that for now.

jbowler:
> Turns out hardware optimisations are [not compatible](https://github.com/gbdev/rgbds/blob/cc2ff925296d4d160db9f9735319ebb96febdc78/cmake/macos-static.cmake#L15-L16) with a multi-arch build _at least in the current setup_, so no testing of that for now. I _am_ vaguely feeling that it might be possible to support them by splitting the SIMD code into separate `OBJECT` targets that get specific `OSX_ARCHITECTURES` properties (akin [to this](https://github.com/gbdev/rgbds/blob/c8e8a2f52256d96a2e261b1bc23228059f2d483e/src/CMakeLists.txt#L3-L12))...

This was broken by the palette related code added by ARM developers.  I fixed it in libpng18 by making a hard rule that hardware stuff goes in the hardware directories and only there.  That stuff is probably broken again by now.

Originally it was possible to do what you are trying to do and the only problem was CMakeLists.txt.  Multi-arch worked just fine (before the ARM change) with configure builds and, if the ARM hardware opts are disabled, configure builds should still work just fine in libpng 1.6

Over the years there have been at least two separate PRs to fix the cmake stuff.  I worked on one of them and I believe the fix was eventually accepted by @ctruta into libpng 1.8, but no guarantees it still works.  @ctruta did not accept the prior fixes for this bug into libpng 1.6

